PR #5, opened 2016-08-04
Add install script of code style and inspections for IntelliJ IDEA

kunny:
Added install script for code style and inspections from https://github.com/square/java-code-styles.


shyiko:
Thank you! :beer:

PR #6, opened 2016-08-05
Fix NoUnusedImportsRule failing on escaped imports

dstarcev:


shyiko:
@dstarcev :+1: Thank you!

PR #56, opened 2017-06-07
Space after property accessor

arturbosch:
And no space if property accessor has no block body. See https://kotlinlang.org/docs/reference/properties.html

```kotlin
var setterVisibility: String = "abc"
    private set
```

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundKeywordRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundKeywordRule.kt
@@ -7,6 +7,7 @@ import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.PsiWhiteSpaceImpl
 import org.jetbrains.kotlin.com.intellij.psi.tree.TokenSet
 import org.jetbrains.kotlin.com.intellij.psi.util.PsiTreeUtil
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.lexer.KtTokens.CATCH_KEYWORD
 import org.jetbrains.kotlin.lexer.KtTokens.DO_KEYWORD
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundKeywordRule.kt
```diff
@@ -16,32 +17,44 @@ import org.jetbrains.kotlin.lexer.KtTokens.IF_KEYWORD
...
+                }
+            } else if (keywordsWithoutSpaces.contains(node.elementType) && node.nextLeaf() is PsiWhiteSpace) {
+                val parent = node.parent
+                if (parent is KtPropertyAccessor && parent.hasBody()) {
+                    emit(node.startOffset, "Unexpected spacing after \"${node.text}\"", true)
+                    if (autoCorrect) {
+                        node.nextLeaf()?.delete()
+                    }
+                }
+            } else if (noLFBeforeSet.contains(node.elementType)) {
```
shyiko: @arturbosch I think we have a regression here. Previously this block would be executed regardless of the preceding if clause (there was no else) (i.e. there could be 2 LintError's emitted for the same node: `Unexpected newline before "else"` & `Missing spacing after "else"`.) but no anymore.

shyiko:
Thank you the PR, Artur! This PR makes sure there is no extra space between set/get and the following "(", right? PR description implies extra space after "private set" but I don't think it's inserted.

ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundKeywordRule.kt
arturbosch: Yes your right, my fault - the else keyword is in both sets.

arturbosch:
Argh yes, did not check my spelling. 

There should be no space after a property accessor. But if the property accessor has no block statement, then it means, the users indention was to change the visibility of the getter/setter which needs no check for spacing. KtLint expected a space which was a false positiv. :)

shyiko:
Thank you :bowing_man:

PR #60, opened 2017-06-23
Fix misleading indent lint error message

jeremymailen:
**Why**
The error message was hardcoded to 4 so read incorrectly when a custom indent size was used.


shyiko:
:man_facepalming: Thank you, Jeremy!

Issue #48, opened 2017-05-30
Wildcard imports


fwilhe:
Please forgive my bikeshedding, but I beg to differ with the "no wildcards" rule. My primary argument is that intellij does add wildcard imports in the default configuration, and I think I should not have to configure intellij to play nice with ktlint. Adding `// ktlint-disable no-wildcard-imports` is a very ugly workaround.

Another acceptable solution for me would be if the format action actually resolved wildcard imports into individual imports, which it does not as of version 0.7.0. I don't want to argue about what's better, I just want to have a consistent style without requiring manual intervention.

shyiko:
Hi @fwilhe. 

I presume you have read https://github.com/shyiko/ktlint#-with-intellij-idea (option 1). 

Resolving * imports into individual imports is actually on my todo list. The thing is, it will require deeper integration with the compiler chain and it's hard to say how much time it's going to take. But, you (and anyone else bold enough to venture that way) are welcome to try it out. I'd happy to merge such PR in.

tldr; One way or another it's going to be done.

fwilhe:
> tldr; One way or another it's going to be done.

Awesome, thanks for the prompt reply.

asarazan:
Would it be possible to override this rule via the .editorconfig file? Seems like adding `no-wildcard-imports = false` would be a pretty elegant solution.

shyiko:
@aem Thank you for the PR. It looks ready to merge (as soon as we have reached some kind of consensus here). 

@asarazan My concern is - EditorConfig [isn't really supposed to deal with language-specific settings](https://github.com/editorconfig/editorconfig/issues/231#issuecomment-139810471). I'm aware of https://github.com/editorconfig/editorconfig/wiki/editorconfig-properties but I'm not sure how I feel about it. On the one side adding this property feels like abusing .editoconfig and bike-shedding, on the other - it could reduce contention. 

> BTW: Intellij does not support anything but `indent_size`, `continuation_indent_size`, `tab_width`, `indent_style` at the moment (note that all of them are *editor-centric*).

aem:
@shyiko I definitely agree that this feels a bit like fitting a square editorconfig into a round hole, not sure where else it would go though. i just went with editorconfig since it's what was used for the indents (which is admittedly the right use of editorconfig)

[...]

PR #61, opened 2017-06-29
(rule):allow wildcard imports via editorconfig

aem:
while some people might not like wildcard imports, importing things like
an entire enum class via a wildcard is acceptable in most styleguides.
this PR adds an editorconfig option to override the default wildcard
functionality.


aem:
this is related to #48 

while it's not up to this tool to be opinionated about IDEs, i think an override option is valid.

jackylimel:
When will this pull request get merged? 

Android Studio cannot auto-import static functions, so wildcard import would be very convenient at this stage.

shyiko:
@jackylimel 

imageedit_2_7341367885

peek 2017-08-28 05-46

jackylimel:
It's very interesting that none of the Mockito static methods, like `when`, or `verify`, can be auto-imported..

jackylimel:
but, anyway, JetBrains IDE can optimize imports on the fly, e.g. if there are many classes from the same package, IDE will auto convert the imports to wildcard import. It would be great to use the IDE feature, rather than import the classes one-by-one

shyiko:
(unrelated to PR)

@jackylimel https://github.com/shyiko/ktlint#-with-intellij-idea contains info on how to turn 
> IDE will auto convert the imports to wildcard import

off (either manually or in completely automated fashion (`ktlint --apply-to-idea`)).

jackylimel:
I still don't understand why you have a strong opinion again this PR.

shyiko:
@jackylimel https://github.com/shyiko/ktlint/issues/48#issuecomment-312027052

henrik242:
+1

ktlint-ruleset-standard/src/test/kotlin/com/github/shyiko/ktlint/ruleset/standard/NoWildcardImportsRuleTest.kt
```diff
@@ -22,4 +22,17 @@ class NoWildcardImportsRuleTest {
...
+    fun testLintWithOverride() {
+        assertThat(NoWildcardImportsRule().lint(
+            """
+            import a.*
+            import a.b.c.*
+            import a.b
+            import kotlinx.android.synthetic.main.layout_name.*
+            """.trimIndent(),
+            mapOf("allow_wildcard_imports" to "true")
+        )).isEqualTo(listOf<LintError>())
```
jakubvimn: What about using "emptyList()"?

sebastienrouif:
@aem could you catch up with master branch ? That would be very helpful to have an updated version of this branch and SHA1 to refer to in our local snapshot. I hope your branch gets merged one day

aem:
@sebastienrouif sure, i'll resolve conflicts and update later today.

aem:
@sebastienrouif @shyiko updated to latest master

neworld:
Any news?

aem:
@iTomkinas please see the discussion in #48. Simply put, it's the maintainer's decision about whether or not this is a valid inclusion in the linter, and if it is valid whether an editorconfig line is the right way to implement it. I made my proposed solution and it seems like it's not the approach the maintainer would choose themselves. I've left the PR open in case that decision is reversed, or in case someone would like to fork the project and include this patch on their own.

aem:
it was decided this isn't a valid lint rule. closing this PR to indicate that the issue has been thoroughly discussed and reached its natural conclusion.

PR #62, opened 2017-06-30
allow for named blocks

aem:
Kotlin supports named blocks in cases where you want to return early or return from an inner block without returning a function, the syntax is as so:

```kotlin
fun printNFromList(list: List<*>, count: Int) {
  list.mapIndexed a@{ item, index ->
    if (index > count) return@a
    println(item)
  }
}
printNFromList(listOf(1, 2, 3), 2)
```

obviously that's contrived because `return@map` would suffice as well, but our code sometimes requires operations like `.map { .map { } }` which means `return@method` doesn't work, named blocks are required.

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundCurlyRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundCurlyRule.kt
@@ -20,7 +20,7 @@ class SpacingAroundCurlyRule : Rule("curly-spacing") {
             val spacingBefore: Boolean
             val spacingAfter: Boolean
             if (node.textMatches("{")) {
-                spacingBefore = prevLeaf is PsiWhiteSpace || (prevLeaf?.node?.elementType == KtTokens.LPAR &&
+                spacingBefore = prevLeaf is PsiWhiteSpace || prevLeaf is KtTokens.AT || (prevLeaf?.node?.elementType == KtTokens.LPAR &&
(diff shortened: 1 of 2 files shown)


shyiko:
Hi @aem. Did you mean `prevLeaf?.node?.elementType == KtTokens.AT`? (compilation failed)

aem:
whoops. yes. this is what i get for using vim 😞

shyiko:
@aem no worries! :) Thank you for the PR! :bowing_man:

aem:
@shyiko always, never a fan of people opening issues on my projects without trying to do something about it, figured i'd be a good OSS citizen myself :)

shyiko:
@aem we might have a problem here. 

I was about to release a new version of ktlint but decided to check what Intellij formatter does when asked to reformat a block of code containing lambda expression with a label and, alas, a space is inserted between \@ and {.

```kotlin
fun foo() {
    ints.forEach lit@{
        if (it == 0) return@lit
        print(it)
    }
}
```
becomes 
```kotlin
fun foo() {
    ints.forEach lit@ {
        if (it == 0) return@lit
        print(it)
    }
}
```
(example taken from https://kotlinlang.org/docs/reference/returns.html)
In other words, previous behavior was in line with IDEA's built-in formatter (which pretty much reflects official code style).

aem:
interesting. looks like you're correct and we've been doing this differently in our codebase. i'm happy to update our styleguide to add the space before the brace if you want to back this change out

shyiko:
@aem :handshake: yeah, I think it's better this way (staying as much IDEA-code-style-compatible as possible). Sorry about all the confusion.

aem:
no worries, should've checked docs/intellij's formatting first

PR #72, opened 2017-08-11
Fix unnecessary space between function signature and return type

### LINKED ISSUE #68: Flag unnecessary space between function signature and return type
I'd except the following code to get flagged since the scape before `:` is not needed:

```kotlin
interface Something {
  fun render(time: Long) : String
}
```

Instead, it should look like this:

```kotlin
interface Something {
  fun render(time: Long): String
}
```

Tested with `0.8.3` and `0.9.0`.

JLLeitschuh:
Closes #68

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundColonRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundColonRule.kt
@@ -2,10 +2,10 @@ package com.github.shyiko.ktlint.ruleset.standard
 
 import com.github.shyiko.ktlint.core.Rule
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
-import org.jetbrains.kotlin.com.intellij.psi.PsiMethod
 import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.PsiWhiteSpaceImpl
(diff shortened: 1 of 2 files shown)


shyiko:
Hi Jonathan. Thank you for the PR! Do you think you could change it a little bit to account for other cases (e.g. `fun fn(a: String, b : String) {}`)? It looks like https://github.com/shyiko/ktlint/blob/master/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/SpacingAroundColonRule.kt#L21 can be used to derive a "stronger" condition - `val redundantSpacingBefore = node.prevSibling is PsiWhiteSpace && node.parent !is KtClassOrObject` (at least at first glance).

JLLeitschuh:
In this example `fun fn(a: String, b : String) {}` which would be considered *correct*? Given the context of the PR I think probably `a: String`.

shyiko:
That's right, `a: String` (https://kotlinlang.org/docs/reference/coding-conventions.html#colon). The rule of thumb: extra space added before ":" in type/supertype declarations, but it's expected to be omitted everywhere else.

JLLeitschuh:
This now also covers reformatting this: `<T : PomModelAspect>`, to this: `<T: PomModelAspect>`

shyiko:
It should `<T : PomModelAspect>`. As far as "whitespaces go", ktlint must be 100% compatible with IDEA's built-in formatter.

JLLeitschuh:
Any hints on how to fix my existing code to make that work?

shyiko:
@JLLeitschuh Sorry, I've been on the Internet with limited availability the last couple of days. I see you already took care of the generics. Is there anything else you want to do in this PR or that's about it?

JLLeitschuh:
Nope, I'm happy with this PR as is!

shyiko:
Thank you, Jonathan! Much appreciated :bowing_man:

PR #73, opened 2017-08-11
Expand .gitignore with gitignore.io

JLLeitschuh:


shyiko:
Any particular reason for this? I feel like having shorter `.gitignore` that concerns itself with project-only files is easier to support + OS & editor specific entries should (arguably) be in developer's `~/.gitignore_global`. I'm curious to know your position on this.

JLLeitschuh:
Different projects have different rules, especially around IDE's.
Some projects choose to let developers check in IDE specific files (my company did this with eclipse files) and the the [Kotlin Project](https://github.com/JetBrains/kotlin) has all of the IDE config files checked into source control.

I don't want to have to swap out my `~/.gitignore_global` file depending upon the repo that I have checked out.

As for OS, the only reason that I've ever found it to be useful is when you open a project in something like a docker mount where other OS files might get generated and end up in the repo.

JLLeitschuh:
Any additional thoughts on this?

shyiko:
I feel like [gitignore.io](https://www.gitignore.io/) is a great starting point for `~/.gitignore_global` but not `<project>/.gitignore`; simply on the basis of YAGNI.

> Some projects choose to let developers check in IDE specific files (my company did this with eclipse files) and the the Kotlin Project has all of the IDE config files checked into source control.

While I personally think adding editor-specific files is generally a terrible idea (I've seen this backfire more times than I care to remember), there is nothing stopping those projects from adding `!.idea/` to their `.gitignore` if they really want to. On a separate note, .gitignore_global (same as .gitignore) entries have no affect on "already commited files".

> I don't want to have to swap out my ~/.gitignore_global file depending upon the repo that I have checked out.

`.gitignore` whatever you want (that is not OS-specific) and I'll merge the PR in. Even `.idea` if you must ;) But I'm not convinced that adding ~200 lines to `.gitignore` is justified (especially considering that most of the entries are not applicable to this project).

shyiko:
Closing per https://github.com/shyiko/ktlint/pull/73#issuecomment-327239350.

PR #85, opened 2017-09-27
Add task group and description to the Gradle example

twaddington:
Ref https://github.com/ktlint/ktlint.github.io/pull/1


shyiko:
@twaddington Thanks again! :bowing_man:

PR #92, opened 2017-10-22
Detect needless blank lines

Ilya-Gh:
Closes #65 as IDEA's formatter is now able to detect blank lines before `}`. 

Please note that additionally to the case described in the issue 
```
fun main() {
    fun a() {
    }
    
}
``` 
The rule also prohibits the following construction: 
```
fun a() {

}
```
and formats it to 
```
fun a() {
}
```

Issue #65, opened 2017-07-06
Suggestion: Needless blank line(s) additions


vanniktech:
The check, Needless blank line(s), flags the following code:

```kotlin
}


private fun something() { }
```

although I'd also expect it to flag the one below as well:

```kotlin
  }

}
```

since the newline between the `}` is not necessary.

shyiko:
Hi @vanniktech. 

While I like the idea, I'm note sure Intellij built-in formatter supports this (there is a setting for Java (Code Style -> Java -> Blank Lines -> Before '}') which can be set to 0 to enforce this kind of style, but not for Kotlin) (at least not in IDEA 2016.3). It's pretty easy to implement but I'd prefer not to add anything that can't be enforced by the IDEA's formatter (if we can).

vanniktech:
[There will be an option with 1.1.4[(https://youtrack.jetbrains.com/issue/KT-15504).

shyiko:
@vanniktech :+1:

vanniktech:
Will this mean this functionality will be added now or once 1.1.4 is out?

shyiko:
I think we should hold off until 1.1.4 is out (just to be on the safe side).

vanniktech:
[Kotlin 1.1.4 is out](https://blog.jetbrains.com/kotlin/2017/08/kotlin-1-1-4-is-out/)

shyiko:
Changed contributed by @Ilya-Gh :tada:. Keeping this issue open until `ktlint-intellij-idea-integration` is updated too.

vanniktech:
This can be closed now. Thanks :)

PR #93, opened 2017-10-23
Update intellij idea integration with no blank line before rbrace rule

Ilya-Gh:
As requested in #65 this PR update `ktlint-intellij-idea-integration` according to new added rule in #92.


shyiko:
Awesome :zap: Thanks again, @Ilya-Gh!

PR #104, opened 2017-11-21
Respect continuation indent size, make IndentationRule more intelligent

### LINKED ISSUE #76: Support .editorconfig's continuation_indent_size
According to ktlint both of these are valid:
```kotlin
fun Project.`spotless`(configure: com.diffplug.gradle.spotless.SpotlessExtension.() -> Unit = {}) =
    extensions.getByName<com.diffplug.gradle.spotless.SpotlessExtension>("spotless").apply { configure() }
```
```kotlin
fun Project.`spotless`(configure: com.diffplug.gradle.spotless.SpotlessExtension.() -> Unit = {}) =
        extensions.getByName<com.diffplug.gradle.spotless.SpotlessExtension>("spotless").apply { configure() }
```

Similarly both of these are valid:
```kotlin
val someSet = setOf(
    "a"
)
```
```kotlin
val someSet = setOf(
        "a"
)
```

MyDogTom:
closes #76 

### Motivation
In my project, indent size is 4, but continuation size is 6. It means, that condition `line.length % indent != 0` always fails.

### Implementation details
Rule now calculates indent size relatively to parent element and compares it to expected indent size. Rule knows several cases when continuation indent should be used. For sure, it's not a complete list of cases. These are emerged during testing on my code base and I expect to see more of them when other people start to use it.

### Configuration
Continuation indent size can be specified via `continuation_indent_size` from `.editorconfig`. By default, regular indent size is used.

### Backward compatibility
Due to the fact, that previous implementation was too permissive, I expected that current rule can find some new violation. But for well formatted code, it should return same result. Of course, for the cases when `continuation_indent_size % indent_size == 0`

### Performance
I haven't noticed any major performance impacts. I've tested it on codebase with approximately 1800 kt files. `0.12.1` takes ~16s and new version takes ~ 17s
Nevertheless, I think about current implementation as "first working approach". I'm pretty sure that it can be improved either by using existing "Utils" class, either my implementing more intelligent and fast parent indent size.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ target
 .gradle
 build
 !ktlint/src/main/resources/config/.idea
+/.idea
+*.iml
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRule.kt
@@ -2,14 +2,26 @@ package com.github.shyiko.ktlint.ruleset.standard
 
 import com.github.shyiko.ktlint.core.KtLint
 import com.github.shyiko.ktlint.core.Rule
+import org.jetbrains.kotlin.KtNodeTypes
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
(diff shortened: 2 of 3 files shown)


shyiko:
This is an amazing work, @MyDogTom! Thank You! :trophy:

PR #106, opened 2017-11-22
new Rule: No line break before assignment

MyDogTom:
Line break is allowed only after assignment (eg. equals sign)
Auto formatting is supported via simplistic approach by putting
same line break with same indent from left side to right side of assignment

Closes shyiko/ktlint#105


shyiko:
Thank you, @MyDogTom. Perfectly executed PR! :100:

PR #125, opened 2017-12-12
new rule: NoLineBreakAfterElse

MyDogTom:
Reports line breaks after "else" clause. Supports auto format by replacing linebreak with space.
Doesn't take into account maximum line length.

closes https://github.com/shyiko/ktlint/issues/122


shyiko:
Thank you, @MyDogTom. As usual :+1:

PR #132, opened 2017-12-27
Use JVM_CONFIG_FILES instead of EMPTY

### LINKED ISSUE #131: ktlint not compatible with kotlin 1.2.20-eap-11
Ref. https://github.com/jeremymailen/kotlinter-gradle/issues/33

It looks like a symbol ktlint dependents on, EnvironmentConfigFiles.EMPTY, has been removed in the EAP version of kotlin-compiler-embeddable.

More info about 1.2.20-eap-11: https://discuss.kotlinlang.org/t/kotlin-1-2-20-eap/5896

MyDogTom:
Because `EnvironmentConfigFiles.EMPTY` is removed in 1.2.20
This is copy-paste of the similar change from Kotlin itself
https://github.com/JetBrains/kotlin/commit/afc9d3ef8b5bd4f2db44823f4eb62c3ce7d5dfd1#diff-0326941c85afd979faa6998b7b95bff8R55

fixes #131

Commit message from kotlin repository says
>Drop EnvironmentConfigFiles.EMPTY and replace its only usage in
preprocessor with JVM_CONFIG_FILES: it's easier and won't affect
anything

Unfortunately, I have no idea if the same reasoning applies to KtLint or not. My feeling says yes, because `KtLint` and `Preprocessor` classes look very very similar.

### DIFF
--- a/ktlint-core/src/main/kotlin/com/github/shyiko/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/github/shyiko/ktlint/core/KtLint.kt
@@ -49,7 +49,7 @@ object KtLint {
         }
         DiagnosticLogger.setFactory(LoggerFactory::class.java)
         val project = KotlinCoreEnvironment.createForProduction(Disposable {},
-            CompilerConfiguration(), EnvironmentConfigFiles.EMPTY).project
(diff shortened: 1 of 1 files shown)


jeremymailen:
Thank you for looking into this!

shyiko:
Thank you, @MyDogTom!

PR #134, opened 2018-01-03
Revert "Changed continuation_indent_size to 4"

bradmcmanus:
This reverts commit ed261175241807e0f7bb0428db8f2d0c253f6e5e.

As per discussion from the GitHub issue [1], the default continuation
indent for Android should be 8 rather than 4. This matches the Android
Kotlin style guide.

[1] https://github.com/shyiko/ktlint/issues/120

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ ktlint recognizes the following [.editorconfig](http://editorconfig.org/) proper
 # possible values: number (e.g. 2), "unset" (makes ktlint ignore indentation completely)  
 indent_size=4
 # possible values: number (e.g. 2), "unset"
+# it's automatically set to 8 on `ktlint --android ...` (per Android Kotlin Style Guide)
 continuation_indent_size=unset
 # true (recommended) / false
 insert_final_newline=unset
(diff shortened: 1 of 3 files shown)


shyiko:
@bradmcmanus :bowing_man:

bradmcmanus:
hey @shyiko -- I notice that this commit is mysteriously not present on new releases of ktlint (I would expect it to be present on all releases after 0.14). did you accidentally force push this commit out? or maybe you removed it on purpose? thanks!

shyiko:
Hi @bradmcmanus. 
I keep forgetting to ask clarification on https://github.com/android/kotlin-guides/issues/37 but it looks like continuation_indent_size=4 might become official (for example, [android/android-ktx](https://github.com/android/android-ktx/blob/ed32bd844e07207bc2751193567e74e95e2f12d9/build.gradle#L99) is using ktlint --android profile) and so I have reverted the change to avoid "back and forth" between 4 and 8.

bradmcmanus:
@shyiko thank you for the clarification. I agree that the back and forth is troublesome. 4 is reasonable, especially if ktx is using 4. 4 it is!

PR #136, opened 2018-01-10
Use HTTPS and upgrade Maven plugins/dependencies

jhult:
Use HTTPS for links in README
Use HTTPS for JCenter
Upgrade Maven plugins and dependencies


shyiko:
Thank you, Jonathan!

PR #139, opened 2018-01-19
IndentationRule: fix auto correction

### LINKED ISSUE #116: ktlint not able to add spaces to indentation
I'm using gradle-kotlinter-plugin 1.6.0 that uses ktlint 0.13.0, and I guess indentation should be fairly trivial to fix?

    > Task :shared-http-client:formatKotlinMain 
    Format could not fix > src/main/java/Http.kt:56:1: Unexpected indentation (4) (it should be 8)

Ref. https://github.com/jeremymailen/kotlinter-gradle/issues/30

MyDogTom:
Corrected whitespace should be equal to expected indent + previous indent

Fixes #116 

@shyiko Would it be possible to release 0.15.1 fix with these changes?

PS. 
While testing I've found some cases when `IndentionRule` doesn't report violation. Also on my code base I had to apply `detektFormat` twice. 
These two are not trivial to fix. I'll describe them later in a separate issue.

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRule.kt
@@ -75,7 +75,7 @@ class IndentationRule : Rule("indent") {
                             "Unexpected indentation (${line.length - previousIndent}) " +
                                 "(it should be $expectedIndentSize)",
                             true)
-                        if (autoCorrect) replaceWithExpectedIndent(node, expectedIndentSize)
+                        if (autoCorrect) replaceWithExpectedIndent(node, previousIndent + expectedIndentSize)
                     }
                     offset += line.length + 1
                 }
--- a/ktlint-ruleset-standard/src/test/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRuleTest.kt
+++ b/ktlint-ruleset-standard/src/test/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRuleTest.kt
@@ -585,4 +585,33 @@ class IndentationRuleTest {
                 "continuation_indent_size" to "6")
         )).isEmpty()
     }
+
+    @Test
+    fun shouldRespectPreviousIntent() {
+        assertThat(
+            IndentationRule().format(
+                """
+            fun setUp() {
+                 helper = ClassA(
+                    paramA,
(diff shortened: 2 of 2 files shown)


shyiko:
Hi @MyDogTom. Thank you for the fix! 

As for 0.15.1: Changes introduced in https://github.com/shyiko/ktlint/pull/137 were not released as part of [0.15.0](https://github.com/shyiko/ktlint/blob/master/CHANGELOG.md#0150---2018-01-18). I want to make absolutely sure we won't introduce any regressions (+ that `indent` rule is feature-complete) before releasing them.

MyDogTom:
Do you have `snapshot` build or something similar? This would allow people to try out these changes and report their findings.

vanniktech:
I'd also be very interested in this.

shyiko:
@MyDogTom @vanniktech done - https://github.com/shyiko/ktlint#access-to-the-latest-master-snapshot (latest snapshot of master is now available through [Sonatype Snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/com/github/shyiko/ktlint/) and [GitHub Releases](https://github.com/shyiko/ktlint/releases)).

vanniktech:
Awesome, I'll test this out tomorrow and report back.

vanniktech:
Quickly tried it out and it works perfectly in my projects. Haven't found something I would not have wanted.

vanniktech:
@shyiko seems like @MyDogTom also tried this out and gave thumps up. When can we expect another release?

MyDogTom:
I applied snapshot version to my library. That time with proper setup :) So far I've found only one "minor" issue. I had to apply auto correction twice. I plan to investigate that case and create separate issue for that. I'll do it during weekend. 

PS. By the way, @vanniktech thank you for your hint [how to use specific snapshot version](https://medium.com/@vanniktech/pinning-a-snapshot-dependency-to-a-specific-version-b477f79469fc). I've found it just in time. 👍

MyDogTom:
Looks like I cannot use specific snapshot version, because they are deleted :( At least [sonatype](https://oss.sonatype.org/content/repositories/snapshots/com/github/shyiko/ktlint/ktlint-core/0.0.0-SNAPSHOT/) contains only version from 1st of February.

>I want to make absolutely sure we won't introduce any regressions (+ that indent rule is feature-complete) before releasing them.

@shyiko What is the next step? Is there anything I can do for that?

vanniktech:
Yeah I've noticed that too that sometimes they're deleted.

MyDogTom:
Actually, I've found one issue :) there is fix #151

PR #142, opened 2018-01-26
Fix erroneous unused import flagging on custom destructure operators

JelloRanger:
I noticed a bug with the unused import rule when I initially integrated in KtLint into one of my projects. It erroneously detected custom componentN operators as unused (which the auto formatter ended up removing and breaking my build :()

The Kotlin std library by default only provides `component1()` up to `component5()`, meaning that you can only destructure up to 5 values by default. See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/index.html and search for "component5" for reference

So if someone wants to destructure six values, i.e.:

`(one, two, three, four, five, six) = someList`

... the IDE will throw an error since `component6()` doesn't exist. To get around this you can write your own component6() method pretty easily:

`inline operator fun <T> List<T>.component6() = get(5)`

.. and when this method is imported it's flagged by KtLint as unused since the unused import rule only checks for the default destructure methods (component1...5()).

To solve this I added a regex that'll match against componentN (where N is some number) and ignore the normal unused import checking.

I'm also open to suggestions if you think there is a better way of handling this case.


shyiko:
Hi Jacob. Thank you for the PR! It's perfect. :heart:

PR #144, opened 2018-01-28
Complete gradle build of ktlint.

liutikas:
- Add new dependencies needed to build/test ktlint
  (missed these due to my branch being out of date)
- Set jar target level be java 8.
- Added shadowjar plugin that allows to build uber jar
  for running ktlint.
  Use: ./gradlew shadowJar
  Then use ktlint/build/libs/ktlint-all.jar


shyiko:
:bowing_man:

PR #148, opened 2018-01-30
Add the packaging metadata to build the ktlint snap

elopio:
This package will let you publish the latest ktlint in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and Linux distributions. You just have to go to https://build.snapcraft.io and enable the automated continuous delivery.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,26 @@
+name: ktlint
+version: git
+summary: An anti-bikeshedding Kotlin linter with built-in formatter
+description: |
+   Kotlin linter in spirit of feross/standard (JavaScript) and gofmt (Go).
+
+confinement: devmode
+
+apps:
+  ktlint:
+    command: java -jar $SNAP/ktlint-0.0.0-SNAPSHOT-shaded.jar
+    plugs: [home]
+    environment:
+      PATH: $PATH:$SNAP/usr/lib/jvm/java-8-openjdk-amd64/bin
+
+parts:
+  ktlint:
+    source: .
+    plugin: nil
+    build-packages: [openjdk-8-jdk]
+    stage-packages: [openjdk-8-jre]
+    build: |
+      ./mvnw
+      ./mvnw -Pcapsule clean package -Dmaven.test.skip=true
+    install:
+      mv ktlint/target/ktlint-0.0.0-SNAPSHOT-shaded.jar $SNAPCRAFT_PART_INSTALL
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -8,7 +8,7 @@ confinement: devmode
 
 apps:
   ktlint:
-    command: java -jar $SNAP/ktlint-0.0.0-SNAPSHOT-shaded.jar
+    command: ktlint
     plugs: [home]
     environment:
       PATH: $PATH:$SNAP/usr/lib/jvm/java-8-openjdk-amd64/bin
@@ -20,7 +20,6 @@ parts:
     build-packages: [openjdk-8-jdk]
     stage-packages: [openjdk-8-jre]
     build: |
-      ./mvnw
       ./mvnw -Pcapsule clean package -Dmaven.test.skip=true
     install:
-      mv ktlint/target/ktlint-0.0.0-SNAPSHOT-shaded.jar $SNAPCRAFT_PART_INSTALL
+      mv ktlint/target/ktlint $SNAPCRAFT_PART_INSTALL


elopio:
Hello @shyiko! We are here experimenting with kotlin to make sure that our story in the snap store works nicely there. I wasn't sure if packaging the shaded jar was the right thing to do, so I would appreciate any feedback you can give us.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,26 @@
...
+      PATH: $PATH:$SNAP/usr/lib/jvm/java-8-openjdk-amd64/bin
+
+parts:
+  ktlint:
+    source: .
+    plugin: nil
+    build-packages: [openjdk-8-jdk]
+    stage-packages: [openjdk-8-jre]
+    build: |
+      ./mvnw
```
shyiko: Is there a reason for this to be included?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,26 @@
...
+      ./mvnw -Pcapsule clean package -Dmaven.test.skip=true
+    install:
+      mv ktlint/target/ktlint-0.0.0-SNAPSHOT-shaded.jar $SNAPCRAFT_PART_INSTALL
```
shyiko: [ktlint/target/ktlint](https://github.com/shyiko/ktlint/releases/tag/0.15.0) (executable jar that [doesn't require](https://github.com/shyiko/ktlint#usage) `java -jar ...` to be run) should be used instead (produced by `./mvnw -Pcapsule clean package -Dmaven.test.skip=true`).

snap/snapcraft.yaml
elopio: I think I just misread your ci scripts. Removed.

snap/snapcraft.yaml
elopio: Awesome! updated.

elopio:
Thanks a lot for the review! Please check it again, and let me know if you have any question about snaps.

shyiko:
Thank you, Leo. Merging in :bowing_man:

shyiko:
@elopio any chance you have an example targeting different arch|s? I have enabled https://build.snapcraft.io/user/shyiko/ktlint build and see that there are i386/armhf/amd64 profiles and yet PATH env in snapcraft.yaml is set to $SNAP/usr/lib/jvm/java-8-openjdk-**amd64**/bin.

elopio:
@shyiko oh, you are right! Let me figure out how to solve that.

elopio:
@shyiko sorry for the delay, got a little busy here. But now I know, we can just move the arch specific path to something common.
And the fails are failing on build.snapcraft.io because it's not being able to go through the proxy. We might need to copy this from our mavehttps://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/maven.py#L121
Have you tried mvnw through a proxy?
Worst case scenario, we can always use travis: https://docs.snapcraft.io/build-snaps/ci-integration#using-travis

shyiko:
maven plugin instead of ./mvnw is fine by me. 
On a separate note, [snap/plugins/x-zulu-jre.py](https://github.com/zmartify/zmarthubsrv-snap/blob/185af1b139de281331cc3663ddb4b75acf7baf9b/snap/snapcraft.yaml#L28) looks like a possible (alternative) solution for arch targeting.

> Have you tried mvnw through a proxy?

I imagine it would work provided -Dhttp.proxyHost, -Dhttp.proxyPort / -Dhttps.proxyHost, -Dhttps.proxyPort are set (http_proxy/https_proxy are not recognized I'm afraid).

elopio:
@shyiko I was thinking of something simpler than the zulu solution, like:
https://github.com/shyiko/ktlint/pull/149

But this is pretty flexible, we can do it in any way you want.

About the proxy, do you mean that we should call it like:
```
proxy_split=$(echo $http_proxy | sed 's/\(.*\):/\1 /')
proxy_host=$(echo $var | awk '{print $1}')
proxy_port=$(echo $var | awk '{print $2}')
./mvnw -Pcapsule clean package -Dmaven.test.skip=true -Dhttp.proxyHost=$proxy_host -Dhttp.proxyPort=$proxy_port -Dhttps.proxyHost=$proxy_host -Dhttps.proxyPort=$proxy_port 
```
?

(please excuse my bash, I'm learning :p)

shyiko:
@elopio https://github.com/shyiko/ktlint/blob/master/snap/snapcraft.yaml updated to use http_proxy/https_proxy - https://build.snapcraft.io/user/shyiko/ktlint (proxy url parsing logic adopted from [gist:4cf82d42beb6387dec7eadc43cef6c17](https://gist.github.com/JPvRiel/4cf82d42beb6387dec7eadc43cef6c17)). The end result is a little bit messy but I guess it's unavoidable (a plugin could probably hide all of that under the carpet but it doesn't really matter at this point). 

Thank You once again for the PR(s) :bowing_man:

elopio:
:man_dancing: 
Thank you @shyiko! Drop me a line whenever you have some feedback about the snaps.

elopio:
ah, and in order to release this to the stable channel, where people will be able to find your snap searching on the store, you have to update it to `confinement: strict`, so it's properly isolated and secure. Here you will find more details: https://docs.snapcraft.io/reference/confinement

PR #149, opened 2018-02-01
snap: move java to an arch independent dir

elopio:
Remove the hard-coded arch path to find the java bin.


shyiko:
Thank you, Leo! :bowing_man:

PR #151, opened 2018-02-05
Improve line length calculation in ClassAndFunctionHeaderFormatRule

MyDogTom:
Fixes StringIndexOutOfBoundsException that can happen for class without body when max line length is specified.
Fixes line size calculation for secondary constructor and function.

How fix works: Just take entire line size, without any attempt to understand where is primary constructor, secondary constructor or class body. If line doesn't fit, than we should break parameter list definition into several lines.


MyDogTom:
As I understand test passes on Mac OS but fails on Windows? Strange... I tried to avoid that by using `System.lineSeparator()`. I'll investigate that further.

MyDogTom:
The problem is that `System.lineSeparator()` on Windows returns `\r\n` (as expected), but data used for tests  still uses `\n`. 
Any ideas how to better handle that?

shyiko:
@MyDogTom Thank You :bowing_man:

Issue #152, opened 2018-02-05
ktlint exception when using java 9


impatient:
Using ktlint from brew, version 0.15.0.  Works fine on Java 8, switched to Java 9 using jenv and get the following error.
```
Exception in thread "Thread-0" java.util.concurrent.ExecutionException: java.lang.NoClassDefFoundError: Could not initialize class com.github.shyiko.ktlint.core.KtLint
	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
	at com.github.shyiko.ktlint.Main$parallel$consumer$1.run(Main.kt:614)
	at java.base/java.lang.Thread.run(Thread.java:844)
Caused by: java.lang.NoClassDefFoundError: Could not initialize class com.github.shyiko.ktlint.core.KtLint
	at com.github.shyiko.ktlint.Main.lint(Main.kt:585)
```
Command is simply: ```ktlint```

shyiko:
#153 merged in but there are still couple of things that need to be fixed before this ticket can be closed:

```
$ ./ktlint/target/ktlint --version
null
 shyiko@shyiko-laptop  /tmp/ktlint master
$ ./ktlint/target/ktlint
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.jetbrains.kotlin.com.intellij.util.text.StringFactory (file:/tmp/ktlint/ktlint/target/ktlint) to constructor java.lang.String(char[],boolean)
WARNING: Please consider reporting this to the maintainers of org.jetbrains.kotlin.com.intellij.util.text.StringFactory
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

ifrins:
The exception and the version showing  have been fixed. What’s remaining is the illegal reflective access. This is a Kotlin compiler that has already been resolved upstream [KT-19051](https://youtrack.jetbrains.com/issue/KT-19051) and will land when Kotlin 1.2.30 is released. 

@shyiko do you have any plans for a release that is compatible with java 9?

shyiko:
@ifrins as soon as Kotlin 1.2.30 is out

shyiko:
UPDATE: 

[0.16.0](https://github.com/shyiko/ktlint/releases/tag/0.16.0) released with @ifrins fixes included (thank you once again, Francesc!). 

[...]

PR #155, opened 2018-02-15
Fix version returning null when running on Java 9

ifrins:
A bug in JDK-9 ([JDK-8190987](https://bugs.openjdk.java.net/browse/JDK-8190987)) caused `Package.getImplementationVersion()` to return null. This commits adds code to read the Manifest file from the JAR to get the Implementation Version when it's not possible to retrieve the version through the normal way.

This PR contributes to #152


shyiko:
Thank you, @ifrins! :+1: Merging in.

Issue #168, opened 2018-03-06
Line wrap for &&


vRallev:
I've seen f13b93e and probably because of that my code is wrong. My snippet looks like this
```kotlin
            .assertValueAt(2) {
                it is ImageChangeEvent.ImageUpdated
                        && it.position == 0
                        && it.changedFields.contains(CollectImage.IMAGE_MODE)
                        && it.image.imageMode == newImageMode
            }
```
Kotlin Lint complaints with `Line must not begin with "&&"`. I couldn't find anything regarding that in the Kotlin style guide and to me this rule seems wrong. Can you point me to the rule in the style guide? Is there a way to disable this specific check?

On a side note, this is working fine and Kotlin Lint doesn't complain
```kotlin
            .assertValueAt(2) {
                it is ImageChangeEvent.ImageUpdated
                        && it.position == 0 &&
                        it.changedFields.contains(CollectImage.IMAGE_MODE)
                        && it.image.imageMode == newImageMode
            }
```

shyiko:
See https://github.com/shyiko/ktlint/issues/163#issuecomment-369418775.
Also, thank you for the snippet.

shyiko:
Fixed in [0.20.0](https://github.com/shyiko/ktlint/releases/tag/0.20.0).

vRallev:
@shyiko No, it's still not working as expected.
```kotlin
if (!isLocal
    && workspace.userId != account.userId
    && workspace.contactId != account.userId) {
}
```
```
Class.kt:378:25: Line must not begin with "&&"
Class.kt:379:25: Line must not begin with "&&"
```
Should I create a new issue or do you want to reopen this one?

shyiko:
@vRallev It works in accordance with https://github.com/shyiko/ktlint/issues/163#issuecomment-369418775.

vRallev:
I can't follow the comments in the link. They're mostly talking about +/- and not && or ||. I couldn't find anything about that. Note that the auto-formatter inside of IntelliJ formats my sample from above like this:
```kotlin
if (!isLocal
    && workspace.userId != account.userId
    && workspace.contactId != account.userId
) {
    // do something after 4 space indent
}
```
But for the formatter the following is also fine
```kotlin
if (!isLocal && 
    workspace.userId != account.userId && 
    workspace.contactId != account.userId
) {
    // do something after 4 space indent
}
```
So there's no evidence which one of them is correct.

starkej2:
[...]

Issue #163, opened 2018-02-28
Why new OperatorWrapping rule?


Tolriq:
Can someone point the source for that decision?

Currently default for Java with checkstyle and many is to break before those.
http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/whitespace/OperatorWrapCheck.html

Kotlin guide https://android.github.io/kotlin-guides/style.html#where-to-break says: 
"When a line is broken at a non-assignment operator the break comes before the symbol."

So it seems consensus was more to have the break before the operator and as such && || at the start of the new line. It allows way better reading.

If there's no way to return to previous way is there a way to manage that via editorconfig or something?

SUPERCILEX:
This technically comes from the [Kotlin style guide](https://kotlinlang.org/docs/reference/coding-conventions.html#formatting-control-flow-statements), but I disagree with them, that if statement looks gross IMO. There should be a way to go back to the old behavior.

shyiko:
"When a line is broken at a non-assignment operator the break comes before the symbol." can lead (in case of `+`/`-`) to confusing situations like:
```kotlin
var v = 1
    - 2
// v == 1 not -1
```
(more on this at https://github.com/android/kotlin-guides/issues/51).

`&&` / `||` at the EOL appears to be the default both in Kotlin Style Guide as @SUPERCILEX pointed out (+ in https://github.com/yole/kotlin-style-guide/issues/9#issue-157734134) and Google own code bases, e.g. https://github.com/android/android-studio-poet. 

If decision is made (in https://github.com/android/kotlin-guides/issues/51) to wrap before `&&` / `||` - ktlint **will** be updated to follow that rule. Until then Kotlin Style Guide, version is enforced for consistency reasons (personally I don't care which one is selected as long as there is only one).

SUPERCILEX:
Alrighty, sounds fair.

Tolriq:
@shyiko well thing is that for me consistency with the rest of the code base is more important :(
My apps are slowly migrating to Kotlin, consistency helps a lot to avoid human mistakes like missing parenthesis or operator orders so is mandatory at least during that phase even if keeping multiple years of habit would be nice.
If there's no way to have a choice on that I guess I'll be stuck at 0.15.1 :(

kolboch:
Can this rule be switched off?

Tapchicoma:
[...]

PR #175, opened 2018-03-16
fix wrapping for && and ||

scottdweber:
I believe this fixes part of the problem described in #168 where certain characters will cause enforcement of the `Line must not being with "&&"` rule to fail.

I also did a bit of refactoring to try to make the code easier to read as I worked through figuring this out.

I'm not entirely sure if I've got the tokens separated properly between the "always" and "sometimes" groups.

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ChainWrappingRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ChainWrappingRule.kt
@@ -7,26 +7,35 @@ import org.jetbrains.kotlin.com.intellij.psi.PsiComment
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.PsiWhiteSpaceImpl
+import org.jetbrains.kotlin.com.intellij.psi.tree.IElementType
 import org.jetbrains.kotlin.com.intellij.psi.tree.TokenSet
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.lexer.KtTokens.ANDAND
+import org.jetbrains.kotlin.lexer.KtTokens.COMMA
 import org.jetbrains.kotlin.lexer.KtTokens.DIV
 import org.jetbrains.kotlin.lexer.KtTokens.DOT
+import org.jetbrains.kotlin.lexer.KtTokens.ELSE_KEYWORD
 import org.jetbrains.kotlin.lexer.KtTokens.ELVIS
+import org.jetbrains.kotlin.lexer.KtTokens.LBRACE
+import org.jetbrains.kotlin.lexer.KtTokens.LPAR
 import org.jetbrains.kotlin.lexer.KtTokens.MINUS
 import org.jetbrains.kotlin.lexer.KtTokens.MUL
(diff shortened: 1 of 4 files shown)


scottdweber:
I'm wondering if the "sometimes" group ought to only be +, -, *? I can't think of valid unary uses of / and %

shyiko:
I'm afraid something like
```kotlin
val s = ((1 + 2)
    / 3)
```
isn't going to be flagged (RPAR before MUL,PLUS,MINUS,DIV,PERC).
Would you be willing to give it another go?

scottdweber:
Hmm, I'm not sure this is the correct behavior anyway. The [Kotlin Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html) don't say anything explicit about how to break around operators, except for [chained calls](https://kotlinlang.org/docs/reference/coding-conventions.html#chained-call-wrapping). There is one code example that shows a wrap after the `&&` operator, but that particular behavior is not described or mentioned anywhere in the text.

The [Android Kotlin Style guide](https://android.github.io/kotlin-guides/style.html#where-to-break) indicates that a line break should come _before_ a non-assignment operator, which would seem to indicate that the `&&` should be on the new line, not on the original line.

The Kotlin spec [also says](https://kotlinlang.org/docs/reference/coding-conventions.html#formatting):

> In most cases, Kotlin follows the Java coding conventions.

Since the Kotlin spec doesn't really address line breaks outside of chained calls, perhaps we should fall back on Java? The [Java spec](http://www.oracle.com/technetwork/java/codeconventions-136091.html) for this matches the Android Kotlin style guide in that it recommends line breaks before non-assignment operators.

Thoughts?

shyiko:
See https://github.com/shyiko/ktlint/issues/163#issuecomment-369418775 (+ linked issues).

scottdweber:
Ah, right. I even read that whole string when I was thinking about working on this PR 🤦‍♂️ 

So, I'm thinking the best we can do is force `&&`, `||`, `/`, `%` at EOL and let `*`, `+`, `-` be at the beginning of a line if they are preceded by one of those "rule overriding" tokens. There are still times when this could be wrong:

```kotlin
var s = getSomeCoolInteger()
    - 1
```

But I'm not sure we can do any better strictly by looking at tokens. Thoughts?

shyiko:
Hey @scottdweber. 

I've noticed that some of the existing (pre-PR) assertions were wrong and so instead of dragging you through additional rounds of changes I decided to fix everything myself (https://github.com/shyiko/ktlint/commit/1219edbc9a29545f081edadbfd9e0dd28a52ab71). I hope you don't mind.

Regadless, thank you for the PR. I appreciate the effort.  

Fixed in [0.20.0](https://github.com/shyiko/ktlint/releases/tag/0.20.0).

scottdweber:
No worries. Thanks!

eventualbuddha:
Earlier in this thread, @scottdweber pointed out this:

> The Android Kotlin Style guide indicates that a line break should come before a non-assignment operator, which would seem to indicate that the `&&` should be on the new line, not on the original line.

My reading of this implies that `ktlint --android` would expect `&&` to appear at the start of a line rather than the end of a line. That is not how it currently works, however. Is following the Android Kotlin Style guide the desired behavior in this case?

shyiko:
@eventualbuddha https://github.com/shyiko/ktlint/pull/175#issuecomment-373938555

hick209:
I looked through the issue on https://github.com/pinterest/ktlint/issues/163#issuecomment-369418775 as you suggested @shyiko, but it sill does not resolve what @eventualbuddha pointed.
As things are, as per v0.30.0, `ktlint` diverges from the Android Kotlin Style.

Are there any plans to conforming with it on the boolean operators?

PR #183, opened 2018-04-08
Update ModifierOrderRule to include annotations as part of its checks

JelloRanger:
Annotations should precede all modifiers according to: https://kotlinlang.org/docs/reference/coding-conventions.html#modifiers

(It might make sense in the future to forbid having multiple annotations on the same line as the declaration per the link above. Not sure if that should be a new rule or not though; at the very least this should be an improvement over the existing check)

Had quite a bit of difficulty implementing the auto-format since annotation entries are a different type than all the other modifiers. I'm open to feedback on how to clean it up. :)

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ModifierOrderRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ModifierOrderRule.kt
@@ -2,7 +2,6 @@ package com.github.shyiko.ktlint.ruleset.standard
 
 import com.github.shyiko.ktlint.core.Rule
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
-import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.PsiWhiteSpaceImpl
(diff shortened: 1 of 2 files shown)


shyiko:
Thank you for the PR, Jacob. 
Any chance you could try preserving newline characters between annotations?
So that 
```kotlin
@A
@B(v = [
    "foo", 
    "bar", 
    "baz"
])
@C
suspend public fun f()
```
wouldn't get squashed into
```kotlin
@A @B(v = [
    "foo", 
    "bar", 
    "baz"
]) @C public suspend fun f()
```
.

JelloRanger:
Good catch! I ended up playing around and found a cleaner solution for the auto format that should fix this.

shyiko:
Great! 
One last thing before we merge this in. 
Could you please update the PR to output 

```
A.kt:3:1: Incorrect modifier order (should be "@annotation... public suspend")
A.kt:9:1: Needless blank line(s)
```

instead of 

```
A.kt:3:1: Incorrect modifier order (should be "@A @B(
    "_"
) @C public suspend")
A.kt:9:1: Needless blank line(s)
```

given something like

```kotlin
//

@A
@B(
    "_"
)
@C suspend public fun f() {}


//
```
(otherwise this can break tools that depend on "one error per line")

shyiko:
Thank you, Jacob! :+1:

PR #184, opened 2018-04-08
Fix syntax of main method in AST section of README

JelloRanger:
This took me longer than I'm willing to admit to realize the mistake when trying to run this example locally. :/


shyiko:
Oh my... Sorry about that.

PR #185, opened 2018-04-12
Proposing a more verbose pre-commit githook

paramsen:
Proposing a more verbose pre-commit githook


shyiko:
Sorry @paramsen but I have to close it. I have received numerous requests to make pre-commit hook less chatty (yep, less verbose than it already is) and this change would complicate things. 
I do appreciate the time you took to submit the PR.

PR #193, opened 2018-04-21
Handle comments better when moving operators to the end of a line.

### LINKED ISSUE #192: chain-wrapping auto fix can incorrectly comment out operator
Here's a minimal example:

```kt
// example.kt
fun and(a: Boolean, b: Boolean): Boolean {
    return (a //
        && b)
}
```

When running `ktlint --format` on the file above, it reformats it to this:

```kt
// example.kt
fun and(a: Boolean, b: Boolean): Boolean {
    return (a // &&
        b)
}
```

This clearly introduces a syntax error due to the operator now being inside the line comment. I'd expect `ktlint` to move the operator before the line comment like so:

```kt
// example.kt
fun and(a: Boolean, b: Boolean): Boolean {
    return (a && //
        b)
}
```

There's another similar problem in this example:

```kt
// example2.kt
fun and(a: Boolean, b: Boolean): Boolean {
    return (a
        //
        && b)
}
```

Which becomes this:

```kt
// example2.kt
fun and(a: Boolean, b: Boolean): Boolean {
    return (a
        // &&
        b)
}
```

eventualbuddha:
Closes #192 

This is probably a poor implementation of a fix, but it seems to work. I'd appreciate pointers in how to make it better.

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ChainWrappingRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ChainWrappingRule.kt
@@ -5,6 +5,7 @@ import org.jetbrains.kotlin.KtNodeTypes
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
 import org.jetbrains.kotlin.com.intellij.psi.PsiComment
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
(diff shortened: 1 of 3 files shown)


shyiko:
Thank you both for the ticket and the PR, Brian. 
I do appreciated the effort.

Having said that, I made an alternative fix (which avoids including operator in PsiWhiteSpaceImpl (this could lead to hard to debug problems when combined with other rules mutating the AST) + fixes a similar issue when moving operators to the next line) in https://github.com/shyiko/ktlint/commit/953d6b0263be3baadaee733fb832a042788c32eb. I hope you don't mind.

Fix available in [ktlint@0.22.0](https://github.com/shyiko/ktlint/releases/tag/0.22.0).

Thanks again :beers:

eventualbuddha:
I don’t mind at all! I figured my approach wouldn’t be optimal since I don’t have all the context for the project. Thanks for the fix!

Issue #79, opened 2017-09-07
Option for prompt before auto-correct


jbarr21:
The `ktlint -F` option to auto-correct is great, but can you please add the ability to confirm before auto-correcting an instance of an issue? Something like the attached image would be nice to see how your source file will be modified.

<img width="1041" alt="checkstyle-autofix" src="https://user-images.githubusercontent.com/3178606/30178958-a7924878-93bf-11e7-842e-55705661dba7.png">

hzsweers:
alternatively - would you be open to a PR adding this option?

shyiko:
@jbarr21 @hzsweers absolutely! 
If you have the time to make a PR - that would be perfect (I can assure you that it will be merged in). If not, (sooner or later) I'll take care of it myself.

blcooley:
@shyiko Doing a little code triage. It looks like this issue can be closed since the PR has been merged.

jbarr21:
that PR is separate, this issue still exists

paul-dingemans:
Closing issue due to lack of progress.

PR #194, opened 2018-04-22
Add rule to check that a single top level class name matches the file name

JelloRanger:
This rule adheres to both Kotlin and Android Kotlin style guides:
https://kotlinlang.org/docs/reference/coding-conventions.html#source-file-names
https://android.github.io/kotlin-guides/style.html#naming

In order to make this change possible I had to plumb through the fileName into each Rule. This will also allow us to add additional style checks in the future (i.e. no underscores in file names, directory path matches package path, etc.)

Regarding autoFormat, I don't think there's a good solution here. If a violation occurs, then it's impossible for the tool to automatically determine whether to use the file name or the class name. The only alternative I can think of to this problem is to have Ktlint prompt the user for which name to use every time a violation occurs, which could be built after #79 is complete.


shyiko:
Hi Jacob (and thank you for the PR).

Everything inside `ktlint-core/` is considered to be part of the public API and as such all changes must be backward compatible. 

Starting from [ktlint@0.22.0](https://github.com/shyiko/ktlint/releases/tag/0.22.0) file path is now accessible via  
`val path: String? = node.psi.containingFile.getUserData(KtLint.FILE_PATH_USER_DATA_KEY)`. 

Could you please update the PR to use that instead?

Also,
- `ktlint-core/src/main/kotlin/com/github/shyiko/ktlint/core/ASTNodeExtensions.kt` needs to be removed (core should stay minimal).
- (ClassNameMatchesFileNameRule) `emit(0, ...)` instead of `emit(topLevelClassNames.first().offset, ...)` (this should allow to simplify rule to a few lines of code)
- `**/*.kts` should not be checked.

JelloRanger:
Thanks for the feedback! (and adding the file path to userData :)) Updated the PR.

ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ClassNameMatchesFileNameRule.kt
```diff
@@ -0,0 +1,44 @@
...
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit
+    ) {
+        // Ignore all non ".kt" files (including ".kts")
+        if (node.getUserData(KtLint.FILE_PATH_USER_DATA_KEY)?.endsWith(".kt") != true) {
+            return
+        }
+
+        val topLevelClassNames = mutableListOf<String>()
+
+        node.visit {
```
shyiko: It's better to avoid whole AST traversal with something like
```kotlin
val topLevelClassNames = node.getChildren(null)
    .filter { it.elementType == KtStubElementTypes.CLASS }
    .mapNotNull { it.findChildByType(KtTokens.IDENTIFIER)?.text }
```

ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ClassNameMatchesFileNameRule.kt
```diff
@@ -0,0 +1,44 @@
...
+            if (it.elementType == KtStubElementTypes.CLASS && it.treeParent.elementType == KtStubElementTypes.FILE) {
+                val classIdentifier = it.findChildByType(KtTokens.IDENTIFIER)
+                classIdentifier?.let {
+                    val className = it.text
+                    topLevelClassNames.add(className)
+                }
+            }
+        }
+
+        val name = Paths.get(node.getUserData(KtLint.FILE_PATH_USER_DATA_KEY)).fileName.toString().substringBefore(".")
```
shyiko: `node.getUserData(KtLint.FILE_PATH_USER_DATA_KEY)` should be extracted to a varible (referenced twice).

ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ClassNameMatchesFileNameRule.kt
```diff
@@ -0,0 +1,44 @@
...
+        if (topLevelClassNames.size == 1 && name != topLevelClassNames.first()) {
+            emit(0,
+                "Single top level class name [${topLevelClassNames.first()}] does not match file name",
```
shyiko: How about `Class ${className} should be declared in a file named ${className}.kt`?

ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ClassNameMatchesFileNameRule.kt
JelloRanger: I like this a lot more (removes the need to check for immediate parent and should also be more efficient :))

ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ClassNameMatchesFileNameRule.kt
JelloRanger: Done!

shyiko:
Perfect. Thank you so much, Jacob!

PR #200, opened 2018-04-30
Apply `parameter-list-wrapping` when `max_line_length` is exceeded

### LINKED ISSUE #96: Max line length rule is enforced but auto-formatting isn't supported?
It's cool `ktlint` can honor `max_line_length` in the `.editorconfig` file, but the auto-formatting doesn't seem to be supported?

For example, if I have this `Person.kt` file:
```
class Person(id: Int, firstName: String, lastName: String, middleName: String?, age: Int, email: String?, phoneNumber: String?, twitterHandle: String?)
```

Then when I do `ktlint -F Person.kt`, it would just complain `Exceeded max line length (120)` without formatting it.

ntsim:
Resolves #96

This commit applies `parameter-list-wrapping` and its associated linting/formatting when `max_line_length` is exceeded by a function or class declaration due to its parameter list being too long. It will also apply for nested function declarations.


ntsim:
@shyiko not really sure why this is failing on the AppVeyor build. Tests run fine locally, and the test failures in AppVeyor don't seem to suggest there's actually anything wrong with the output. 

Any ideas?

shyiko:
Not to worry (https://github.com/shyiko/ktlint/commit/15b284f2cb9b7bb75b45ce02010c9a15807290a4 take care of it).

Thank you for the PR, Nicholas! :beers:

PR #201, opened 2018-05-01
Fix ParameterListWrappingRule to correctly identify misplaced left parens

JelloRanger:
Noticed a dev in my team writing code like this that passed Ktlint, so adding a quick fix here for it.

Sidenote: It feels like identifying dangling left parens should be part of a broader rule. Right now, the following code is valid according to Ktlint (athough I think it'd be pretty rare to see a dev actually write something like this, Ktlint should still probably catch it):

```
if
(someCondition) {
    doSomething()
}
```

However, given that this error is likely a lot more common in class/function headers I just added the fix there for now (also because we're handling right paren formatting in it). Let me know what you think though!


shyiko:
Thanks once again, Jacob. Merging right in. :trophy:

Issue #198, opened 2018-04-25
Spaces on both sides of the double slash (//) that begins an end-of-line comment


egor-n:
The [JetBrains Kotlin Coding Conventions suggest](https://kotlinlang.org/docs/reference/coding-conventions.html#horizontal-whitespace) placing a space after `//`:

![image](https://user-images.githubusercontent.com/378493/39243277-9eafb52a-4895-11e8-97d3-94b567d374c4.png)

The [Android Kotlin Style guide suggests](https://android.github.io/kotlin-guides/style.html#horizontal) placing a space before and after the double slash (`//`) if it begins an end-of-line comment:

![image](https://user-images.githubusercontent.com/378493/39242963-9576e862-4894-11e8-82d0-259bf9c2b7c4.png)

----

Ktlint should flag lack of spaces around `//`.

**Correct:**

```kotlin
// comment
var debugging = false // comment
var debugging = false // comment//word
```

**Incorrect:**

```kotlin
//comment
var debugging = false// comment
var debugging = false //comment
var debugging = false//comment

vanniktech:
It should not only flag it but also autocorrect it.

shyiko:
Added in https://github.com/shyiko/ktlint/commit/4385ea52274cbaa0b732ca90c27b64a168a5d698 (thanks once again, @JelloRanger!)

marciogranzotto:
what about `//region` and `//endregion` comments on Android Studio? They only work if there are no spaces after the slashes

PR #202, opened 2018-05-01
Add CommentSpacingRule for end of line comment padding (#198)

JelloRanger:
- Should add spaces before end of line comments if they are not at the beginning of a line, and add spaces after the double slash (//) if not present

In the future this rule can support further comment formatting, esp. for KDocs if we want.


shyiko:
Thank you!

PR #207, opened 2018-05-05
Add ability to specify editorconfig path

charbgr:
* Adds the ability to user to specify the `.editorconfig` path.

### The problem:

I setup ktlint in my [project](https://github.com/charbgr/cliffhanger) and then I installed the [pre-commit](https://github.com/shyiko/ktlint/blob/master/ktlint/src/main/resources/ktlint-git-pre-commit-hook.sh) hook. "Real" code is under `cliffhanger/` folder and the `.editorconfig.` is placed under that folder. 
The issue is that pre-commit hook is being installed in the parent git directory (where not "Real" code exists) and the `.editorconfig` is placed in a different folder, resulting to wrong lint warnings.

With this fix I can modify my pre-commit hook and set editoconfig path by
```
$ git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs ktlint --relative . --editorconfig-path="cliffhanger/"
```

Let me know what you think and looking forward to your constructive feedback!

### DIFF
--- a/ktlint/src/main/kotlin/com/github/shyiko/ktlint/Main.kt
+++ b/ktlint/src/main/kotlin/com/github/shyiko/ktlint/Main.kt
@@ -10,6 +10,7 @@ import com.github.shyiko.ktlint.core.RuleExecutionException
 import com.github.shyiko.ktlint.core.RuleSet
 import com.github.shyiko.ktlint.core.RuleSetProvider
 import com.github.shyiko.ktlint.internal.EditorConfig
+import com.github.shyiko.ktlint.internal.EditorConfigFinder
 import com.github.shyiko.ktlint.internal.IntellijIDEAIntegration
 import com.github.shyiko.ktlint.internal.MavenDependencyResolver
 import com.github.shyiko.ktlint.test.DumpAST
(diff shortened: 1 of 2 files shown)


shyiko:
Hi @charbgr. 

How about we change .editorconfig lookup logic so that instead of checking only the workdir it would check the directory of [every matched file](https://github.com/shyiko/ktlint/blob/master/ktlint/src/main/kotlin/com/github/shyiko/ktlint/Main.kt#L464)?

[editorconfig.org](http://editorconfig.org/):
> Where are these files stored?
When opening a file, EditorConfig plugins look for a file named .editorconfig **in the directory of the opened file** and in every parent directory. A search for .editorconfig files will stop if the root filepath is reached or an EditorConfig file with root=true is found.

I realise it would require 1 more `stat (2)`(dir/.editorconfig) syscall for every new (sub)directory (unless [klob](https://github.com/shyiko/klob) is modified to expose [this](https://github.com/shyiko/klob/blob/master/src/main/java/com/github/shyiko/klob/internal/package.kt#L31)) and that the results will need to be cached to avoid excessive I/O but I believe making ktlint behave closer to the way editors (e.g. VSCode) do is the way to go here.

What do you think? Would you be willing to give it a try?

charbgr:
Sure, I will give a try!

What if I found two or more `.editorconfig`s in different directories? Should ktlint behave different for each one?

shyiko:
Yes. 

Given
```
/project/.editorconfig # let's say indent_size = 4
/project/a/.editorconfig # indent_size = 2
/project/a/A.kt
/project/b/.editorconfig # indent_size = 8
/project/b/B.kt
/project/c/C.kt
```
A.kt should observe `indent_size = 2`, B.kt `indent_size = 8` and C.kt - `indent_size = 4`.

https://github.com/shyiko/ktlint/blob/master/ktlint/src/main/kotlin/com/github/shyiko/ktlint/internal/EditorConfig.kt#L19

charbgr:
Hi again,

sorry for being late on this. I was busy these weeks.

I pushed my first implementation for custom `.editorconfig` per filepath. My solution doesn't require klob to expose anything.

What I did is that I am iterating files found from klob, iterate them and for each of them I am going backwards till I find the working directory(`workDir`). 

Let me know what you think!

charbgr:
Hey,

did you have the time to check the last update?

shyiko:
Sorry, I've been busy the last couple of weeks. Merging in! :+1:

bethcutler:
Is there any chance you might reconsider the flag to specify the configuration file?

In our use case, we are attempting to create a consistent style across a broad spectrum of projects.  Prior to this change, it was possible to enforce compliance by having one .editorconfig in the directory from which ktlint was run.  After this change, it is possible for any project to circumvent the configuration just by adding their own .editorconfig file in their directory.

shyiko:
@bethcutler sure. PR is welcome.

PR #212, opened 2018-05-16
CommentSpacingRule: Exclude //noinspection.

vanniktech:
Is there any magic that one needs to do in order to execute the tests?

`gradle test` didn't quite work.


vanniktech:
Any reason why there are two build systems used here?

shyiko:
@vanniktech Google's internal build infrastructure does not support Maven. Gradle scripts here are for their convenience. https://github.com/shyiko/ktlint/issues/102

vanniktech:
Ah good to know. It's a pity for everyone who wants to contribute though as they're super confused - just like me.

vanniktech:
Would it be possible to cut a new release with this?

shyiko:
@vanniktech sure, I'll do it within the next couple of days.

PR #229, opened 2018-06-16
Adds pre-push hook

charbgr:
Added a pre push hook, in case someone wants to run ktlint only before pushing.
It compares files that changed from remote to local branch.

You can install it with:
```bash
$ ktlint --install-git-pre-push-hook
```


shyiko:
Thank you!

PR #137, opened 2018-01-13
Class and function header format rules, autocorrection for IndentionRule

MyDogTom:
#### This PR contains
* new rule `ClassAndFunctionHeaderFormatRule`  that implements #130 
* autocorrection for `IndentionRule` (fixes for #116)

Changes from #126  and #118 are included into current PR.

#### New rule notes
Applies to class definition (eg. primary constructor), secondary constructor and function.
Wrapping is needed in two cases:
* line break is already present
* overall header length exceeds maximum allowed line length

#### Example

```kotlin
class WithLongClassHeader(parameter1: Int, parameter2: Int, parameter3: Int) {
    constructor(parameter1: Int, parameter2: Int, parameter3: Int) {
    }

    fun withLongDefinition(param1: Int, parameter2: Int, parameter3: Int): String {
        return " "
    }
}
```

becomes this

```kotlin
class WithLongClassHeader(
    parameter1: Int,
    parameter2: Int,
    parameter3: Int
) {
    constructor(
        parameter1: Int,
        parameter2: Int,
        parameter3: Int
    ) {
    }

    fun withLongDefinition(
        param1: Int,
        parameter2: Int,
        parameter3: Int
    ): String {
        return " "
    }
}
```

See tests for more examples

PR #239, opened 2018-06-22
Indent rule: auto correction and curly braces

### LINKED ISSUE #141: Indentation is not flagged correctly when using use
I'm in a 2 spaces project with the following code.

```kotlin
class ReaderFacade {
  fun run() {
    val output: Output = CsvOutput()

    Okio.buffer(Okio.sink(File("output.${output.fileEnding}"))).use {
        it.writeUtf8(stringOutput)
      }
  }
}
```

I'd expect it to get flagged since the line in use as well as the `}` is not indented correctly. Instead it should look like this:

```diff
-        it.writeUtf8(stringOutput)
-      }
+      it.writeUtf8(stringOutput)
+    }
```

I'm running version 0.15.0. Didn't try the snapshot out as it should only add quick fixing in case it gets reported. Correct me if I'm wrong.

MyDogTom:
This PR contains two changes: Auto correction and support for curly braces `}`.
Please review it commit by commit. Or let me know, if you prefer two separate PRs instead.

**Auto correction** ([jump to commit](https://github.com/shyiko/ktlint/commit/9bdec7eef13422c335245c6c2aa76310104443b7)]
Bring back auto correction support, that was initially introduced in #137 . 
How it works: Just adjust indent according to expected indent. Since rule is executed line by line, we can be sure that previous indent is already correct.

**Support for curly brances** ([jump to commit](https://github.com/shyiko/ktlint/commit/e5a018c74fd57dbb28d925dfb719750518d13b84))
Fixes #141 
How it works: We start distinguish between "regular indent" and "absence of indent".  Knowledge which indent to use is encapsulated in `expectedIndentForNode`. Also there are cases, when we should immediately "jump" to next parent during "previous indent" calculation. That knowledge is encapsulated in `getParentNodeForIndent`.

**Refactoring** ([jump to commit](https://github.com/shyiko/ktlint/commit/38ac35e1352e79806f3d55e06df892314948947f))
Group all cases when we skip calculation in one place. Add comments with a reason, why we are skipping them.

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/IndentationRule.kt
@@ -7,13 +7,14 @@ import org.jetbrains.kotlin.com.intellij.lang.FileASTNode
 import org.jetbrains.kotlin.com.intellij.psi.PsiComment
 import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
+import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.psi.KtParameterList
 import org.jetbrains.kotlin.psi.KtTypeConstraintList
 import org.jetbrains.kotlin.psi.stubs.elements.KtStubElementTypes
 
 class IndentationRule : Rule("indent") {
 
-    private var indentSize = -1
+    private var config: EditorConfig = EditorConfig(-1, -1, -1, false)
 
     override fun visit(
         node: ASTNode,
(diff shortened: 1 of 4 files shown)


ubuntudroid:
@shyiko can this be merged (after resolving the conflicts). It seems to be really useful! :)

shyiko:
Unfortunatelly, no.  
These changes do not really fix the problem: 

```sh
$ git clone --branch=indent-rule-autocorrection-and-curly-braces \
  https://github.com/MyDogTom/ktlint
$ cd ktlint/
$ ./mvnw -Pcapsule clean package -Dmaven.test.skip=true

$ cat <<EOF > A.kt
class A {
fun f() {
if (global() == true) {
println("")
}
}
}
EOF

$ ktlint/target/ktlint -F A.kt

$ cat A.kt 
class A {
fun f() {
if (global() == true) {
println("")
}
}
}

# nothing changed
```

I realize that they might catch some additional cases but as any change can result in a regression I prefer to abstain from merging PRs that do not fix the underlying issue.

What I'm looking for is a PR that would calculate & fix the indentantion is one pass (for simplicity & performance reasons).

MyDogTom:
@shyiko  In your example, indent is zero. This is not detected by library (at least at the moment of PR creation) and PR doesn't aim to fix that.
This PR adds autocorrection only for the cases that are are already detected as errors.
As for curly braces, current PR adds support for them, but indent should be greater then zero. Again this is a more global error, that I don't even try to fix here.

>What I'm looking for is a PR that would calculate & fix the indentantion is one pass (for simplicity & performance reasons).

Then this PR is definitely not what you expect. PR only adds autocorrection to what is already detected by library without changing how it detects errors. Sorry, I misunderstood what is needed :(

shyiko:
@MyDogTom if anyone has to be sorry in this case then it should be me for not communicating properly. Don't get discouraged by something that is clearly not your fault.

Ultimately, every issue/PR (merged or not) gets us closer to a "better" ktlint.
I do appreciate all the work you've put into ktlint, so keep it up.

PR #254, opened 2018-07-20
Use permanent link to file as FilenameRule showcase reference

Ilya-Gh:
Changed the link to a rule showcase to point to a spicific commit, rather than to master, because the version of a file at the head of branch can change as new commits are made and the file contents might not be the same when someone looks at it later.


shyiko:
Thank you!

PR #257, opened 2018-07-25
Fix escape characters in the generated json.

### LINKED ISSUE #256: Generated json from json-reporter is malformed.
Version 0.25.0.

Currently the json generated by the reporter looks like the following:

```json
{
	"file": "src\main\java\com\saa\sample_app\base\SomeClass.kt",
	"errors": [
		{
			"line": 23,
			"column": 63,
			"message": "Unused import",
			"rule": "no-unused-imports"
		}
	]
}
```

As you can see, the slashes in the file name are not properly escaped. Presumably this is an error in [JsonReporter](https://github.com/shyiko/ktlint/blob/master/ktlint-reporter-json/src/main/kotlin/com/github/shyiko/ktlint/reporter/json/JsonReporter.kt)

I don't have much experience with Kotlin yet, but I'll see about getting a merge request going soon.

Locitao:
Fixes #256 

I've ensured that the generated json regarding the LinterError is correctly escaped according to the json specification. I've also included a unit test to verify this new behaviour.


shyiko:
Thank you!

PR #260, opened 2018-07-27
Fix for #259: use a separate class loader.

### LINKED ISSUE #259: --reporter/--ruleset Java 9+ compatibility
I checked out the repo and ran `./gradlew build` in `ktlint-ruleset-template` to generate the jars. I then tried running ktlint with all of those jars against a kotlin class and received

``` java
Exception in thread "main" java.lang.ClassCastException: java.base/jdk.internal.loader.ClassLoaders$AppClassLoader cannot be cast to java.base/java.net.URLClassLoader
	at com.github.shyiko.ktlint.Main.loadJARs(Main.kt:567)
	at com.github.shyiko.ktlint.Main.main(Main.kt:253)
```

I can't see anything wrong with the jar structure or the paths I am using, but I'm guessing I have done something wrong.

The command I used was:
`ktlint -R build/libs/ktlint-ruleset-template.jar src/main/kotlin/yourpkgname/NoVarRule.kt`

I also tried using the sources jar.

Locitao:
Fixes #259 

Seems to have solved the issue with the class loader; created a separate one instead of hooking into the system one.

Testing by running the following command: 
```
 java -jar ktlint\build\libs\ktlint-all.jar -R ktlint-ruleset-template\build\libs\ktlint-ruleset-template.jar ktlint\src\main\kotlin\com\github\shyiko\ktlint\internal\MavenDependencyResolver.kt
```

Fix is as suggested by @shyiko

### DIFF
--- a/ktlint/src/main/kotlin/com/github/shyiko/ktlint/Main.kt
+++ b/ktlint/src/main/kotlin/com/github/shyiko/ktlint/Main.kt
@@ -30,6 +30,8 @@ import java.io.FileNotFoundException
 import java.io.IOException
 import java.io.PrintStream
 import java.math.BigInteger
+import java.net.URL
+import java.net.URLClassLoader
 import java.net.URLDecoder
 import java.nio.file.Path
 import java.nio.file.Paths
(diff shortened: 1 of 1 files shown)


shyiko:
I'm affraid this isn't going to work. https://github.com/shyiko/ktlint/blob/master/ktlint/src/main/kotlin/com/github/shyiko/ktlint/Main.kt#L272 knows nothing about newly constructed class loader (same as https://github.com/shyiko/ktlint/blob/master/ktlint/src/main/kotlin/com/github/shyiko/ktlint/Main.kt#L421).

Locitao:
Oh my bad, looked over that. I'll have another look tomorrow.

Issue #157, opened 2018-02-22
Respect @Suppress("RemoveCurlyBracesFromTemplate")


JLLeitschuh:
If I have some code like this:
```kotlin
val prettyString = computeSomeString()

@Suppress("RemoveCurlyBracesFromTemplate")
val expectedPrint = """
|${TAB}==========================================================
|${TAB}=                     GRAPH OVERVIEW                     =
|${TAB}==========================================================
|${TAB}Nodes: 2
|${TAB}Links: 2
|${TAB}
|${TAB}Link Breakdown:
|${TAB}${TAB}[Aout1->Bin1]; Weight: 1.0, Capacity: 10000.0
|${TAB}${TAB}[Bout1->Ain1]; Weight: 1.0, Capacity: 10000.0
""".trimMargin()

assertEquals(expectedPrint, prettyString)
```

Ktlint reformats it to this:
```kotlin
@Suppress("RemoveCurlyBracesFromTemplate")
val expectedPrint = """
|$TAB==========================================================
|$TAB=                     GRAPH OVERVIEW                     =
|$TAB==========================================================
|${TAB}Nodes: 2
|${TAB}Links: 2
|$TAB
|${TAB}Link Breakdown:
|${TAB}$TAB[Aout1->Bin1]; Weight: 1.0, Capacity: 10000.0
|${TAB}$TAB[Bout1->Ain1]; Weight: 1.0, Capacity: 10000.0
""".trimMargin()
```

The reason that I've put on the `@Suppress("RemoveCurlyBracesFromTemplate")` is because I want to have the "fake" indentation to be consistent. As you can see the character count between `$TAB` is off by two characters from `${TAB}`. I'm going for readability in my test over what is the "most correct".

AleksandrSl:
Can I try to implement this one?

shyiko:
@AleksandrSl That would be great! 
As a starting point - take a look at https://github.com/shyiko/ktlint/blob/master/ktlint-core/src/main/kotlin/com/github/shyiko/ktlint/core/KtLint.kt#L397. 

[...]

PR #263, opened 2018-07-30
Suppress remove curly braces from template #157

AleksandrSl:


AleksandrSl:
Hi! What about this one, should I modify something?

AleksandrSl:
Hi @shashachu, is this issue still needed to be done? If so I can resolve merge conflicts, otherwise I think it's time to close PR and issue)

shashachu:
Hi, sorry about that. Can you resolve the conflicts and then I or someone else here will take a look.

ktlint-ruleset-standard/src/test/resources/spec/string-template/lint.kt.spec
```diff
@@ -15,15 +15,44 @@ fun main() {
...
+    @Suppress("RemoveCurlyBracesFromTemplate", "Unused")
+    println("${s0}")
+    @Suppress("RemoveCurlyBracesFromTemplate")
+    val t = "${s0}"
 }
 
 class B(val k: String) {
     override fun toString(): String = "${super.toString()}, ${super.hashCode().toString()}, k=$k"
+
+    @Suppress("RemoveCurlyBracesFromTemplate", "Shit")
```
NightlyNexus: what is the second suppression here testing?

AleksandrSl: Sorry for this, seems like hard debug time
Cleaned

AleksandrSl:
@NightlyNexus I think all work is done, any other comments?

ktlint-ruleset-standard/src/test/resources/spec/string-template/lint.kt.spec
```diff
@@ -15,15 +15,45 @@ fun main() {
...
+    val a
+        get() = "${s0}"
+}
+
+@Suppress("RemoveCurlyBracesFromTemplate")
+class C {
+    override fun toString(): String = "${s0}"
+}
+
+// Ensure that suppression scope is as wider as it should be
```
NightlyNexus: typo: `is as wide as it`

AleksandrSl: Thanks, fixed

AleksandrSl:
Hi, sorry to bother you again) Is there a reason that prevents PR from being merged?

shashachu:
> Hi, sorry to bother you again) Is there a reason that prevents PR from being merged?

Hi! No, this PR just fell through the cracks.

shashachu:
> > Hi, sorry to bother you again) Is there a reason that prevents PR from being merged?
> 
> Hi! No, this PR just fell through the cracks.

Can you rebase and just make sure the build still passes? I can merge after.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -490,6 +515,12 @@ object KtLint {
                 comment.replace(Regex("\\s"), " ").replace(" {2,}", " ").split(" ")
 
             private fun <T> List<T>.tail() = this.subList(1, this.size)
+
+            private val suppressAnnotationRuleMap = mapOf(
```
AleksandrSl: I think this map could be moved to some config file, to ease addition of new suppression 
 rules

AleksandrSl:
Done!

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -466,6 +470,27 @@ object KtLint {
...
+                                if (suppressedRules.isNotEmpty()) {
+                                    result.add(
+                                        SuppressionHint(
+                                            IntRange(psi.startOffset, psi.endOffset),
+                                            suppressedRules.toSet()
+                                        )
+                                    )
+                                }
+                            }
+                    }
```
Tapchicoma: nit: could you move this to some private method for better readability?

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
Tapchicoma: I am not sure about having this mapping:
- "RemoveCurlyBracesFromTemplate" just becomes an alias to "string-template" rule
- This alias is detached from the rule itself
- Overall this PR goes into direction of adding another way for [suppress rules](https://github.com/pinterest/ktlint#how-do-i-suppress-an-error-for-a-lineblockfile)

To be consistent, I would propose to use following approach:
```kotlin
@Suppress("ktlint-string-template")
```
to disable rule for the block of code and remove this map.

AleksandrSl: I agree that this map looks weird, however original purpose for this PR was to respect annotations like `@Suppress("RemoveCurlyBracesFromTemplate")` to disable *both* IntellijIDEA and ktlint checks at the same time. Otherwise we will have to add both annotations
```
@Suppress("ktlint-string-template")
@Suppress("RemoveCurlyBracesFromTemplate")
println("Not found: ${notFound}")
```

and this way all sense in introduction of support for `@Suppress` annotation seems lost to me, cause it becomes specific for ktlint, and as you mentioned we already have ways to suppress ktlint checks

Tapchicoma: > I agree that this map looks weird, however original purpose for this PR was to respect annotations like @Suppress("RemoveCurlyBracesFromTemplate") to disable both IntellijIDEA and ktlint checks at the same time.

Sorry, missed that. Then, as intermediate step, having this map looks ok. In the future would be better that rule provides a list of suppress names aliases.

PR #271, opened 2018-08-21
Issue#238 Check directory structure

yokotaso:
There are rules about package name:
- https://kotlinlang.org/docs/reference/coding-conventions.html#naming-rules
- https://android.github.io/kotlin-guides/style.html#package-names

 if we will check it, Backword compatibility is missing. So, This rule check structure of directory.

See also:
* https://kotlinlang.org/docs/reference/coding-conventions.html#directory-structure
* https://github.com/shyiko/ktlint/pull/246#issuecomment-414248741


shyiko:
Closed in favor of #246.

PR #283, opened 2018-09-20
Fix for extra newlines when using --stdin

bethcutler:
https://github.com/shyiko/ktlint/issues/282


shyiko:
Thank you :bowing_man:

PR #291, opened 2018-10-10
Update ktlint-gradle plugin info in README.

Tapchicoma:
"The very first ktlint gradle plugin" doesn't reflect what plugin actually supports and do.


shyiko:
Thank you! :+1:

PR #301, opened 2018-10-23
Fixing issue with CommentSpacingRule which would insert a space into a comment that was all forward slashes

shashachu-pinterest:
In our code base, we have some block comments in the form of:

```
////////////////
// functionName
////////////////
```
Previously, ktlint would insert spaces, turning that block into:
```
// //////////////
// functionName
// //////////////
```

This PR fixes that.


shyiko:
:bowing_man:

shashachu-pinterest:
🎉

PR #304, opened 2018-10-26
Doc: bind ktlint to maven validate phase in example

yuppie-flu:
The common approach is to do code style checks before compilation and tests execution. In maven lifecycle it means `validate` phase, not `verify`. Some people just copy-paste maven configuration example from README, so it's better to update it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ $ ktlint --install-git-pre-commit-hook
     <executions>
         <execution>
             <id>ktlint</id>
-            <phase>verify</phase>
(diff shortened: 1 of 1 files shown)


shyiko:
Personally, I'd argue that running ktlint before code compilation takes place is a waste of time. When I'm developing something I don't care about the code style until I'm ready to commit (all I want is to make the code work). The choice of "verify" instead of "validate" was intentional. 
I've added a comment to the README.md per your suggestion, though, so thank you!

PR #305, opened 2018-10-28
Update README.md to remove extra spaces in installation command

sanogueralorenzo:
Made installation command a single line to avoid extra spaces.


sanogueralorenzo:
@shyiko feel free to merge or if a new version is coming soon you could do that in the same PR (removing extra spaces)

PR #336, opened 2019-02-03
fix brew install step

mapyo:
Thank you for this repository.

I fixed it because I needed tap when running brew install.


shyiko:
🙇

mapyo:
thanks!

PR #339, opened 2019-02-11
Add missing rules to ruleset

rom4ek:
I've just noticed that the latest release was missing the latest added rules in the standard ruleset. I've added them, but you'd probably need to release again, right?


shyiko:
🤦‍♂️ Thank you, @rom4ek!

Issue #341, opened 2019-02-14
Opening brace on new line is permitted on a function w/ return type


choey:
Affects `ktlint:0.30.0`.

Consider the following:
```
fun foo()
{
}
```
Lint fails due to unexpected new line before `{`.

But if the function has a non-Unit return type, like
```
fun foo(): String
{
    return ""
}
```
then lint does not fail.

Here's what the [official Kotlin coding conventions doc](https://kotlinlang.org/docs/reference/coding-conventions.html#formatting) says about the opening brace:
> [f]or curly braces, put the opening brace in the end of the line where the construct begins

The only scenario (called out by this doc) in which `{` starts on a new line is when class header is too long.

So, I'm wondering whether the ktlint rule should be tighter around `{` on function declaration and fail on the second example above.

PR #342, opened 2019-02-15
Disallow newline between function return type and LBRACE

choey:
Addresses #341.


shyiko:
I'm afraid checking for INDENTIFIER is not enough:
```kotlin
val f: () -> () -> Unit = {
    val x = 1
    val y = x
    {
        println(y)
    }
}
```

shashachu:
Fixed in https://github.com/pinterest/ktlint/pull/360 so closing this.

PR #348, opened 2019-02-21
FilenameRule: Ignore kotlin backtick escape character on class name

antonydenyer:
- validates file name from a class name that is escaped.


shyiko:
Thank you, @antonydenyer! 🙇

PR #351, opened 2019-03-10
Java 9+ compatibility

### LINKED ISSUE #259: --reporter/--ruleset Java 9+ compatibility

marschwar:
This PR attempts to resolve #259.

It looks like much more than it actually is. Essentially I changed.

```
(ClassLoader.getSystemClassLoader() as java.net.URLClassLoader)
    .addURLs(artifactUrls)
```
to
```
URLClassLoader(artifactUrls.toTypedArray(), parentClassloader)
```

All the other changes are basically refactoring to be able to write a test for it. I tried to make small commits so there is a chance for someone else to review the changes. Unfortunately I find it really hard to make sure I did not break anything.

One more thing: In order to test that a custom ruleset is downloaded via maven, I added the first one I could find (https://github.com/gabrielittner/ktlint-rules). This obviously is not ideal. I hope someone else has a better idea.


.appveyor.yml
```diff
@@ -5,6 +5,7 @@ skip_tags: true
 environment:
```
jaredsburrows: This file is no longer in master. Can you remove this file?

jaredsburrows:
If you are attempting to solve #259, can you remove changes that are unrelated such as upgrading gradle?

marschwar:
Sure. As there was no feedback up until now I did not bother to rebase onto master. If you think it is going in the right direction, I would be happy to do so. If not I would close the PR.

jaredsburrows:
Usually, larger the PRs or greater the impact may take more time. Keeping the change small and specific will help code reviews go faster :).

jacquerie:
According to https://blog.codefx.org/java/java-9-migration-guide/#Casting-To-URL-Class-Loader, this is not completely straightforward:
> If you’ve used the URLClassLoader to dynamically load user provided code (for example as part of a plugin infrastructure) by appending to the class path, then you have to find a new way to do that as it can not be done with Java 9.

So, perhaps, some additional complexity introduced by this PR might be warranted. I can take care of rebasing and removing extraneous stuff such as style changes and version bumps!

PR #360, opened 2019-03-27
Make new line after function return type declaration fail.

### LINKED ISSUE #341: Opening brace on new line is permitted on a function w/ return type

NightlyNexus:
Fixes #341.


rom4ek:
@vanniktech it's working already, isn't it? You can see it in the test

shashachu:
Thanks, Eric!

PR #363, opened 2019-03-30
Fix failing on KDoc semicolon after other nodes.

### LINKED ISSUE #362: Semicolons (at end of line) in KDoc aren't allowed on lines containing references, but should be
Ktlint doesn't accept the following:

```kotlin
/**
 * [x];
 */
```

Ktlint doesn't allow semicolons at the end of a KDoc line. However, they are perfectly fine, as they are a natural language token, not a Kotlin token.

It seems this only happens when the KDoc contains references to code, e.g..: `[x]`.
Note that just formatted code with a semicolon, e.g. `` `x`;``, is accepted by ktlint.

The suggestion is to accept EOL semicolons in KDoc lines that contain Kotlin code.

NightlyNexus:
```
/**
 * x;
 */
```
was working fine because the text was processed as "x;" but
```
/**
 * [x];
 */
```
failed because the left bracket, the identifier, and the right racket were processed as separate nodes, and then the node with the single ";" text was checked and failed.
Simply opting out of the check for all KDOC_TEXT seems fine.

Fixes #362.


erikhuizinga:
Thanks!

PR #364, opened 2019-03-31
Fix 'Missing newline before ")"' when no params

### LINKED ISSUE #327: 'Missing newline before ")"' in long function names
In ktlint version 0.29.0, there is an issue when a function name is too long. It's not common in normal functions, but it's when writing test functions using spaces. Example:
```kotlin
fun `any long function with more than 120 chars should only break 'Exceeded max line length (120)' and not any other rule`() {
   ...
}
```
When analyzing the code with ktlint, it says there are two broken rules:
`Unexpected spacing after "("`
`Exceeded max line length (120)`

In this case, the only broken rule should be `Exceeded max line length (120)`, since adding a line break after the parenthesis when there are no parameters I think it does not make sense.

It was not happening in previous versions.

jlmd:
Fix an issue where the `Exceeded max line length` rule is always trying to put parameters on separate lines, even when there are no parameters, causing the rule `Missing newline before ")"` to break.

Fixes #327

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ParameterListWrappingRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/ParameterListWrappingRule.kt
@@ -39,6 +39,8 @@ class ParameterListWrappingRule : Rule("parameter-list-wrapping") {
             return
         }
         if (node.elementType == VALUE_PARAMETER_LIST &&
+            // skip when there are no parameters
+            node.firstChildNode?.treeNext?.elementType != RPAR &&
             // skip lambda parameters
             node.treeParent?.elementType != FUNCTION_LITERAL
         ) {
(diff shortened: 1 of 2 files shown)


shashachu:
@jlmd just to confirm - you've tested this is still broken on 0.30.0? @shyiko addressed #327 for that release, so just want to make sure this is an update to that commit.

jlmd:
@shashachu  yes, I tested it on 0.30.0 and is still broken. I'm not sure what the fix of @shyiko  does, but for sure it doesn't fix #327 issue. Maybe the issue was not clear enough, that's why I created the PR.

Do you think it makes sense to revert in this PR that commit (b6f9ad50366315604b1f24742bdc900867d96566), since it seems that it doesn't really fix any issue? Even the tests that he added pass by removing the condition that he added.

shashachu:
Let's wait a day or so to see if @shyiko responds about what his original change was meant to do.

shyiko:
@jlmd is correct (b6f9ad5 should be reverted)

shashachu:
Awesome thanks for confirming, @shyiko. @jlmd why don't you also revert that commit then.

jlmd:
Cool! Thanks @shashachu and @shyiko . I've reverted that commit in https://github.com/pinterest/ktlint/pull/364/commits/c24eb5ad013da1b5e470ccfd6cd314e1ba7528d5

shashachu:
Thanks for the PR!

PR #381, opened 2019-04-17
Updating packages from com.github.shyiko -> com.pinterest

shashachu:
Just code packages for now - no external groupId changes.


shashachu:
meh.

PR #385, opened 2019-04-19
Update groupId to com.github.pinterest

shashachu:


shashachu:
boo. i'm terrible at git.

PR #386, opened 2019-04-19
Update groupId to com.pinterest.ktlint

shashachu:
* Also updated various `shyiko` -> `pinterest` refs

### DIFF
--- a/ktlint-core/pom.xml
+++ b/ktlint-core/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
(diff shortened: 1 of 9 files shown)


JakeWharton:
Why retain 'github'? You own pinterest.com, right?

shashachu:
> Why retain 'github'? You own pinterest.com, right?

The short answer is 'weird historical reasons' but double-checking about it here.

shashachu:
Snapshot appears to have successfully deployed: https://oss.sonatype.org/#nexus-search;gav~com.pinterest~ktlint~~~~kw,versionexpand

Issue #368, opened 2019-04-04
IndentationRule inserting unnecessary line break after assignment followed by any continuation


shashachu:
```
$ echo $'fun test() {
    val variable = first
        ?: second
}' > test.kt

$ ./ktlint -F --experimental test.kt
```

Expected:
```
fun test() {
    val variable = first
        ?: second
}
```

Actual:
```
fun test() {
    val variable =
        first
            ?: second
}
```

Seems to have something to do with the elvis operator; I haven't gotten it to break otherwise. Also works fine if the elvis operator is on the same line. Also, 
```
fun test() {
    val variable = first
        second
}
```
produces 
```
fun test() {
    val variable = first
    second
}
```
as expected.

shashachu:
Also occurs with:
```
$ echo $'fun test() {
    val variable = first
        ?: second
}' > test.kt

$ ./ktlint -F --experimental test.kt
```

Output
```
fun test() {
    val variable =
        first
            .function()
}
```

Issue #380, opened 2019-04-17
MissingNewLine triggered on builder chain like pattern


vanniktech:
When using the `--experimental` flag with ktlint 0.31.0 and the following configuration:

```
[*.{kt,kts}]
indent_size=2
max_line_length=200
```

the following code is flagged:

```kotlin
    val shareEvent = ShareEvent()
        .putMethod(event.method)
        .putContentName(event.contentName)
        .putContentType(event.contentType)
```

This is the output:

`Missing newline after "="`

My code is perfectly intended. I don't know why I'd want to have the `ShareEvent()` on its own line.

shashachu:
I think your first bug is the same one as this one https://github.com/pinterest/ktlint/issues/368 - i didn't know how to describe it properly, but an assignment followed by some other statements seems to trigger it.

shashachu:
@vanniktech I'm going to close this one since it's a dupe of #368 - feel free to open a separate ticket for the parentheses case. I'm not sure yet whether these all have the same root cause or not.

vanniktech:
Thanks. Once that other issue has a PR / fix, I can try it out again.

PR #387, opened 2019-04-19
Remove code that was inserting newline after = in a chained call

shashachu:
* Fixes https://github.com/pinterest/ktlint/issues/368 and https://github.com/pinterest/ktlint/issues/380
* Kotlin styleguide: https://kotlinlang.org/docs/reference/coding-conventions.html#chained-call-wrapping
  * Says 'The first call in the chain usually should have a line break before it, but it's OK to omit it if the code makes more sense that way.'

### DIFF
--- a/ktlint-ruleset-experimental/src/test/kotlin/com/pinterest/ktlint/ruleset/experimental/IndentationRuleTest.kt
+++ b/ktlint-ruleset-experimental/src/test/kotlin/com/pinterest/ktlint/ruleset/experimental/IndentationRuleTest.kt
@@ -285,7 +285,7 @@ class IndentationRuleTest {
     }
 
     @Test
-    fun testLintNewlineAfterEqALlowed() {
+    fun testLintNewlineAfterEqAllowed() {
         assertThat(
             IndentationRule().lint(
                 // Previously the IndentationRule would force the line break after the `=`. Verify that it is
--- a/ktlint-ruleset-experimental/src/test/kotlin/com/pinterest/ktlint/ruleset/experimental/IndentationRuleTest.kt
+++ b/ktlint-ruleset-experimental/src/test/kotlin/com/pinterest/ktlint/ruleset/experimental/IndentationRuleTest.kt
@@ -283,4 +283,22 @@ class IndentationRuleTest {
         assertThat(IndentationRule().format(ktScript, mapOf("indent_size" to "2")))
             .isEqualTo("fun main() {\n  return 0\n}")
     }
+
+    @Test
+    fun testLintNewlineAfterEqALlowed() {
+        assertThat(
+            IndentationRule().lint(
+                // Previously the IndentationRule would force the line break after the `=`. Verify that it is
+                // still allowed.
+                """
+                private fun getImplementationVersion() =
+                    javaClass.`package`.implementationVersion
+                        ?: javaClass.getResourceAsStream("/META-INF/MANIFEST.MF")
+                            ?.let { stream ->
+                                Manifest(stream).mainAttributes.getValue("Implementation-Version")
+                            }
+                """.trimIndent()
+            )
+        ).isEmpty()
+    }
 }
--- a/ktlint-ruleset-experimental/src/test/resources/spec/indent/format-dot-qualified-expression-expected.kt.spec
+++ b/ktlint-ruleset-experimental/src/test/resources/spec/indent/format-dot-qualified-expression-expected.kt.spec
@@ -1,10 +1,9 @@
 class A {
-    private fun getImplementationVersion() =
-        javaClass.`package`.implementationVersion
-            ?: javaClass.getResourceAsStream("/META-INF/MANIFEST.MF")
-                ?.let { stream ->
-                    Manifest(stream).mainAttributes.getValue("Implementation-Version")
-                }
+    private fun getImplementationVersion() = javaClass.`package`.implementationVersion
+        ?: javaClass.getResourceAsStream("/META-INF/MANIFEST.MF")
+            ?.let { stream ->
+                Manifest(stream).mainAttributes.getValue("Implementation-Version")
+            }
 
     fun f() {
         x()?.apply {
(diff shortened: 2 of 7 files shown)


shashachu:
@vanniktech thoughts on this one? I just removed the code that was inserting the newline after the `=` because it felt more like a bug than a feature.

ktlint-ruleset-experimental/src/test/resources/spec/indent/format-dot-qualified-expression-expected.kt.spec
```diff
@@ -1,10 +1,9 @@
 class A {
-    private fun getImplementationVersion() =
-        javaClass.`package`.implementationVersion
-            ?: javaClass.getResourceAsStream("/META-INF/MANIFEST.MF")
-                ?.let { stream ->
-                    Manifest(stream).mainAttributes.getValue("Implementation-Version")
-                }
+    private fun getImplementationVersion() = javaClass.`package`.implementationVersion
```
JLLeitschuh: For functions like this, I actually think I like that Ktlint forces you to put your function body starting on a new line. It reads better.

ktlint-ruleset-experimental/src/test/resources/spec/indent/format-raw-string-trim-indent-expected.kt.spec
```diff
@@ -87,11 +87,10 @@ fun f() {
...
 class C {
-    val CONFIG_COMPACT =
-        """
-        {
-        }
-        """.trimIndent()
+    val CONFIG_COMPACT = """
+    {
+    }
+    """.trimIndent()
```
JLLeitschuh: This kinda makes sense and kinda doesn't.
The consistent requirement on the `"""` being lined up would actually be better.

ktlint-ruleset-experimental/src/test/resources/spec/indent/format-dot-qualified-expression-expected.kt.spec
shashachu: Yeah in this case I'm sort of +/- on the change. Really the case that seems broken is:
```
val foo = someClass
    .aFunction()
    .anotherFunction()
```

becomes
```
val foo = 
    someClass
        .aFunction()
        .anotherFunction()
```

So maybe this is too heavy a hammer.

shashachu: @JLLeitschuh one thing to remember is that with this PR, ktlint won't disallow the code as it was originally formatted, it just won't force it.

shashachu:
I re-introduced the rule for raw strings but removed the rest. I also added a test verifying the linebreak after = in chained calls is still allowed, just not forced.

ktlint-ruleset-experimental/src/test/kotlin/com/pinterest/ktlint/ruleset/experimental/IndentationRuleTest.kt
```diff
@@ -283,4 +283,22 @@ class IndentationRuleTest {
         assertThat(IndentationRule().format(ktScript, mapOf("indent_size" to "2")))
             .isEqualTo("fun main() {\n  return 0\n}")
     }
+
+    @Test
+    fun testLintNewlineAfterEqALlowed() {
```
NightlyNexus: accidentally capitalized L? `Allowed`

ktlint-ruleset-experimental/src/test/resources/spec/indent/format-eq-expected.kt.spec
```diff
@@ -5,21 +5,19 @@ class C {
 }
 
 fun f() {
-    val x =
-        "a" +
-            "b2"
+    val x = "a" +
```
NightlyNexus: i don't know if this better...

ktlint-ruleset-experimental/src/test/resources/spec/indent/format-eq-expected.kt.spec
```diff
@@ -5,21 +5,19 @@ class C {
...
 
-    val x =
-        paths.flatMap { dir ->
-            "hello"
-        } + f0(
-            "there"
-        ) + f1(
-            "sssss"
-        )
+    val x = paths.flatMap { dir ->
```
NightlyNexus: this does seem slightly better, though

ktlint-ruleset-experimental/src/test/resources/spec/indent/format-property-accessor-expected.kt.spec
```diff
@@ -1,9 +1,8 @@
 class C {
 
     private val Any.className
-        get() =
-            this.javaClass.name
-                .fn()
+        get() = this.javaClass.name
```
NightlyNexus: am also unsure here.

ktlint-ruleset-experimental/src/test/kotlin/com/pinterest/ktlint/ruleset/experimental/IndentationRuleTest.kt
shashachu: fixed.

ktlint-ruleset-experimental/src/test/resources/spec/indent/format-eq-expected.kt.spec
shashachu: really? it seems better to me. but the previous formatting is still allowed, just not forced.

ktlint-ruleset-experimental/src/test/resources/spec/indent/format-property-accessor-expected.kt.spec
shashachu: both look bad to me, but the extra newline irks me. maybe it's just me. :)

PR #389, opened 2019-04-20
fix SpacingAroundCommaRule when comma goes after comment

### LINKED ISSUE #367: Comma on newline preceded by // comment will move code into comment
I got a when phrase. The case is

before
```java
when (type) {
            IHomeLivingConstant.LIVING_TYPE_GUESS_YOU_LIKE //a
                , IHomeLivingConstant.Live_MODULE_CONTENT_CODE // b
                , IHomeLivingConstant.LIVE_MODULE_AUTO_PLAY  //c
-> {
    // do something
}
```

after formatting, I turn out
```java
when (type) {
            IHomeLivingConstant.LIVING_TYPE_GUESS_YOU_LIKE //a, IHomeLivingConstant.Live_MODULE_CONTENT_CODE // b, IHomeLivingConstant.LIVE_MODULE_AUTO_PLAY  //c
-> {
    //do something
} 

```

This lead to bugs cause it will not invoke compile error, but is wrong logically.  And it is hard to find the bug in the hundreds of formatted files.

z3d1k:
Disable autocorrection for comma spacing rule if comma on newline preceded by // comment.

Fixes #367

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundCommaRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundCommaRule.kt
@@ -3,6 +3,8 @@ package com.pinterest.ktlint.ruleset.standard
 import com.pinterest.ktlint.core.Rule
 import com.pinterest.ktlint.core.ast.isPartOfString
 import com.pinterest.ktlint.core.ast.nextLeaf
+import com.pinterest.ktlint.core.ast.nextSibling
+import com.pinterest.ktlint.core.ast.prevCodeLeaf
 import com.pinterest.ktlint.core.ast.prevLeaf
 import com.pinterest.ktlint.core.ast.upsertWhitespaceAfterMe
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundCommaRule.kt
```diff
@@ -19,9 +20,10 @@ class SpacingAroundCommaRule : Rule("comma-spacing") {
         if (node is LeafPsiElement && node.textMatches(",") && !node.isPartOfString()) {
             val prevLeaf = node.prevLeaf()
             if (prevLeaf is PsiWhiteSpace) {
-                emit(prevLeaf.startOffset, "Unexpected spacing before \"${node.text}\"", true)
-                if (autoCorrect) {
-                    prevLeaf.node.treeParent.removeChild(prevLeaf.node)
+                val canBeAutoCorrected = prevLeaf.prevLeaf { it !is PsiWhiteSpace } !is PsiComment
+                emit(prevLeaf.startOffset, "Unexpected spacing before \"${node.text}\"", canBeAutoCorrected)
```
shashachu: Since this is a weird edge case, might be nice to add some more info about why it can't be auto corrected.

z3d1k: I added a comment with link to a related issue, is that enough?

shashachu: @z3d1k in-code comments are great but I actually meant the text output by ktlint. Most people will probably be using ktlint as a tool and not looking at the code, and since *most* of the time, this rule can be auto-corrected, it might look like a bug that it doesn't autocorrect in this case. Thoughts?

JLLeitschuh: > in-code comments are great but I actually meant the text output by ktlint.
I agree with this sentiment.

z3d1k:
@shashachu I managed to make auto correction work, but ran into another comment related case.
This code 
`fun main() { x(1 /* comment */ , 3); x(1 /* comment */, 3) }`
 will be formatted by current version as follows
`fun main() { x(1 /* comment */, 3); x(1 /* comment */, 3) }`
After this fix it became
`fun main() { x(1, /* comment */ 3); x(1 /* comment */, 3) }`
Is this behavior correct, or it should be left as it is now?

Also, maybe you can come up with other scenarios where auto correction might not work as expected in this rule?

shashachu:
> @shashachu I managed to make auto correction work, but ran into another comment related case.
> This code
> `fun main() { x(1 /* comment */ , 3); x(1 /* comment */, 3) }`
> will be formatted by current version as follows
> `fun main() { x(1 /* comment */, 3); x(1 /* comment */, 3) }`
> After this fix it became
> `fun main() { x(1, /* comment */ 3); x(1 /* comment */, 3) }`
> Is this behavior correct, or it should be left as it is now?

Sorry for the late reply. That doesn't look correct, as the intended behavior is to attach the comment to the first parameter, not the second.

z3d1k:
I agree. My last commit fixed it, now the comma is only moved before the comment in case if comma is on a new line. Check out the tests.

shashachu:
Sorry for the delay - looks good to me. Thanks!

PR #391, opened 2019-04-22
Updating README refs to 0.32.0

shashachu:
`.announce` script not working; manually updating.


shashachu:
> https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-template/build.gradle looks like it also needs updating

Ah, good catch. Fixed.

Issue #392, opened 2019-04-23
Travis snapshot build broken


shashachu:
Need to update encrypted environment variables

shashachu:
Fixed: https://travis-ci.org/pinterest/ktlint/builds/523293666

PR #393, opened 2019-04-23
Updating env variables for travis snapshot build

shashachu:


shashachu:
Should fix https://github.com/pinterest/ktlint/issues/392

PR #416, opened 2019-05-10
Added Freeletics to Adopters.md

sockeqwe:
Thanks for taking ownership and for your commitment to this project. We at Freeletics use and love Ktlint to and have with @gabrielittner and @Tapchicoma already two contributors and would like to continue do that in the future.


shashachu:
No need to leave Pinterest at the top - can you add it in Alphabetical order? Thanks!

sockeqwe:
Thanks. I changed that.

PR #432, opened 2019-05-22
Implementing the concept of Issue Types

RodrigoAmaroEng:
After i start using ktlint i was amazed how easy it was to write my own rules.
Since my team was planing to use it on our PR build pipeline the tool would be perfect for us to enforce all the constraints that we agree to follow.

But, given that some agreements can be very aggressive to be enforce suddenly, we decided that they should appear as Warning only until we could put it for real.

Searching through the code base i wasn't able to find any configuration for this so i decided to customize it and it's working very well for us.

**So how does it works?**

I had to change the way we emit the problems encountered on the code evaluation. Instead of emitting line, col and ruleId we now have to emit an Issue. It contains all three mentioned parameters plus an IssueType parameter (defaults to Error) that can be used to inform if it's an Error or a Warning.

All the code was refactored to emit Issues and the reporter classes now can also provide this information. For example, when using plain reporter we receive this:

```
/path/to/file
[ERROR] 12:3 Issue description
```

I believe in the future it can be improved to allow the user to configure which rules should be treated as Warnings or Erros, or even create new types of Issue

### DIFF
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/Issue.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/Issue.kt
@@ -1,0 +1,19 @@
+package com.pinterest.ktlint.core
+
+data class Issue(
+    val offset: Int,
+    val errorMessage: String,
+    val canBeAutoCorrected: Boolean,
+    val type: IssueType = IssueType.ERROR
+) {
+    fun toLintIssue(line: Int, col: Int, ruleId: String): LintIssue {
+        return when (type) {
+            IssueType.ERROR -> LintError(line, col, ruleId, errorMessage, canBeAutoCorrected)
+            IssueType.WARNING -> LintWarning(line, col, ruleId, errorMessage, canBeAutoCorrected)
+        }
+    }
+}
+
+enum class IssueType {
+    ERROR, WARNING
+}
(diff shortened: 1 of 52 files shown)


shashachu:
This is a really interesting idea, Rodrigo. Can you give an outline of how you're thinking it would extended to allow user configuration of the rules? I only took a quick look but it looks like right now it requires changing the source of the rules.

RodrigoAmaroEng:
Indeed, i had to rewrite a small part for all rules embedded on the library. I understand it's a major change that breaks compatibility since every rule that 3rd parties have written must be fix to work with this approach. 

Truth be told, it was a first attempt to write something that would provide the features our team needed and at that moment it wasn't a concern although i know that putting this on the baseline would be a problem to many people.

On the other hand, i do think that using a class (Issue) instead of parameters (line, col and ruleId) would be major improvement for the library since it helps on future releases and adds a layer to prevent that changes like this to happen in a near future. (Of course it's only true for increments over the Issue class)

Maybe i could add a 4th parameter with a default value instead of this class but i sense that this is only postponing the problem

akuleshov7:
Good Idea, but actually I think that for static analysis there should not be difference between errors and warnings. It is useful for compilers, but is not useful for static analyzers. If analysis finds an issue - it simply should be reported and that's all. I do not think that it is really useful in `ktlint`.

If I can understand you correctly - your particular problem is that you are having too many errors and you are not able to fix all of them right now. We had the following issues, that's why we made our rule set configurable and added suppress mechanism:
https://github.com/cqfn/diKTat#customizations-via-diktat-analysisyml

shashachu:
I'm going to close this PR now that we have the concept of baselines and we have disable-able rules. I believe some combination of those should solve your issue.

Issue #433, opened 2019-05-22
"Missing newline after "("" triggered with parameter comments


bethcutler:
I've enabled the --experimental flag on ktlint 0.32, and now we are seeing this error when there are comments on parameters, e.g.:

```
foo(
  /*param1=*/ param1,
  /*param2=*/ param2
)
```

If foo() is defined in Kotlin, we can fix the error by changing the comments to parameter names.  However, if foo() is defined in Java, there seems to be no way to name the parameters without triggering this erroneous message.

shashachu:
I think I see the issue. I'll put up a PR soon.

bethcutler:
Thanks for the quick fix!

PR #434, opened 2019-05-22
Fixing issue where ktlint would erroneously want to insert a newline if parameter comments were used

shashachu:
Fixes https://github.com/pinterest/ktlint/issues/433


ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/IndentationRule.kt
```diff
@@ -208,11 +208,11 @@ class IndentationRule : Rule("indent"), Rule.Modifier.RestrictToRootLast {
             return
         }
         if (!node.nextCodeLeaf()?.prevLeaf {
-                // Skip comments, whitespace, and empty nodes
-                !it.isPartOfComment() &&
+            // Skip comments, whitespace, and empty nodes
+            !it.isPartOfComment() &&
                 !it.isWhiteSpaceWithoutNewline() &&
```
shashachu: This formatting looks weird to me, but IntelliJ does the same thing.

NightlyNexus: looks right; it's a continuation indent.

Test.kt
```diff
@@ -0,0 +1,11 @@
+fun main() {
```
shashachu: doh i did not mean to check this in.

Issue #405, opened 2019-04-26
Unused import is not reported


Ilya-Gh:
Unused import a of static function is not reported if the function is used referenced with its parent class.

Android Studio/IDea does report the import as unnecessary using [KotlinUnusedImportInspection](https://github.com/JetBrains/kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/inspections/KotlinUnusedImportInspection.kt):

![image](https://user-images.githubusercontent.com/10222326/56818066-8edc0600-684f-11e9-8f6f-d042be96d2f3.png)

The [following test added to `NoUnusedImportsRuleTest`](https://github.com/Ilya-Gh/ktlint/commit/eeb77aef20e82f57ca309d66d674f06efaed0e8d#diff-2b4cd66b8a3a6ecff83990f820622d1dR196) currently [fails](https://travis-ci.org/Ilya-Gh/ktlint/builds/524993641):

```kotlin
@Test
fun testParentPackImport() {
    assertThat(
        NoUnusedImportsRule().lint(
            """
            import A.b

            fun main() {
                A.b
            }
            """.trimIndent()
        )
    ).isEqualTo(
        listOf(
            LintError(1, 1, "no-unused-imports", "Unused import")
        )
    )
}
```
The error:
```
java.lang.AssertionError: 
Expecting:
 <[]>
to be equal to:
 <[LintError(line=1, col=1, ruleId=no-unused-imports, detail=Unused import)]>
but was not.
	at com.pinterest.ktlint.ruleset.standard.NoUnusedImportsRuleTest.testParentPackImport(NoUnusedImportsRuleTest.kt:208)
```

shashachu:
Re-opening because I reverted the fix, as it caused a couple regressions with false-positives.

PR #437, opened 2019-05-25
Report unused import when parent reference is already present

sowmyav24:
This PR fixes the issue [#405](https://github.com/pinterest/ktlint/issues/405)

- While traversing nodes and building `ref`,  built a set of non redundant imports which omits the imports where parent reference is already present.
- Imports with aliases should contain both name and alias in the importSet
- Static function call and variable have been distinguised


shashachu:
@sowmyav24 thanks for the PR! A couple requests:
1) would be great to add a few comments into the code - we're trying to slowly add comments into the standard ruleset since they make great examples for others trying to write their own custom rules
2) I don't know why it wouldn't work, but can you add some tests to ensure auto-formatting works properly for these unused parent imports?

sowmyav24:
@shashachu Thank you for the feedback.
- Added a test to ensure auto formatting works for parent imports.
- Added comments for the functions introduced. Let me know if needs to be more descriptive.

shashachu:
Thanks for the fixes!

PR #441, opened 2019-05-30
Don't autoformat imports if the import list contains top-level comments

### LINKED ISSUE #408: ktlint-disable removed from imports after auto-formatting
/* ktlint-disable .... */ 
is removed after auto-formating

shashachu:
Fixes #408 

Refusing to auto-import seemed much easier than trying to figure out which imports the comments were attached to.

### DIFF
--- a/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ImportOrderingRule.kt
+++ b/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ImportOrderingRule.kt
@@ -1,6 +1,8 @@
 package com.pinterest.ktlint.ruleset.experimental
 
 import com.pinterest.ktlint.core.Rule
+import com.pinterest.ktlint.core.ast.ElementType.BLOCK_COMMENT
+import com.pinterest.ktlint.core.ast.ElementType.EOL_COMMENT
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ImportOrderingRule.kt
```diff
@@ -23,10 +25,17 @@ class ImportOrderingRule : Rule("import-ordering") {
             val children = node.getChildren(null)
             if (children.isNotEmpty()) {
                 val imports = children.filter { it.elementType == IMPORT_DIRECTIVE }
+                val hasComments = children.find { it.elementType == BLOCK_COMMENT || it.elementType == EOL_COMMENT } != null
```
NightlyNexus: a broad brush. should we check for the suppression message? though perhaps unlikely to be a problem in practice.

shashachu: The suppression message causes a real problem, but it would also seem annoying if you had comments (for some reason?) in an import list that they'd get blown away by the autoformatter.

NightlyNexus: true. sounds good.

Issue #443, opened 2019-05-30
Enforce expression body for single-line methods


bbaldino:
In the Kotlin style guide, it says to prefer expression body style for single-line methods:

![image](https://user-images.githubusercontent.com/4016469/58663336-20bcb000-82e1-11e9-813f-c258fa74420c.png)

So thought it would be nice to have a rule for it.  It does say "prefer", so perhaps this was omitted because there are legitimate cases to avoid it, but wanted to ask about it.  Was a rule for this left out intentionally?

rom4ek:
Would definitely fit experimental ruleset I think

shashachu:
I don't know if it was left out intentionally, but the Android style guide has the same recommendation: https://developer.android.com/kotlin/style-guide#expression_functions

Feel free to open a PR.

shashachu:
Since this is a suggestion rather than a rule, I'm going to close this.

PR #444, opened 2019-05-31
add a rule to enforce single-expression methods use an expression body

bbaldino:
This is for https://github.com/pinterest/ktlint/issues/443.

I fully expect this code is not yet acceptable, but I wanted to get some feedback.  At the very least, I think this rule could be auto-correctable, but I've held off on doing that part until I get some feedback on the approach.


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/MethodExpressionBodyRule.kt
```diff
@@ -0,0 +1,36 @@
...
+ * Prefer using an expression body for functions with the body consisting of a single expression
+ */
+class MethodExpressionBodyRule : Rule("method-expression-body") {
+
+    override fun visit(node: ASTNode, autoCorrect: Boolean, emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit) {
+        if (node.elementType == KtNodeTypes.FUN) {
+            node.findChildByType(KtNodeTypes.BLOCK)?.let {
+                it as CompositeElement
+                if (it.getBody().size == 1) {
+                    emit(node.firstChildNode.startOffset, "Single expression methods should use expression body", false)
```
shashachu: nit: `Single expression methods should use *an* expression body"

shashachu:
This approach actually looks fine to me, but I'd prefer holding off landing it until auto-correction works. I think it should be relatively straightforward - you would need to delete the return type, insert an `=`, and delete the brackets and newlines. The one annoying thing would be that it's possible you would then cause a max-line error if the single expression were very long, but that feels like a rare enough occurrence that I'm okay with it.

Also, in case you're not aware, `ktlint` runs itself on its own code, which is why your CI build is failing.

Finally, please land this into the `experimental` ruleset for now.

bbaldino:
Ok, added auto correction and moved things to experimental.  Let me know if this looks alright.

ktlint-ruleset-standard/src/test/resources/spec/method-expression-body/format.kt.spec
```diff
@@ -0,0 +1,15 @@
+class C0 {
+    fun good(): Int = funcReturningInt()
+    fun bad(): Int {
+        return funcReturningInt()
```
arturbosch: Thank you for building auto correcting of this rule!
Please take a look at some additional test case we have starting https://github.com/arturbosch/detekt/blob/04a9ef332de2ee5bd45c4be9bd480bf28ecabd01/detekt-rules/src/test/kotlin/io/gitlab/arturbosch/detekt/rules/style/ExpressionBodySyntaxSpec.kt#L58-L66.
Be aware that some of our users (and maybe also KtLint users) prefer not to use the expression syntax for functions if the return statement reaches multiple lines:
```
fun stuff(arg: Int): Int {
    return when(arg) {
        0 -> 0
		else -> 1
    }
}
```

bbaldino: Applying the rule in that case seems like what you'd want if this rule was turned on, no?
```
fun stuff(arg: Int): Int = when(arg) {
        0 -> 0
	else -> 1
    }
}
```

bbaldino: Is there a mechanism in ktlint to apply a sort of 'sub option' to a rule, or if we wanted to allow the user to apply this but not for multi-line (like above), would it have to be a separate rule?

bbaldino:
added some more test cases.  thanks @arturbosch for the pointer to detekt's cases.

ktlint-ruleset-standard/src/test/resources/spec/method-expression-body/format.kt.spec
shashachu: @bbaldino there is not. The Android guide says:
```
Otherwise, if an expression function grows to require wrapping, use a normal function body, a return declaration, and normal expression wrapping rules instead.
```
So perhaps you can make this even stricter that it will only fail the rule if the expression is also a single *line*.

bbaldino: ok, i'll take a look at that.

bbaldino:
Ok, made some changes to ignore multi-line, single-expression methods.  Also realized I didn't have things in the right place, so moved them to the proper 'experimental' project.

bbaldino:
The build should be fixed after the newline changes, not sure if Travis is picking them up to trigger a re-run.

bbaldino:
@shashachu any other thoughts?

shashachu:
Hey @bbaldino, thanks for the work. I know I'm the one who suggested you open the PR, but re-reading the styleguides, I'm a bit on the fence. The JetBrains styleguide says 'prefer', and the Android styleguide says 'can'. I think since we're going to be releasing some sort of way to globally disable rules it's probably okay to drop this into the experimental rules for now.

bbaldino:
@shashachu sure, sounds good if you think it's a fit.  personally i do think it'd be nice to be able to enforce this, but i do understand it's not a hard-and-fast rule.  if that fits well with 'experimental' then that'd be great.

shashachu:
@bbaldino sorry for letting this sit so long; after some thought I think it's better to leave this as a custom rule. Thanks for your work on it.

PR #446, opened 2019-06-02
Enable final new line by default

### LINKED ISSUE #312: Turn on `insert_final_newline` by default
Unfortunately #103 has been closed, without actually doing the change...
So I would like to propose the change again.

The Readme says: 

> Newline at the end of each file (_not enabled by default, **but recommended**_)
> (set insert_final_newline=true in .editorconfig to enable)

But Ktlint tries to be an "anti-bikeshedding Kotlin linter" with 
> "No configuration. Which means no decisions to make"

So recommending something, but disabling it by default makes not much sense to me. Thats why I think that should be changed.

jaredsburrows:
@shashachu 
Closes #312


JLLeitschuh:
Yes please!

Issue #445, opened 2019-06-02
Remove Maven, Use Gradle


jaredsburrows:
There are Maven and Gradle files. Can we go through and make this a Gradle based project?

shashachu:
Yes, I'd love for it to be gradle-based. Just haven't had the time to do it.

jaredsburrows:
Do you have a list of items that remain? project seems to build just file. What about testing and publishing?

shashachu:
We haven't tried testing and deploying with gradle yet so I'm unsure if they work or not.

bethcutler:
On a related note, it would be nice to remove the [MavenDependencyResolver](https://github.com/pinterest/ktlint/blob/master/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/MavenDependencyResolver.kt).  The dependency on Maven source is the #1 blocker to being able to compile ktlint within the Google codebase.  I will likely keep looking for a solution for it, but would appreciate others chiming in if they have any information about how this is used or what it could be replaced with.

shashachu:
@bethcutler doing some GitHub archaeology, it looks like it was added to support loading 3rd party reporters as JARs from the command line: https://github.com/pinterest/ktlint/issues/176

shashachu:
I would personally be fine slightly breaking that functionality and requiring command-line users to download reporter jars locally instead of resolving them automatically, but I don't have a good sense of how important this functionality is. @jeremymailen @JLLeitschuh is this used in either of your plugins?

jeremymailen:
Nope, kotlinter-gradle doesn't need the MavenDependencyResolver.

Reporters aren't extensible yet and if they were, would use Gradle's ability to resolve dependencies similar to adding rulesets https://github.com/jeremymailen/kotlinter-gradle#custom-rules

JLLeitschuh:
@shashachu No, we use Gradle's jar resolver.

The only security vulnerability ever reported against Ktlint was due to the internal KtLint repository resolver.

https://nvd.nist.gov/vuln/detail/CVE-2019-1010260
https://github.com/shyiko/ktlint/pull/332

Thankfully, that's been fixed.

I don't know how the Ktlint maven plugin works though.

https://github.com/gantsign/ktlint-maven-plugin

CC: @Tapchicoma

Tapchicoma:
Originally ticket was about using Gradle build system to build ktlint itself.

[...]

PR #454, opened 2019-06-05
Process of migrating to gradle - update gradle to 5.4.1

jaredsburrows:
#445

@shashachu

Update gradle to 5.4.1


JLLeitschuh:
Moving forward, please ensure the validity of the Gradle wrapper jar before merging in the future.

I know @jaredsburrows is trusted, but in general, a PR with a giant blob of binary should always be checked for validity.

https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapper_checksum_verification

jaredsburrows:
You would think that this would be build into the wrapper.

shashachu:
@JLLeitschuh good call-out. Thanks

JLLeitschuh:
> You would think that this would be build into the wrapper.

You're referring to the `gradlew` and `gradlew.bat` files? It could be done that way, but then every Gradle launch would have the added overhead of an SHA checksum check and network access.

Feel free to open an issue with gradle/gradle if you want to discuss this idea and I can discuss it with the team internally. (I start with Gradle officially on June 17th).

PR #459, opened 2019-06-06
Implement .ktlintignore for globally disabling rules

### LINKED ISSUE #208: Support for globally disabling a rule
I've seen that this issue of support for globally disabling a rule comes up [multiple times](https://github.com/shyiko/ktlint/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+disable+rule). I'd like to propose simple support (not even for a given scope, just across the board) for disabling a comma separated list of rule names in the .editorconfig or some ktlint config file or via command line. This would be an opt-in advanced feature not put any burden on someone who does not want to make any config decisions.

shashachu:
Fixes #208

* Added support for .ktlintignore file - takes a list of rule Ids to disable (ignores comments)
* Currently only supports globally disabling rules, but added some plumbing so that it's easier to support per-file/per-path disabling, or multiple .ktlintignore files
* Re-enabled NoWildcardImports, PackageNameRule
* Un-commented AnnotationRule, MultiLineIfElseRule, and NoItParamInMultilineLambdaRule, and moved disabling into .ktlintignore

### DIFF
--- a/.ktlintignore
+++ b/.ktlintignore
@@ -1,0 +1,7 @@
+// disabled ("./mvnw clean verify" fails with "Internal Error")
+annotation
+// disabled until auto-correct is working properly
+// (e.g. try formatting "if (true)\n    return { _ ->\n        _\n}")
+multiline-if-else
+// disabled until it's clear what to do in case of `import _.it`
+no-it-in-multiline-lambda
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
@@ -37,6 +37,7 @@ object KtLint {
     val EDITOR_CONFIG_USER_DATA_KEY = Key<EditorConfig>("EDITOR_CONFIG")
     val ANDROID_USER_DATA_KEY = Key<Boolean>("ANDROID")
     val FILE_PATH_USER_DATA_KEY = Key<String>("FILE_PATH")
+    val DISABLED_RULES = Key<Set<String>>("DISABLED_RULES")
 
     private val psiFileFactory: PsiFileFactory
     private val nullSuppression = { _: Int, _: String, _: Boolean -> false }
@@ -186,15 +187,17 @@ object KtLint {
                 userData
             }
         node.putUserData(FILE_PATH_USER_DATA_KEY, userData["file_path"])
-        node.putUserData(EDITOR_CONFIG_USER_DATA_KEY, EditorConfig.fromMap(editorConfigMap - "android" - "file_path"))
+        node.putUserData(EDITOR_CONFIG_USER_DATA_KEY, EditorConfig.fromMap(editorConfigMap - "android" - "file_path" - "disabled_rules"))
         node.putUserData(ANDROID_USER_DATA_KEY, android)
+        node.putUserData(DISABLED_RULES, userData["disabled_rules"]?.split(",")?.toSet() ?: setOf())
     }
 
     private fun visitor(
         rootNode: ASTNode,
         ruleSets: Iterable<RuleSet>,
         concurrent: Boolean = true,
-        filter: (fqRuleId: String) -> Boolean = { true }
+        filter: (rootNode: ASTNode, fqRuleId: String) -> Boolean = this::filterDisabledRules
+
     ): ((node: ASTNode, rule: Rule, fqRuleId: String) -> Unit) -> Unit {
         val fqrsRestrictedToRoot = mutableListOf<Pair<String, Rule>>()
         val fqrs = mutableListOf<Pair<String, Rule>>()
@@ -204,7 +207,7 @@ object KtLint {
             val prefix = if (ruleSet.id === "standard") "" else "${ruleSet.id}:"
             for (rule in ruleSet) {
                 val fqRuleId = "$prefix${rule.id}"
-                if (!filter(fqRuleId)) {
+                if (!filter(rootNode, fqRuleId)) {
                     continue
                 }
                 val fqr = fqRuleId to rule
@@ -252,6 +255,10 @@ object KtLint {
                 }
             }
         }
+    }
+
+    private fun filterDisabledRules(rootNode: ASTNode, fqRuleId: String): Boolean {
+        return rootNode.getUserData(DISABLED_RULES)?.contains(fqRuleId) == false
     }
 
     private fun calculateLineColByOffset(text: String): (offset: Int) -> Pair<Int, Int> {
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
@@ -20,6 +20,7 @@ import java.io.IOException
 import java.io.PrintStream
 import java.math.BigInteger
 import java.net.URLDecoder
+import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
 import java.security.MessageDigest
@@ -234,6 +235,12 @@ object Main {
     )
     private var experimental: Boolean = false
 
+    @Option(names = arrayOf("--ktlintignore"), description = arrayOf("Path to .ktlintignore (defaults to working dir)"))
+    private var ktlintIgnorePath: String? = null
+
+    // Rules from .ktlintignore that will be globally disabled
+    private val disabledRules = mutableSetOf<String>()
+
     @Parameters(hidden = true)
     private var patterns = ArrayList<String>()
 
@@ -300,6 +307,10 @@ object Main {
             exitProcess(0)
         }
         val start = System.currentTimeMillis()
+
+        // Parse the .ktlintignore file
+        parseKtlintIgnore(ktlintIgnorePath ?: ".")
+
         // load 3rd party ruleset(s) (if any)
         val dependencyResolver = lazy(LazyThreadSafetyMode.NONE) { buildDependencyResolver() }
         if (!rulesets.isEmpty()) {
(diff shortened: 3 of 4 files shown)


.ktlintignore
```diff
@@ -0,0 +1,7 @@
+// disabled ("./mvnw clean verify" fails with "Internal Error")
```
shashachu: This file is kind of goofy, but I liked the idea of allowing comments so that we can explain why a rule is disabled

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
```diff
@@ -425,7 +436,11 @@ object Main {
                         }
                     }
                     ?: emptyMap<String, String>()
-                ) + cliUserData + ("file_path" to fileName)
+                ) +
+                cliUserData +
+                // For now, use a global list
+                ("disabled_rules" to disabledRules.joinToString(",")) +
```
shashachu: Here is where we'd handle logic around a .ktlintignore file that's formatted more like an .editorconfig file.

shashachu:
Open for suggestions on good ways to test this.

shashachu:
@JLLeitschuh @jaredsburrows @jeremymailen thoughts?

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
```diff
@@ -300,6 +307,10 @@ object Main {
             exitProcess(0)
         }
         val start = System.currentTimeMillis()
+
+        // Parse the .ktlintignore file
+        parseKtlintIgnore(ktlintIgnorePath ?: ".")
```
jeremymailen: It would be useful to put support classes in the core module to locate any .ktlintignore files and parse them to disabled rules userdata. This would allow tools not invoking the CLI portion of ktlint to have an easy way to support .ktlintignore exactly the same way.

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
```diff
@@ -234,6 +235,12 @@ object Main {
     )
     private var experimental: Boolean = false
 
+    @Option(names = arrayOf("--ktlintignore"), description = arrayOf("Path to .ktlintignore (defaults to working dir)"))
```
jeremymailen: Would they pass in a path or would it happen by convention such as working directory plus home directory similar to .gitignore? I guess global prefs make less sense here -- possibly at each package level? Although I'm not sure if it's easy to support or even worth it.

.ktlintignore
jeremymailen: nit: .xyzignore files are conventionally lists of file expressions to ignore. What if it was `.ktlintconfig` and perhaps in something like toml?

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
jeremymailen: For `kotlinter-gradle` I suppose I'd start by saying the file had to be in the gradle project root. Of course ktlint CLI is agnostic to the idea of a project.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -254,6 +257,10 @@ object KtLint {
         }
     }
 
+    private fun filterDisabledRules(rootNode: ASTNode, fqRuleId: String): Boolean {
+        return rootNode.getUserData(DISABLED_RULES)?.contains(fqRuleId) == false
```
rom4ek: Note: this will currently work only with the standard ruleset. If you have an experimental one (or any other custom ruleset), you have to specify the rule with the prefix inside `.ktlintignore` (as you can see it's being added to the rule id at the line 207). So something like `experimental:indent` - I guess this should be mentioned in the docs or something :) Or we could change this to `any { fqRuleId.contains(it) }`, if we want to be independent of the ruleset (as I assume rule ids most of the time are unique)

shashachu: That's correct. I'd rather require fully namespaced rules than unexpectedly disabling a rule with a name collision. We would definitely update the docs if we move forward with this.

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
shashachu: If you look a bit further down, it does default to ".".

And yes, I was unsure whether we want/need to support multiple ignore/config files. I can certainly imagine a scenario (eg codegen files) but I'm unsure how common it would be. It seems like the biggest ask is for global disabling.

I originally considered a `.ktlintconfig` with the same syntax as `.editorconfig` so you could easily disable rules on a per-path basis but thought maybe the simpler approach was fine for now.

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
shashachu: sounds good. You actually know much more about the internals of `ktlint` than I do... 🤦‍♀

JLLeitschuh:
CC: @Tapchicoma

.ktlintignore
shashachu: Yes - it's goofy. Probably a better format is something like `.editorconfig`:

```
[*]
disable_rules=no-wildcard-imports

[**/library/*]
disable_rules=no-wildcard-imports,package-name
```

jaredsburrows:
This project already uses `.editorconfig`, why not add this functionality to `.editorconfig`? This way we do not have to have another file to configure ktlint.

shashachu:
> This project already uses `.editorconfig`, why not add this functionality to `.editorconfig`? This way we do not have to have another file to configure ktlint.

It felt like an abuse of `.editorconfig`. I suppose disabling lint rules falls loosely under 'coding style', but it doesn't really seem like it fits the spirit of what an `.editorconfig` is supposed to do. I'm more on the side of it being a separate file, but it's not a dealbreaker for me.

bethcutler:
Agreed that that would be an abuse of .editorconfig, which is a format that extends beyond just ktlint.

On the other hand, each new configuration file required for ktlint makes integration with other systems a bit more difficult.  Would it make sense to add a new custom configuration file that could handle both the ignore rules and the editorconfig rules?  Ideally, editorconfig would still be supported, but have fewer features.

shashachu:
> Agreed that that would be an abuse of .editorconfig, which is a format that extends beyond just ktlint.
> 
> On the other hand, each new configuration file required for ktlint makes integration with other systems a bit more difficult. Would it make sense to add a new custom configuration file that could handle both the ignore rules and the editorconfig rules? Ideally, editorconfig would still be supported, but have fewer features.

That would be fine as well. I do think that the ignore rules are currently the only ktlint-specific config, though.

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
Tapchicoma: I would agree with @shashachu, that it should be possible to add additionally rules to exclude in subprojects and `.ktlintconfig` should behave like `.editorconfig`.

.ktlintignore
Tapchicoma: > Yes - it's goofy. Probably a better format is something like .editorconfig:

yes, this format of the file looks better and will allow in the future add more `ktlint` specific configuration here.

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
```diff
@@ -784,6 +799,29 @@ object Main {
                 map
             }
 
+    /**
+     * Parses the .ktlintignore file, a list of ids of rules to globally disable. One rule per line.
+     */
+    private fun parseKtlintIgnore(path: String) {
```
Tapchicoma: I think, this function should repeat the logic of `.editorconfig` lookup - https://editorconfig.org/#file-location

Implementation: https://github.com/pinterest/ktlint/blob/6f8bb56709e5394497a372402c7570326ffc32dd/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/EditorConfig.kt#L20

This would allow to define global ktlint rules in project root and have sub-project specific one, that will enhance/override global rules.

Tapchicoma:
> This project already uses .editorconfig, why not add this functionality to .editorconfig? This way we do not have to have another file to configure ktlint.

I agree with @jaredsburrows, from what I see - adding just custom property `ktlint-disabled-rules` to `.editorconfig` that falls under [domain-specific-properties](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#ideas-for-domain-specific-properties) should be enough.

Though required change would be move detecting and parsing `.editorconfig` files to `ktlint-core`, so @jeremymailen plugin could also support this.

shashachu:
Thanks for the feedback, everyone. I'll close this PR and open another one using either `.editorconfig` or a `.ktlintconfig` file.

jaredsburrows:
Sounds good, thanks!

JorgeCastilloPrz:
How's this going @shashachu ? Are there plans on opening that alternative PR? Let me know if you need any help, this is a feature I'm interested in.

shashachu:
@JorgeCastilloPrz yes, there are still plans to do the PR; things have just been extremely busy. I have a PR in-progress locally, so I will aim to put it up by the end of the week.

JorgeCastilloPrz:
Is there a place we can reach you for direct chat @shashachu ? Maybe Kotlinlang Slack? We can try to help on coding it if you're busy, we'd just need some indications since we're not familiar with the codebase.

shashachu:
All, thanks for your patience. Pushed the updated editorconfig-based-PR here: https://github.com/pinterest/ktlint/pull/503

Please feel free to patch in and test locally.

@JorgeCastilloPrz I am working on getting a slack channel setup in the Pinterest slack team so that I can invite external contributors.

JorgeCastilloPrz:
Thanks for your work on it! We'll keep a close eye to that PR :)

PR #460, opened 2019-06-08
Process of migrating to gradle - organize gradle dependencies

jaredsburrows:
#445

@shashachu

Clean up and organize Gradle dependencies across the project. This is similar to other open source Gradle projects such as https://github.com/JakeWharton/SdkSearch/blob/master/build.gradle#L18.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,27 +1,39 @@
 plugins {
-  id "org.jetbrains.kotlin.jvm" version "1.3.31" apply false
+  id 'org.jetbrains.kotlin.jvm' version '1.3.31' apply false
 }
 
-ext.libraries = [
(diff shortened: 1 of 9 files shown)


shashachu:
Interesting! I've never seen this particular format. Very nice.

shashachu:
BTW, probably outside the scope of this migration, but I'd love to eventually port the build scripts to kotlin as well.

jaredsburrows:
@shashachu I can make a PR for that as well.

rom4ek:
@jaredsburrows Btw, it's already too late to ask, but is there a specific reason of not using the `implementation`/`api` instead of `compile`? Just noticed that.

jaredsburrows:
@rom4ek This CL was to organize the dependencies.

PR #464, opened 2019-06-08
Replace usage of arrayOf with array literal.

Tapchicoma:
And made command line options style consistent.

### DIFF
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
@@ -82,7 +82,7 @@ Examples:
 
 Flags:""",
     synopsisHeading = "",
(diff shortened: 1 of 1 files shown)


shashachu:
Is there a reason why array literals are preferred?

jaredsburrows:
@shashachu The Kotlin plugin in the IDE shows this suggestion.

Tapchicoma:
> @shashachu The Kotlin plugin in the IDE shows this suggestion.

Yes, that was main trigger for me.

Imho, visually, it looks better in annotations parameters, then `arrayOf(..)`.

shashachu:
Ah ok I just did some research and array literal syntax is only supported for annotations, and only since Kotlin 1.2. I guess people just thought `arrayOf` in annotations looked ugly. :)

shashachu:
lgtm, thanks!

PR #465, opened 2019-06-08
Process of migrating to gradle - Fix main class name for this project

jaredsburrows:
#445

@shashachu 

From `com.github.shyiko.ktlint.Main` to `com.pinterest.ktlint.Main`


shashachu:
Thanks! I missed this one during the changeover.

PR #466, opened 2019-06-08
Build with JDK8 and JDK11

jaredsburrows:
@shashachu

Build with JDK8 and build with JDK11.


jaredsburrows:
@shashachu Can we merge this one?

PR #469, opened 2019-06-09
Process of migrating to gradle - Run ktlint on itself

### LINKED ISSUE #463: Run ktlint over itself
Would be nice to add a step on CI that runs `ktlint` over own sources, so code style in this project would be consistent.

jaredsburrows:
Closes #463

@shashachu 

This runs ktlint on the entire project similar to how Maven is running ktlint on the project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,15 @@ install: true
 script:
   - |
     if [[ $TRAVIS_REPO_SLUG == "pinterest/ktlint" && $TRAVIS_BRANCH == "master" && $TRAVIS_PULL_REQUEST == "false" ]]; then
-      ./gradlew build && ./mvnw -Ddeploy=maven-central -DskipStaging=true -Dgpg.skip=true && ./mvnw -Ddeploy=github -Dgpg.skip=true -Dgithub.draft=true -Dgithub.description="https://github.com/pinterest/ktlint#access-to-the-latest-master-snapshot"
+      # Gradle
+      ./gradlew build
+      # Maven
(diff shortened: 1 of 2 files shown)


jaredsburrows:
@shashachu Once this is merged, I will make PRs to fix all the ktlint issues.

.travis.yml
```diff
@@ -7,9 +7,16 @@ install: true
...
-      ./gradlew build && ./mvnw -Ddeploy=maven-central -DskipStaging=true -Dgpg.skip=true && ./mvnw -Ddeploy=github -Dgpg.skip=true -Dgithub.draft=true -Dgithub.description="https://github.com/pinterest/ktlint#access-to-the-latest-master-snapshot"
+      # Gradle
+      ./gradlew build
+      # Maven
+      ./mvnw -Ddeploy=maven-central -DskipStaging=true -Dgpg.skip=true && ./mvnw -Ddeploy=github -Dgpg.skip=true -Dgithub.draft=true -Dgithub.description="https://github.com/pinterest/ktlint#access-to-the-latest-master-snapshot"
     else
-      ./mvnw clean verify
+      # Gradle
+      ./gradlew build # Make sure it builds on each PR
+      ./gradlew ktlint || true # Return true for now until we have fixed all of the ktlint errors
```
Tapchicoma: Actually maven already runs ktlint on this project (sorry, missed that and opened #463), so `|| true # Return true for now until we have fixed all of the ktlint errors` should be not needed here.

Tapchicoma: Check this line: https://github.com/pinterest/ktlint/blob/b09945530513a6bf4edf7c5f45524609c74b26ce/ktlint/pom.xml#L199

jaredsburrows: If it does run, why are there no errors? Look at the TravisCI output of this build, there are many formatting errors.

Tapchicoma: It doesn't use `--android` flag.

jaredsburrows: Removed.

jaredsburrows:
@shashachu Can we merge this one?

PR #461, opened 2019-06-08
Adding experimental rule for detecting first line blank in method block

sowmyav24:
This experimental rule addresses issue [Issue #440](https://github.com/pinterest/ktlint/issues/440).

In addition to checking for first line blank in a method block, it will check for any first blank line in code blocks of the method.Please let me know if the subsequent code blocks in the method need to be ignored from this check.

Due to this new rule, there were violations reported in the code. Have fixed them also.

Issue #440, opened 2019-05-30
Add "First method line is empty" check


igorwojda:
OK
```
fun bar() {
   val a = 2
}
```

Fail
```
fun bar() {

   val a = 2
}
```

igorwojda:
Probably we should also add an equiwalent check for last empty method line

OK
```
fun bar() {
   val a = 2
}
```

Fail
```
fun bar() {
   val a = 2

}
```

sowmyav24:
The `last empty method line` check should be covered by the rule `NoBlankLineBeforeRbraceRule`. Probably a `NoBlankLineAfterLbraceRule` should cover the first scenario.

sowmyav24:
Created a PR for `first method line empty` rule. In addition to the scenarios present above, few others that would be covered are
```
fun bar() {
   if(someCondition) {
       
    val a = 2
   }
}
```

shashachu:
Fixed in https://github.com/pinterest/ktlint/pull/474

Zayankovsky:
This method triggers `no-empty-first-line-in-method-block` error:
```
fun whatever() {
    window.addMouseListener(object : MouseAdapter() {

        override fun mouseClicked(e: MouseEvent) { /*...*/ }

        override fun mouseEntered(e: MouseEvent) { /*...*/ }
    })
}
```
Should it be this way? I prefer to keep first line in complex object expressions empty, just like in classes.

sowmyav24:
https://github.com/pinterest/ktlint/pull/461#issue-286397143

> In addition to checking for first line blank in a method block, it will check for any first blank line in code blocks of the method.Please let me know if the subsequent code blocks in the method need to be ignored from this check.

Since, the object expression is inside the method, it triggers a violation for the same.

Zayankovsky:
Both [Kotlin Coding Conventions](https://kotlinlang.org/docs/reference/coding-conventions.html) and [Android Kotlin style guide](https://developer.android.com/kotlin/style-guide) don't mention anything about such a requirement. Why is it forced then? I believe object expressions should be excluded from this check.

JakeWharton:
Actually the Android Kotlin style guide [explicitly mentions](https://developer.android.com/kotlin/style-guide#vertical) this whitespace:

> _Optionally_ before the first statement in a function, before the first member of a class, or after the last member of a class (neither encouraged nor discouraged).

Zayankovsky:
[...]

PR #474, opened 2019-06-10
Adding experimental rule for detecting first line blank in method block

sowmyav24:
This experimental rule addresses issue Issue [#440](https://github.com/pinterest/ktlint/issues/440)
In addition to checking for first line blank in a method block, it will check for any first blank line in code blocks of the method.Please let me know if the subsequent code blocks in the method need to be ignored from this check.

Due to this new rule, there were violations reported in the code. Have fixed them also.

This is same as PR [#461](https://github.com/pinterest/ktlint/pull/461).


shashachu:
> In addition to checking for first line blank in a method block, it will check for any first blank line in code blocks of the method

Does this mean that
```
fun bar() {
    if (something) {

        foo()
    }
}
```
will not change during formatting, but
```
fun bar() {

    if (something) {

        foo()
    }
}
```
will format to
```
fun bar() {
    if (something) {
        foo()
    }
}
```
?

If so, that inconsistency seems unexpected.

sowmyav24:
```
fun bar() {
    if (something) {

        foo()
    }
}
```
would also format to 
```
fun bar() {
    if (something) {
        foo()
    }
}
```

Added a test for this scenario also

shashachu:
looks good to me- thanks!

PR #476, opened 2019-06-10
Add missing SpacingAroundUnaryOperatorsRule to standard rule set

rom4ek:
I guess you'd also like to update the release changelog, and maybe release a new patch version where the rule is included :) 

As this happens not for the first time, that newly implemented rules are not being added to the ruleset, I came up with a test which checks whether a provider returns all the rules from the `src` folder or not. I will follow on this up with a PR.


shashachu:
🤦‍♀

shashachu:
But actually that test won't work until we have some sort of `.ktlintignore` or `.ktlintconfig` working, since currently there are a few rules commented out

shashachu:
https://twitter.com/thepracticaldev/status/687672086152753152

rom4ek:
@shashachu Exactly, so as I mentioned in the other PR, either we wait until global ignores land on master, or I could also implement ignores in that test :)

shashachu:
Hmm looks like this rule is buggy. It's failing on `val el = node.psi.findElementAt(offset + line.length - 1)!!.node`

cc @z3d1k 

I'm actually going to revert the rule for now, and when we bring it back, let's put it under experimental.

PR #477, opened 2019-06-10
Add tests for missing rules in a ruleset

rom4ek:
This is a follow up for #476. I would suggest merging this after the `.ktlintignore` landed on `master`. Or I could also implement ignores list for this kind of tests for now, but I would rather wait for global ignores. It just grabs the source files from the specified location and then checks that all of them are being returned from the specified ruleset.

### DIFF
--- a/ktlint-test/src/main/kotlin/com/pinterest/ktlint/test/RuleSetProviderTest.kt
+++ b/ktlint-test/src/main/kotlin/com/pinterest/ktlint/test/RuleSetProviderTest.kt
@@ -1,0 +1,29 @@
+package com.pinterest.ktlint.test
+
+import com.pinterest.ktlint.core.RuleSetProvider
+import java.io.File
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+open class RuleSetProviderTest(
+    private val rulesetClass: Class<out RuleSetProvider>,
+    private val packageName: String
+) {
+
+    @Test
+    fun checkAllRulesProvided() {
+        val srcLocation = rulesetClass.protectionDomain.codeSource.location.path
+        val rulesDir = File(srcLocation + packageName.replace(".", "/"))
+        val packageRules = rulesDir.listFiles()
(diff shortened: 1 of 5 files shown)


ktlint-test/src/main/kotlin/com/pinterest/ktlint/test/RuleSetProviderTest.kt
```diff
@@ -0,0 +1,28 @@
+package com.pinterest.ktlint.test
+
+import com.pinterest.ktlint.core.RuleSetProvider
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import java.io.File
+
+open class RuleSetProviderTest(
+    private val rulesetClass: Class<out RuleSetProvider>,
+    private val packageName: String
```
rom4ek: nit: We could, in theory, get the package name from `rulesetClass` and this would work with current ktlint structure, but in case of custom rulesets, the rules could be located in a different folder, so I would rather leave it like it is now

ktlint-test/src/main/kotlin/com/pinterest/ktlint/test/RuleSetProviderTest.kt
```diff
@@ -0,0 +1,28 @@
+package com.pinterest.ktlint.test
+
+import com.pinterest.ktlint.core.RuleSetProvider
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import java.io.File
+
+open class RuleSetProviderTest(
```
rom4ek: I'm not sure that this is the best way for the centralized test code - this could be just a simple top-level function, which may be called from within a test method (e.g. in case `StandardRuleSetProviderTest` contains other tests), so I'm open for changing this, but currently it works.

shashachu:
@rom4ek I just landed the rule disabling PR (https://github.com/pinterest/ktlint/pull/503) so this should be able to pass now.

rom4ek:
@shashachu updated the PR accordingly. Thanks!

PR #478, opened 2019-06-10
Use picoli help mixin.

Tapchicoma:
This removes explicit version and help fields.
Version detection was moved to separate class.

More info about mixin: https://picocli.info/#_mixin_standard_help_options


shashachu:
Nice

PR #487, opened 2019-06-11
Add "installGitPreCommitHook" sub command.

Tapchicoma:
Move old one `--install-git-pre-commit-hook` option to be a ktlint
subcommand - new syntax: `ktlint installGitPreCommitHook`. Old option
`--install-git-pre-commit-hook` is still working, but deprecated.

This sub command acts as a separate action, that is not running checks, and should not confuse users oppose to option.


shashachu:
👍

PR #471, opened 2019-06-09
Process of migrating to gradle - Add maven publish

jaredsburrows:
@shashachu 

Add support for publish Ktlint via Gradle
I used @vanniktech's plugin and here is an example of integrating it: https://github.com/square/okhttp/commit/c53f8db73d53eb8fc1c5d426568d35e9b99c2fa2.

PR #489, opened 2019-06-14
Migrating to gradle: Add gradle shadow plugin

shashachu:
* Integrated https://github.com/johnrengelman/shadow
* `./gradlew clean build shadowJar` outputs `ktlint/build/ktlint-all.jar`
* `java -jar ktlint/build/ktlint-all.jar` works

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,7 +46,11 @@ configurations {
 
 dependencies {
   ktlint project(':ktlint')
-  compile subprojects
+  subprojects.forEach { subproject ->
+    if (subproject.name != 'ktlint-ruleset-template') {
+      compile subproject
+    }
+  }
 }
 
 task ktlint(type: JavaExec, group: LifecycleBasePlugin.VERIFICATION_GROUP) {
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,7 @@
 plugins {
   id 'org.jetbrains.kotlin.jvm' version '1.3.31' apply false
+  id 'com.github.johnrengelman.shadow' version '5.0.0'
+  id 'java'
 }
 
 ext.versions = [
@@ -44,6 +46,7 @@ configurations {
 
 dependencies {
   ktlint project(':ktlint')
+  compile subprojects
 }
 
 task ktlint(type: JavaExec, group: LifecycleBasePlugin.VERIFICATION_GROUP) {
@@ -57,4 +60,14 @@ allprojects {
   repositories {
     gradlePluginPortal()
   }
+}
+
+jar {
+  manifest {
+    attributes 'Main-Class': 'com.pinterest.ktlint.Main'
+  }
+}
+
+shadowJar {
+  mergeServiceFiles()
 }


shashachu:
actually doesn't seem to be working properly

build.gradle
```diff
@@ -44,6 +46,11 @@ configurations {
 
 dependencies {
   ktlint project(':ktlint')
+  subprojects.forEach { subproject ->
+    if (subproject.name != 'ktlint-ruleset-template') {
```
shashachu: This is pretty dopey, but excluding the project in the `shadowJar` config as described in this issue https://github.com/johnrengelman/shadow/issues/39 doesn't seem to work anymore.

shashachu:
@jaredsburrows thoughts? This is cobbled together from some shadow plugin examples.

build.gradle
rom4ek: But what's the reasoning behind including the template module into the project? I mean `settings.gradle`

shashachu: Presumably to ensure it keeps building, but maybe it just causes more issues than it solves.

rom4ek: Could be a separate pipeline maybe

jaredsburrows:
@shashachu I actually removed that plugin here: https://github.com/pinterest/ktlint/pull/465/files#diff-6bb5647686cc692dfa23d5d48d46e32aL1.

If you run `gradlew build`, then look in `./ktlint/ktlint/build/distributions`, you will see `ktlint-0.3.4-SNAPSHOT.zip`. If you unzip it, `unzip ktlint-0.3.4-SNAPSHOT.zip`, you will see the `bin` and `lib` folder. The `bin` folder has the `ktlint` bash script. Then you run `./ktlint --help`

jaredsburrows:
@shashachu If you want this functionality back, I think you should follow the setup here: https://github.com/pinterest/ktlint/pull/465/files#diff-6bb5647686cc692dfa23d5d48d46e32aL1.

shashachu:
We don't want to use the bash distribution, as it requires the `lib` folder and all the separate jar dependencies. I think the executable jar is the right approach. Sorry, I didn't notice the plugin had been removed in that diff.

shashachu:
> @shashachu If you want this functionality back, I think you should follow the setup here: https://github.com/pinterest/ktlint/pull/465/files#diff-6bb5647686cc692dfa23d5d48d46e32aL1.

I had originally tried the same setup and it didn't actually include any of the jars built from the sub-modules. I can give it another stab this weekend.

jaredsburrows:
I have added the plugin to my PR here: https://github.com/pinterest/ktlint/pull/471.

Tapchicoma:
My 5 cents here: @shashachu if you will use Gradle for next release publishing, could you ensure that `ktlint` executable behaves the same as if it was built with maven?

My point that it should work without any additional jars in the class path and run by simply calling `./ktlint`

shashachu:
> My 5 cents here: @shashachu if you will use Gradle for next release publishing, could you ensure that `ktlint` executable behaves the same as if it was built with maven?
> 
> My point that it should work without any additional jars in the class path and run by simply calling `./ktlint`

Totally agree. And there's no reason why the Gradle build needs to be tied to any particular release since it's used internally. Whenever we do make the switch, we will make sure the behavior is identical to the existing maven build.

build.gradle
```diff
@@ -1,5 +1,7 @@
 plugins {
   id 'org.jetbrains.kotlin.jvm' version '1.3.31' apply false
+  id 'com.github.johnrengelman.shadow' version '5.0.0'
+  id 'java'
```
JLLeitschuh: Consider using the `java-library` plugin instead. The `java` plugin is legacy.

PR #492, opened 2019-06-16
Fix ktlint build and configure language compatibility from root level

jaredsburrows:
Fix ktlint build and configure language compatibility from root level

@shashachu


jaredsburrows:
@shashachu Once we merge this, we can rebase the other pull requests.

Issue #490, opened 2019-06-14
Annotation rule throws IllegalStateException


zsavely:
Two tests added to the [AnnotationRuleTest.kt](https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRuleTest.kt). The first one failed, the second one succeeds.

> com.pinterest.ktlint.core.RuleExecutionException: java.lang.IllegalStateException: Annotations list should not be empty

```kotlin
@Test
fun `no annotation present fails`() {
    val code =
        """
        package com.example.application.a.b

        data class FileModel(val uri: String, val name: String)
        """.trimIndent()
    assertThat(AnnotationRule().format(code)).isEqualTo(
        """
        package com.example.application.a.b

        data class FileModel(val uri: String, val name: String)
        """.trimIndent()
    )
}

@Test
fun `no annotation present succeeds`() {
    val code =
        """
        package com.example.application.a

        import android.os.Environment

        class PathProvider {
            fun gallery(): String = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES).path
        }
        """.trimIndent()
    assertThat(AnnotationRule().format(code)).isEqualTo(
        """
        package com.example.application.a

        import android.os.Environment

        class PathProvider {
            fun gallery(): String = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES).path
        }
        """.trimIndent()
    )
}
```

sowmyav24:
Included the above two examples in the `AnnotationRuleTest` in the PR

sowmyav24:
The [PR](https://github.com/pinterest/ktlint/pull/497) for this issue has been merged. But just now noticed that in the `StandardRuleSetProvider`, the rule is disabled due to this issue. Can the rule be re-enabled and this issue closed ?

shashachu:
Disabling that rule predates us taking over this project, but I can verify that all checks pass once it's re-enabled.

sowmyav24:
Sure ! Noticed all the rules disabled earlier have been re-enabled in this open [PR](https://github.com/pinterest/ktlint/pull/503/files).

shashachu:
It's a bit confusing - some of the disabled ones are still disabled, but the disabling is moved into the .editorconfig file. I'll add a comment into the StandardRulesetProvider in that PR to clarify.

PR #497, opened 2019-06-22
Excluding the keyword data from annotation rule

sowmyav24:
Fixes the issue [#490](https://github.com/pinterest/ktlint/issues/490)

Annotation rule looks for element type belonging to `MODIFIER_LIST`. The keyword `data` also falls under this element type. Excluding it explicitly from the list.


shashachu:
Thank you!

PR #499, opened 2019-06-22
Consider comments when searching for multiline annotations

### LINKED ISSUE #498: AnnotationOnSeparateLine should respect comments
Following file has three annotations with a comment next to it: 
https://github.com/arturbosch/detekt/blob/8ae133c2876a94814973ba397c6809cd4a5e514b/detekt-test/src/main/kotlin/io/gitlab/arturbosch/detekt/test/ThresholdedCodeSmellAssert.kt#L8-L9

Unfortunately KtLint does not respect this.

The first testcase in AnnotationRuleTest modified to get a violation reported:
```
    @Test
    fun `lint single annotation may be placed on line before annotated construct`() {
        assertThat(
            AnnotationRule().lint(
                """
                @Suppress("UnsafeCast") // False positive, see issue #1137 
                class A {
                    @JvmField
                    var x: String
                }
                """.trimIndent()
            )
        ).isEmpty()
    }
```

The check which fails for comments:
https://github.com/pinterest/ktlint/blob/a726c87514de5a9ef547529bb4bff299df54dd4d/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRule.kt#L50-L52

arturbosch:
Closes #498 

Edit: Well after reading the property name again, do we want single parameter annotations also be on separate lines?
Imo the checks need a `it.valueParameters.size > 1`.

### DIFF
--- a/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRuleTest.kt
+++ b/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRuleTest.kt
@@ -23,6 +23,18 @@ class AnnotationRuleTest {
     }
 
     @Test
+    fun `lint single annotation with parameters ends with a comment`() {
+        assertThat(
+            AnnotationRule().lint(
+                """
+                @Suppress("AnnotationRule") // this is a comment
+                class A
+                """.trimIndent()
+            )
+        ).isEmpty()
+    }
+
+    @Test
     fun `format single annotation may be placed on line before annotated construct`() {
         val code =
             """
--- a/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRuleTest.kt
+++ b/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRuleTest.kt
@@ -35,6 +35,28 @@ class AnnotationRuleTest {
     }
 
     @Test
+    fun `lint single annotation with parameters on the same line with a comment`() {
+        assertThat(
+            AnnotationRule().lint(
+                """
+                @Suppress("AnnotationRule") /* this is a comment */ class A
+                """.trimIndent()
+            )
+        ).hasSize(1)
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


shashachu:
> Closes #498
> 
> Edit: Well after reading the property name again, do we want single parameter annotations also be on separate lines?
> Imo the checks need a `it.valueParameters.size > 1`.

https://kotlinlang.org/docs/reference/coding-conventions.html#annotation-formatting says
> A single annotation without arguments may be placed on the same line as the corresponding declaration

so yes, it does seem like this should apply to multiple annotations or annotations with arguments.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRuleTest.kt
```diff
@@ -22,6 +22,18 @@ class AnnotationRuleTest {
         ).isEmpty()
     }
 
+    @Test
+    fun `lint single annotation with parameters ends with a comment`() {
+        assertThat(
+            AnnotationRule().lint(
+                """
+                @Suppress("Annotationule") // this is a comment
```
shashachu: nit: typo

arturbosch:
> > Closes #498
> > Edit: Well after reading the property name again, do we want single parameter annotations also be on separate lines?
> > Imo the checks need a `it.valueParameters.size > 1`.
> 
> https://kotlinlang.org/docs/reference/coding-conventions.html#annotation-formatting says
> 
> > A single annotation without arguments may be placed on the same line as the corresponding declaration
> 
> so yes, it does seem like this should apply to multiple annotations or annotations with arguments.

Yep, the statement is alright then with `it.valueArgumentList != null`

ddulaney:
Hello! Sorry to be digging up an old PR, but is it possible to allow **multiple** annotations with comments? For example, this is failing:
```kotlin
@Annotation1 // comment 1
@Annotation2 // comment 2
class MyClass {
}
```
with this error:
```
Multiple annotations should not be placed on the same line as the annotated construct
```

romtsn:
@ddulaney should be resolved by #740

ddulaney:
Great, thanks! I'll wait for that to be merged & released 👍

PR #500, opened 2019-06-23
Add "printAST" sub command.

Tapchicoma:
Move old one (--print-ast) option to be a ktlint sub command - new syntax:
`ktlint printAST`. Old one `--print-ast` is still working, but
deprecated.

### DIFF
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/PrintASTSubCommand.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/PrintASTSubCommand.kt
@@ -1,0 +1,82 @@
+package com.pinterest.ktlint.internal
+
+import com.pinterest.ktlint.KtlintCommandLine
+import com.pinterest.ktlint.core.ParseException
+import com.pinterest.ktlint.core.RuleSet
+import com.pinterest.ktlint.test.DumpAST
+import java.io.File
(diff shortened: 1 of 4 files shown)


ktlint/src/main/kotlin/com/pinterest/ktlint/internal/PrintASTSubCommand.kt
```diff
@@ -0,0 +1,82 @@
...
+        } catch (e: Exception) {
+            if (e is ParseException) {
+                throw ParseException(e.line, e.col, "Not a valid Kotlin file (${e.message?.toLowerCase()})")
+            }
+            throw e
+        }
+    }
+
+    companion object {
+        internal const val COMMAND_NAME = "printAST"
```
rom4ek: May I ask what's the reasoning behind using a camel-cased command name instead of good old '--do-something' pattern for CLI tools? The old way just looks more natural imo.

Tapchicoma: Main reason to move into subcommands direction - reduce user confusion and make command line usage more clear.

For example, now with only options, it is not clear if using `--print-ast` option will also do linting.

Though I am open to subcommand name changes as I don't have strong opinion here.

PR #503, opened 2019-06-26
Add support for disabled_rules property in .editorconfig for globally disabling rules

### LINKED ISSUE #208: Support for globally disabling a rule

shashachu:
Fixes #208

* Takes in a comma-separated list of rule ids, with namespaces (e.g. `experimental:indent`)
* Re-enabled `NoWildcardImports`, `PackageNameRule`
* Un-commented `AnnotationRule`, `MultiLineIfElseRule`, and `NoItParamInMultilineLambdaRule`, and moved disabling into default `.editorconfig`
* Moved `.editorconfig` parsing into `ktlint-core` so that plugins that don't invoke ktlint on the command line can take advantage of rule disabling
* Also cleaned up params passed to lint and format

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,6 +12,11 @@ insert_final_newline = true
 
 [*.{java,kt,kts,scala,rs,xml,kt.spec,kts.spec}]
 indent_size = 4
+# annotation - "./mvnw clean verify" fails with "Internal Error")
+# multiline-if-else - disabled until auto-correct is working properly
+#   (e.g. try formatting "if (true)\n    return { _ ->\n        _\n}")
+# no-it-in-multiline-lambda - disabled until it's clear what to do in case of `import _.it`
+disabled_rules=annotation,multiline-if-else,no-it-in-multiline-lambda
 
 [{Makefile,*.go}]
 indent_style = tab
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
@@ -1,8 +1,14 @@
 package com.pinterest.ktlint.core
 
 import com.pinterest.ktlint.core.ast.prevLeaf
+import com.pinterest.ktlint.core.internal.EditorConfigInternal
+import java.io.File
+import java.nio.file.Path
+import java.nio.file.Paths
 import java.util.ArrayList
 import java.util.HashSet
+import java.util.concurrent.ConcurrentHashMap
+import org.jetbrains.kotlin.backend.common.onlyIf
 import org.jetbrains.kotlin.cli.common.CLIConfigurationKeys
 import org.jetbrains.kotlin.cli.common.messages.MessageCollector
 import org.jetbrains.kotlin.cli.jvm.compiler.EnvironmentConfigFiles
@@ -40,6 +46,27 @@ object KtLint {
 
     private val psiFileFactory: PsiFileFactory
     private val nullSuppression = { _: Int, _: String, _: Boolean -> false }
+
+    /**
+     * @param fileName path of file to lint/format
+     * @param text Contents of file to lint/format
+     * @param ruleSets a collection of "RuleSet"s used to validate source
+     * @param userData Map of user options
+     * @param cb callback invoked for each lint error
+     * @param script true if this is a Kotlin script file
+     * @param editorConfigPath optional path of the .editorconfig file (otherwise will use working directory)
+     * @param debug True if invoked with the --debug flag
+     */
+    data class Params(
+        val fileName: String? = null,
+        val text: String,
+        val ruleSets: Iterable<RuleSet>,
+        val userData: Map<String, String> = emptyMap(),
+        val cb: (e: LintError, corrected: Boolean) -> Unit,
+        val script: Boolean = false,
+        val editorConfigPath: String? = null,
+        val debug: Boolean = false
+    )
 
     init {
         // do not print anything to the stderr when lexer is unable to match input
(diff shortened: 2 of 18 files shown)


ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/EditorConfigInternal.kt
```diff
@@ -8,8 +8,8 @@ import java.util.Properties
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.ConcurrentHashMap
 
-class EditorConfig private constructor (
-    val parent: EditorConfig?,
+class EditorConfigInternal private constructor (
```
shashachu: Totally open for better names for this class.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -93,65 +120,25 @@ object KtLint {
...
      * Check source for lint errors.
      *
-     * @param text source
-     * @param ruleSets a collection of "RuleSet"s used to validate source
-     * @param cb callback that is going to be executed for every lint error
-     *
      * @throws ParseException if text is not a valid Kotlin code
      * @throws RuleExecutionException in case of internal failure caused by a bug in rule implementation
      */
-    fun lint(text: String, ruleSets: Iterable<RuleSet>, cb: (e: LintError) -> Unit) {
```
shashachu: Removed all the overrides and just pass in `Params` now.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -172,7 +159,59 @@ object KtLint {
         }
         errors
             .sortedWith(Comparator { l, r -> if (l.line != r.line) l.line - r.line else l.col - r.col })
-            .forEach(cb)
+            .forEach { e -> params.cb(e, false) }
+    }
+
+    private fun userDataResolver(editorConfigPath: String?, debug: Boolean): (String?) -> Map<String, String> {
```
shashachu: Moved from `Main.kt`

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/EditorConfigInternal.kt
arturbosch: What's wrong with just EditorConfig ;) ?

shashachu: @arturbosch there's already an `EditorConfig` interface inside of `ktlint-core` so if you want to use one or the other you need a lot of ugly namespacing. :) We can also rename that one, but I couldn't think of a name for that one!

arturbosch: Oh, didn't know that.
I like to use a `Default`for my implementations of an interface.
So maybe consider `DefaultEditorConfig`.

rom4ek: @arturbosch but that's the thing - the internal one is actually not an implementation of the `EditorConfig` interface ;)

shashachu: haha correct! I don't have context about why it was implemented this way. I thought about naming the internal one `ParsedEditorConfig` or `EditorConfigParser` or something...

shashachu:
@jaredsburrows @Tapchicoma @jeremymailen @JLLeitschuh @bethcutler Thoughts on this PR?

ktlint-core/src/test/kotlin/com/pinterest/ktlint/core/DisabledRulesTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        )
+
+        assertThat(
+            ArrayList<LintError>().apply {
+                KtLint.lint(
+                    KtLint.Params(
+                        text = "var foo",
+                        ruleSets = listOf(RuleSet("standard", NoVarRule())),
+                        cb = { e, _ -> add(e) },
+                        userData = mapOf(("disabled_rules" to "no-var"))
```
jeremymailen: I love the ability to pass in the disabled_rules as a regular userData param or via editor config 👍

.editorconfig
```diff
@@ -12,6 +12,11 @@ insert_final_newline = true
 
 [*.{java,kt,kts,scala,rs,xml,kt.spec,kts.spec}]
 indent_size = 4
+# annotation - "./mvnw clean verify" fails with "Internal Error")
+# multiline-if-else - disabled until auto-correct is working properly
+#   (e.g. try formatting "if (true)\n    return { _ ->\n        _\n}")
+# no-it-in-multiline-lambda - disabled until it's clear what to do in case of `import _.it`
+disabled_rules=annotation,multiline-if-else,no-it-in-multiline-lambda
```
jeremymailen: When a rule is not prefixed with a`namespace:` is it assumed to be in the standard set?

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -93,65 +120,25 @@ object KtLint {
...
+        val psiFileName = if (params.script) "file.kts" else "file.kt"
+        val psiFile = psiFileFactory.createFileFromText(psiFileName, KotlinLanguage.INSTANCE, normalizedText) as KtFile
         val errorElement = psiFile.findErrorElement()
         if (errorElement != null) {
             val (line, col) = positionByOffset(errorElement.textOffset)
             throw ParseException(line, col, errorElement.errorDescription)
         }
         val rootNode = psiFile.node
-        injectUserData(rootNode, userData)
+        val mergedUserData = params.userData + userDataResolver(params.editorConfigPath, params.debug)(params.fileName)
```
jeremymailen: Does this mean that editor config overrides userData passed in? If so maybe it's more intuitive to reverse it?

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -254,6 +294,10 @@ object KtLint {
         }
     }
 
+    private fun filterDisabledRules(rootNode: ASTNode, fqRuleId: String): Boolean {
+        return rootNode.getUserData(EDITOR_CONFIG_USER_DATA_KEY)?.disabledRules?.contains(fqRuleId) == false
```
jeremymailen: I'm probably missing something, but why is it looking at the editor config key while executing the rules? I thought it resolved any params from the editor config into a merged userData upstream of this?

.editorconfig
rom4ek: Yep, see [this](https://github.com/pinterest/ktlint/pull/503/files#diff-99e7e840295caf6ea84c484c30d73158R244)

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
shashachu: It's kind of roundabout, but the values in the userData map gets parsed into the members of the `EditorConfig` class. Now that you mention it, I'm not actually sure if this is required, or if we could just access the userData map directly. I was mostly trying to follow existing patterns. I think maybe it's not actually necessary because no individual rules will actually need to access the disabled rules, which is what it seems like the EditorConfig class is used for. I'll try to take a look tonight and see if I can simplify this code.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
shashachu: Hm, good question. I don't have a strong opinion either way.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
shashachu: I've moved it into its own key now.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
shashachu: Reversed it.

JorgeCastilloPrz:
Does it require some docs on README?

shashachu:
Yes; I like to wait to update the README until we're ready to release so it doesn't look like features are available that have not yet released.

elenigen:
Do we know when is planned the next release of ktlin to include this new feature of allowing to disable rules?

shashachu:
> Do we know when is planned the next release of ktlin to include this new feature of allowing to disable rules?

@elenigen I am targeting the end of this week.

Issue #306, opened 2018-10-30
NoItParamInMultilineLambdaRule doesn't account for variables or functions that could be named "it"


Trixmegist:
NoItParamInMultilineLambdaRule simply checks that there're no identifiers named "it" inside a multiline lambda. That makes it impossible to use this rule with several testing frameworks, e.g. Kotlintest, because they require to use "it()" functions inside the specs:

```
class MySpec : DescribeSpec({
    describe("if this") {
        it("does that") { // NoItParamInMultilineLambdaRule doesn't allow "it"
        }
    })
```

You already have a "fixme" in the source code of the rule, so I'm wondering when is this going to be fixed?
```
class NoItParamInMultilineLambdaRule : Rule("no-it-in-multiline-lambda") {

    override fun visit(...) {
        // fixme: we are not accounting for "it" variable that can be defined in the same scope
        ...
    }
}
```

ofdrykkja:
from which code-style is this rule? can't find this suggestion anywhere expect ktlint

Trixmegist:
> from which code-style is this rule? can't find this one anywhere expect ktlint

https://github.com/shyiko/ktlint/blob/master/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/NoItParamInMultilineLambdaRule.kt

ofdrykkja:
i mean what document is this rule based on? i asked because didn't find nothing similar nor in android code style guide nor in kotlin code style guide

ofdrykkja:
just can't understand for whats reason i must specify explicitly name of "it" in multiline lambda case

shyiko:
@ofdrykkja https://github.com/yole/kotlin-style-guide/issues/10#issue-157736558

sowmyav24:
The `// fixme: we are not accounting for "it" variable that can be defined in the same scope` has been addressed in this [PR #507](https://github.com/pinterest/ktlint/pull/507) along with accounting `it` as a function call for frameworks such as `Kotlintest`

In `StandardRuleSetProvider`, `NoItParamInMultilineLambdaRule` is disabled for reason `// disabled until it's clear what to do in case of 'import _.it' `. Allowing function call would solve this issue. Is there anything else that is missing for this rule to be enabled ?

sowmyav24:
We can close this issue as the rule is no longer supported

PR #507, opened 2019-06-29
Allow  'it' for variable and function names  in multiline lambda rule

sowmyav24:
Fixes [#306](https://github.com/pinterest/ktlint/issues/306)

In multiline lambda we can now 
- Allow variables named `it`
- function call of `it` for test frameworks like `Kotlintest `
- Ignore when function calls are made on variable `it` or when it is passed in a function.

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoItParamInMultilineLambdaRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoItParamInMultilineLambdaRule.kt
@@ -2,9 +2,15 @@ package com.pinterest.ktlint.ruleset.standard
 
 import com.pinterest.ktlint.core.Rule
 import com.pinterest.ktlint.core.ast.ElementType.BLOCK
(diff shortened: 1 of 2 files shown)


shashachu:
@sowmyav24 thanks for the PR! Do you have a link to the styleguide where it says you should name params in multiline lambdas? I see this section https://kotlinlang.org/docs/reference/coding-conventions.html#lambda-formatting which mentions "When declaring parameter names in a multiline lambda, put the names on the first line, followed by the arrow and the newline" which seems slightly different than "you must name `it` params in multiline lambdas".

I also realize this is an existing rule so you may not have more context than I do.

sowmyav24:
@shashachu I came to know of this only through the open issue. In the issue link, this was the link posted originally. 
https://github.com/pinterest/ktlint/issues/306#issuecomment-462646947

shashachu:
@sowmyav24 it looks like it was based on this discussion, but never made it into the official styleguide: https://github.com/yole/kotlin-style-guide/issues/10.

In that case, I believe we should actually remove this rule from the standard ruleset.

shashachu:
Deleting the rule here: https://github.com/pinterest/ktlint/pull/510

Issue #479, opened 2019-06-10
Indentation on lambda's closing brace


bethcutler:
What's the correct way to format this?
```
foo.func {
  param1, param2 ->
    doSomething()
    doSomething2()
}
```
Intellij formats it as this:
```
foo.func {
  param1, param2 ->
  doSomething()
  doSomething2()
}
```
But ktlint expects this:
```
foo.func {
  param1, param2 ->
    doSomething()
    doSomething2()
  }
```
The solution is a bit more obvious when the params can fit on the same line as the opening brace:
```
foo.func { param1, param2 ->
  doSomething()
  doSomething2()
}
```

shashachu:
Hmm... I don't really know. The IntelliJ and the ktlint format both seem a little bit wrong.

shashachu:
@bethcutler I think IntelliJ is right:
> For curly braces, put the opening brace in the end of the line where the construct begins, and the closing brace on a separate line aligned horizontally with the opening construct.

PR #508, opened 2019-07-03
Fixing bug with extra indentation after lambda arrow

shashachu:
Fixes https://github.com/pinterest/ktlint/issues/479


shashachu:
> LGTM 👍
> 
> Generally, for lambdas, according to [kotlin styleguide](https://kotlinlang.org/docs/reference/coding-conventions.html#lambda-formatting), this rule should format either to:
> 
> ```kotlin
> foo.func { param1, param2 ->
>     doSomething()
>     doSomething2()
> }
> ```
> 
> or, if lambda params can't fit in max line:
> 
> ```kotlin
> foo.func { 
>     param1, 
>     param2
>     ->
>     doSomething()
>     doSomething2()
> }
> ```

Thanks; I did notice that as well. I'll open up another issue about moving params to the same line as the open bracket. It might be difficult to have it not conflict with the max_line_length rule. It looks like IntelliJ only moves the params if they're on the same line as the arrow, so we can follow that rule as well.

PR #510, opened 2019-07-03
Deleting NoItParamInMultilineLambdaRule

shashachu:
It was based on a styleguide discussion (yole/kotlin-style-guide#10) that never made it into the official styleguide.


Zhuinden:
Well that's kind of disappointing.

I'll just leave this here: https://twitter.com/kotlin/status/1220383288109535232

Tapchicoma:
@Zhuinden you could add it as a custom rule. If you think that it should be in standard ruleset, please open a new issue.

PR #511, opened 2019-07-03
Fix typo from IntentStyle to IndentStyle in EditorConfig

caseykulm:
This seems to be a typo of `IntentStyle` instead of `IndentStyle` as it's referenced by the variables below it.


shashachu:
Thanks, @caseykulm! I actually made the same fix as part of this larger PR, which should get merged today: https://github.com/pinterest/ktlint/pull/503

Issue #506, opened 2019-06-28
False positive from UnusedImports on org.gradle.kotlin.dsl.provideDelegate


vlsi:
See https://github.com/arturbosch/detekt/issues/1608

shashachu:
I'm not sure this is a solvable problem. There's no way via static analysis we can tell that `provideDelegate` is used in that file. In this very specific case, I'd recommend a `ktlint-disable` directive to suppress the error.

vlsi:
I wonder if `provideDelegate` can be added to default excludes.

I see it might require compilation in order to tell if the import is used or not, however it is really sad that default `spotless` + `ktlint` configuration produces false positives on `provideDelegate` call.

vlsi:
Alternative option is to allow `provideDelegate` if source code contains at least one `by`

JLLeitschuh:
Could you check the type of `provideDelegate` to see if it's a delegate type and leave it in place if it is?

shashachu:
> Alternative option is to allow `provideDelegate` if source code contains at least one `by`

Yes I see that detekt has just whitelisted that import. We could certainly do that. 

> Could you check the type of `provideDelegate` to see if it's a delegate type and leave it in place if it is?

@JLLeitschuh I'm not sure what you mean here.

vlsi:
> Yes I see that detekt has just whitelisted that import. We could certainly do that

or even "whitelist it provided there's ` by ` in the source code"

shashachu:
It's not very sophisticated, but I put up a PR for this.

PR #513, opened 2019-07-05
Allow unreferenced provideDelegate if there is a by keyword in the file

shashachu:
Fixes https://github.com/pinterest/ktlint/issues/506


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRule.kt
```diff
@@ -44,18 +45,39 @@ class NoUnusedImportsRule : Rule("no-unused-imports") {
...
     override fun visit(
         node: ASTNode,
         autoCorrect: Boolean,
         emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit
     ) {
         if (node.isRoot()) {
+            rootNode = node
             ref.clear() // rule can potentially be executed more than once (when formatting)
             ref.add("*")
+            imports.clear()
```
shashachu: This is an unrelated change, but based on the comment on Line 78, it seems necessary.

Issue #512, opened 2019-07-04
High severity issues in security dependency check


drcabral:
Hi everyone,

I'm working with ktlint and i checked for dependency security issues in my project. In the latest version, ktlint presents 7 high severity issues, what i think that is a lot.

Is there any work going on to take a look and solve these problems? Thanks!

![Screen Shot 2019-07-04 at 18 24 42](https://user-images.githubusercontent.com/6181708/60688051-203baa00-9e89-11e9-97fc-7af563f61a1b.png)

Analysis made with [owasp dependency check](https://github.com/jeremylong/dependency-check-gradle)

shashachu:
Hi! It looks like this is all due to a dependency? Does that dependency have an updated version without whatever security issues?

drcabral:
I'm using ktlint directly in my project, and i updated to the `0.33.0` what i'm thinking that is the last version available and stable. So, Ktlint does not have a new version without security issues i guess.

shashachu:
Sorry, yes I'm one of the maintainers of the project. I was just wondering if you'd looked into solutions at all. We will be releasing 0.34.0 soon and we can aim to include a fix for this.

drcabral:
Thank you!

Tapchicoma:
@drcabral are this security issues raised only for latest release or releases before? And could you provide more details about this issues, like what exactly dependencies have security issues?

drcabral:
Hi @Tapchicoma, I was using the version `0.29.0` and I updated to `0.33.0` thinking that the issues could be solved, but all the same issues still remained.

I can share with you the messages about the issues that the dependency check identified:

```
-> Ktlint-core-0.33.0.jar

Published Vulnerabilities
CVE-2019-1010260  suppress

Using ktlint to download and execute custom rulesets can result in arbitrary code execution as the served jars can be compromised by a MITM. This attack is exploitable via Man in the Middle of the HTTP connection to the artifact servers. This vulnerability appears to have been fixed in 0.30.0 and later; after commit 5e547b287d6c260d328a2cb658dbe6b7a7ff2261.
CVSSv2:
* Base Score: HIGH (9.3)
* Vector: /AV:N/AC:M/Au:N/C:C/I:C/A:C
CVSSv3:
* Base Score: HIGH (8.1)
* Vector: /AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H

References:
* MISC - https://github.com/shyiko/ktlint/pull/332

[...]

PR #514, opened 2019-07-06
Added OWASP Dependency Check

Fleshgrinder:
See #512 where the discussion around vulnerabilities should be kept.

I don‘t know how exactly you want to integrate this into your build and we should add the plugin to Maven as well (although I personally don‘t understand why this project uses Gradle _and_ Maven).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 plugins {
   id 'org.jetbrains.kotlin.jvm' version '1.3.31' apply false
   id 'com.vanniktech.maven.publish' version '0.8.0' apply false
+  id 'org.owasp.dependencycheck' version '5.1.0'
 }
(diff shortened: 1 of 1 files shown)


Tapchicoma:
> although I personally don‘t understand why this project uses Gradle and Maven

Initially project used Maven and had half-working Gradle configuration. Now we are migrating project to use Gradle only.

Adding it only to Gradle setup is ok, as currently both Maven and Gradle builds run on CI.

shashachu:
Is there anything more you'd like to do with this PR?

Fleshgrinder:
@shashachu gauge your interest in adding it and ideas on how to add it to CI. Once we have the latter I'd change it from a draft to a PR.

PR #515, opened 2019-07-07
Add task "shadowJarExecutable" to generate self-running ktlint jar.

Tapchicoma:
This task behaves similar to https://github.com/brianm/really-executable-jars-maven-plugin maven plugin - it creates unix runnable file, that executes `ktlint-all.jar` without need of additional shell script.

Generated file could be checked by executing `ktlint/build/run/ktlint --help` command.

Additional flag `-Xmx512m` was added as per current maven configuration: https://github.com/pinterest/ktlint/blob/e5fd74cba3cb8c12bbd81b8103c5db82f75c6c68/ktlint/pom.xml#L347

### DIFF
--- a/ktlint/build.gradle
+++ b/ktlint/build.gradle
@@ -44,3 +44,22 @@ dependencies {
   testImplementation deps.junit
   testImplementation deps.assertj
 }
+
+// Implements https://github.com/brianm/really-executable-jars-maven-plugin maven plugin behaviour.
+// To check details how it works, see http://skife.org/java/unix/2011/06/20/really_executable_jars.html.
+tasks.register("shadowJarExecutable", DefaultTask.class) {
+  description = "Creates self-executable file, that runs generated shadow jar"
+  group = "Distribution"
+
+  inputs.files tasks.named("shadowJar")
+  outputs.file "$buildDir/run/ktlint"
+
+  doLast {
+    File execFile = outputs.files.singleFile
+    execFile.withOutputStream {
+      it.write "#!/bin/sh\n\nexec java -Xmx512m -jar \"\$0\" \"\$@\"\n\n".bytes
+      it.write inputs.files.singleFile.bytes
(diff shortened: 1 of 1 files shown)


ktlint/build.gradle
```diff
@@ -44,3 +44,20 @@ dependencies {
   testImplementation deps.junit
   testImplementation deps.assertj
 }
+
+tasks.create("shadowJarExecutable") {
```
rom4ek: I think it's better to use `register` here if there are no obstacles, according to [task configuration avoidance guide](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:how_do_i_defer_creation). Cause in case of `create`, the task will be created and configured eagerly.

ktlint/build.gradle
```diff
@@ -44,3 +44,20 @@ dependencies {
...
+  description = "Creates self-executable file, that runs generated shadow jar"
+  group = "Distribution"
+
+  inputs.files tasks.named("shadowJar")
+  outputs.file "$buildDir/run/ktlint"
+
+  doLast {
```
shashachu: Would be nice to add a link to either the maven plugin or the blog post that inspired it, just so it's clear what we're doing here.

ktlint/build.gradle
Tapchicoma: You are totally right, I am always confusing what to use to create a task :+1:

Tapchicoma: Fixed.

ktlint/build.gradle
Tapchicoma: Added comments above task defintion.

ktlint/build.gradle
rom4ek: Gradle things :D

shashachu:
Tested locally and it seems to work. Thanks!

PR #525, opened 2019-07-15
Prep new rules for 0.34.0 release

shashachu:
* Per the project's policy, moving annotation, multiline if/else, and package name to experimental.
* Promoting import ordering to standard.
* renamed `no-first-line-blank-in-method-block-rule` to `no-empty-first-line-in-method-block`


CristianGM:
`import ordering` 😭 but why??

Issue #528, opened 2019-07-16
Can't printAST from stdin (Paths.get(fileName).parent must not be null)


jacquerie:
The very example from the README (https://github.com/pinterest/ktlint#ast) does not appear to work with the latest `ktlint` installed through Homebrew:
```
$ ktlint --version
0.34.0

$ printf "fun main() {}" | ktlint --color printAST --stdin
Exception in thread "main" java.lang.IllegalStateException: Paths.get(fileName).parent must not be null
	at com.pinterest.ktlint.core.KtLint$userDataResolver$2.invoke(KtLint.kt:198)
	at com.pinterest.ktlint.core.KtLint$userDataResolver$2.invoke(KtLint.kt:41)
	at com.pinterest.ktlint.core.KtLint.lint(KtLint.kt:138)
	at com.pinterest.ktlint.internal.FileUtilsKt.lintFile(FileUtils.kt:51)
	at com.pinterest.ktlint.internal.FileUtilsKt.lintFile$default(FileUtils.kt:49)
	at com.pinterest.ktlint.internal.PrintASTSubCommand.printAST(PrintASTSubCommand.kt:69)
	at com.pinterest.ktlint.internal.PrintASTSubCommand.run(PrintASTSubCommand.kt:47)
	at com.pinterest.ktlint.Main.handleSubCommand(Main.kt:76)
	at com.pinterest.ktlint.Main.main(Main.kt:63)
```

shashachu:
@jacquerie Thanks for the report; this is indeed a regression in 0.34.0. We have one other possible regression (#526), so I'll put up a fix for this one, and assuming that one also needs a fix, we can push out a 0.34.1 sometime this week. In the meantime, `printAST` that takes in a filename is still working, so you can use that as a workaround.

PR #529, opened 2019-07-16
Properly handle --stdin flag for printAST command

shashachu:
* Fixes https://github.com/pinterest/ktlint/issues/528
* Also properly pass through `--debug` flag from `printAST` command
* Small cleanup of unused variable

### DIFF
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/PrintASTSubCommand.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/PrintASTSubCommand.kt
@@ -1,6 +1,7 @@
 package com.pinterest.ktlint.internal
 
 import com.pinterest.ktlint.KtlintCommandLine
+import com.pinterest.ktlint.core.KtLint
 import com.pinterest.ktlint.core.ParseException
 import com.pinterest.ktlint.core.RuleSet
 import com.pinterest.ktlint.test.DumpAST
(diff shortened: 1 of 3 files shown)


ktlint/src/main/kotlin/com/pinterest/ktlint/internal/PrintASTSubCommand.kt
```diff
@@ -77,6 +77,6 @@ internal class PrintASTSubCommand : Runnable {
 
     companion object {
         internal const val COMMAND_NAME = "printAST"
-        private const val STDIN_FILE = "<stdin>"
+        private const val STDIN_FILE = "<text>"
```
shashachu: the code explicitly uses this as a flag for stdin. we should probably move this into a constant somewhere.

Tapchicoma: Sorry, missed that as it is not a constant. Yes, you are right - would be better to add `const val STDIN_FILE_NAME = "<stdin>"` to https://github.com/pinterest/ktlint/blob/master/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt and use it here and in other places, so in the future it will not happen again.

Generally, I think, we should bring some integration test to command line module.

shashachu: @Tapchicoma I'll make that fix in a separate PR.

shashachu: @Tapchicoma updated with the constant

Issue #531, opened 2019-07-17
Extension imports from 3rd party library detected as "Unused import"


tinder-mhernandez:
I have an Android library module that depends on the `toSpannable()` extension from Android `core-ktx` and version `0.34.0` of ktlint is detecting `import androidx.core.text.toSpannable` as an unused import. When I run the formatter, it removes the import which then cause the code to fail compilation.

Sample Repro:
```Kotlin
package com.foo

import android.text.Spannable
import androidx.core.text.toSpannable // <-- Considered as unused

fun foo(text: String): Spannable {
    return text.toSpannable()
}
```

tinder-mhernandez:
No repro on `0.33.0` Only seems to repro on `0.34.0`.

shashachu:
Thanks, Marcelo. This is the second regression reported on this rule in 0.34.0 so I'll take a look but if it's not obvious what's going on, I'll go ahead and revert the changes we made in this release. Sorry for the inconvenience!

PR #532, opened 2019-07-17
Revert "Report used import when parent reference is already present (#437)"

shashachu:
This reverts commit 238b774b86cdf6c96dbb8a84d250e7c4a20ac7e7.

Fixes https://github.com/pinterest/ktlint/issues/531, re-introduces https://github.com/pinterest/ktlint/issues/405

@sowmyav24 FYI I'm reverting this diff because it seems to have caused a couple regressions. I'd prefer the false-negative of the original issue the diff was fixing to the false positive of the regression.


sowmyav24:
@shashachu Sure will try to them to to the test cases

Issue #533, opened 2019-07-17
Use cli param value as a source of globally disabled rules


idntfy:
**Context**
* #503 - adds support for globally disabled rules through editorconfig
    >\# Comma-separated list of rules to disable (Since 0.34.0)
    >\# Note that rules in any ruleset other than the standard ruleset will need to be prefixed 
    >\# by the ruleset identifier.
    >disabled_rules=no-wildcard-imports,experimental:annotation,my-custom-ruleset:my-custom-rule

* #48 - summarising, is a very long discussion about abusing the authority by enabling no-wildcard-import by default in the tool we all love.

**This issue is an attempt to find a pragmatic consensus.**

Although it would seem that #503 could resolve the situation around #48 but it still hurts the wide adoption of the tool.

On a scale it's really hard to adopt now when it's enabled. We have both big number of projects and even more bigger number of developers. Committing .editorconfig to all the projects or asking everyone to call --apply-to-idea-project are just no way to go, because it cannot be centralized thus hardly manageable.

**The Proposal 1**

Introduce a cli param value of which would be used as a source to the functionality already introduced in #503 - so basically it's nothing new.

In this form it can be configured widely even across many projects, e.g. if parent-pom or some common maven/gradle plugin is used, without committing .editorconfig to all the projects and without asking an every single developer to change default idea settings by calling --apply-to-idea-project.

**The Proposal 2**

Create several sets of code styles:
* default, Jetbrains-based on official kotlin coding conventions and intelliJ idea default settings - this one is not supported now.
* android-specific, which would match google's recommendation for android development - it seems ktlint tends to support this standard.

Make it switchable through cli param. Thus all the parties - android developers and the rest - would use what they need.

----

Bottomline: while 2 could be too extreme, 1 is just a simple addition to the functionality that already exists.

shashachu:
An `--android` CLI param actually already exists, but it's not being used in this capacity. I don't think your proposal to move the no wildcards rule into and `android` ruleset is a bad, as it does exist only in the Android Kotlin styleguide. I admit that as an Android developer I'm biased towards that direction, but I don't have a good pulse on how non-Android Kotlin developers work. The reason I'm hesitant is that it opens the door to a lot of arguing about whether a rule should be android-only, and as far as I know, the no wildcard imports rule would be the only rule in the android ruleset. Let me give this some thought.

JakeWharton:
[...]

PR #534, opened 2019-07-17
Add --disabled_rules flag to CLI

shashachu:
Fixes https://github.com/pinterest/ktlint/issues/533

Note: command line flag will override `.editorconfig`


shashachu:
@idntfy FYI. thanks for the feedback.

Issue #526, opened 2019-07-15
Import alias flagged as unused import


gabrielittner:
Something like `import a.b.c.MyClass as SomeOtherName` is marked as an unused import after upgrading to 0.34.0.

shashachu:
@gabrielittner this is a regression in 0.34.0?

gabrielittner:
Yes. 0.33 was working without this issue on the same code

shashachu:
@gabrielittner could you provide a more complete code snippet? it looks like the rule attempts to handle import aliases (and tests that have import aliases are passing), so I want to understand what broke.

gabrielittner:
I will try to put together something reproducible tomorrow

shashachu:
@gabrielittner I'm going to revert the changes to the unused imports rule that we made in 0.34.0, and I suspect it will fix your issue, but when you have a repro case let me know and I can test.

shashachu:
@gabrielittner if you can provide your repro case, #535 is attempting to fix the regressions in 0.34.0, so we want to make sure your case is covered.

echorebel:
i also have this case.
`import com.product....JavaClass as AnotherName` in a Kotlin unit test in our case
will be removed with 0.34.
I might also try to put something together in case @gabrielittner doesn't have something already.

sowmyav24:
Similar example was provided in the PR review. It has been added as a test case and fixed now.

romtsn:
Fixed via #535

PR #535, opened 2019-07-18
Report unused imports when parent reference of the static import present

sowmyav24:
Fixes [#405 ](https://github.com/pinterest/ktlint/issues/405)
This PR contains the changes done in[#437 ]( https://github.com/pinterest/ktlint/pull/437) and in addition to it has the scenario covered that was reported in [#531](https://github.com/pinterest/ktlint/issues/531) .

Previously the redundant import did not consider the full import path. Added the check in this PR.

There is another issue reported in [#526 ](https://github.com/pinterest/ktlint/issues/526). Once the steps to reproduce is given, will add it as a test case in this PR to ensure no other regressions are caused.


ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRuleTest.kt
```diff
@@ -316,4 +441,22 @@ class NoUnusedImportsRuleTest {
...
+        assertThat(
+            NoUnusedImportsRule().lint(
+                """
+                package com.foo
+
+                import android.text.Spannable
+                import androidx.core.text.toSpannable
+
+                fun foo(text: String): Spannable {
+                    return text.toSpannable()
```
shashachu: This is kind of a funny case because I realized that when I renamed the `text` parameter of `foo` to anything else, lint no longer fired. In this case it's not even a static import, it's just a function call with a variable that happens to match the name of an extension functino.

sowmyav24: Yes,  since there was import ending with `text.toSpannable`, the code asssumed it to be a static import. 😞 
In this PR, in addition to the prev fixes, just made sure parent import (with full package name check) exists.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRule.kt
```diff
@@ -112,5 +126,43 @@ class NoUnusedImportsRule : Rule("no-unused-imports") {
         }
     }
 
+    // Builds a list of imports. Takes care of having aliases in case it is assigned to imports
+    private fun buildNonRedundantImports(text: String) {
+        val element = text.split("import").last().trim()
+        if (element.contains("as ")) {
+            imports.addAll(listOf(element.split("as").first().replace("`", "").trim(), element.split("as").last().trim()))
```
gersonsosa:
```suggestion
            imports.addAll(listOf(element.split(" as ").first().replace("`", "").trim(), element.split(" as ").last().trim()))
```

There is a bug here when the import contains the letter "as" for example `com.fastcompany.foo.Anything` this will cut the whole import in the wrong place and end up reporting unused import

gersonsosa: If you use the following test is going to fail
```
    @Test
    fun `Should not report unused here`() {
        assertThat(
            NoUnusedImportsRule().lint(
                """
                package com.fastcompany.foo

                import com.fastcompany.common.Awesome as CommonAsset

                class SomeConfig {
                
                    fun fooFunction() = CommonAsset()
                }

                """.trimIndent()
            )
        ).isEmpty()
    }
```

sowmyav24: Thank you for this test ! In addition to it, i changed the contains condition above it also to `contains(" as ") ` and added one more test to check that.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRule.kt
```diff
@@ -112,5 +126,43 @@ class NoUnusedImportsRule : Rule("no-unused-imports") {
         }
     }
 
+    // Builds a list of imports. Takes care of having aliases in case it is assigned to imports
+    private fun buildNonRedundantImports(text: String) {
+        val element = text.split("import").last().trim()
+        if (element.contains(" as ")) {
```
shashachu: there is an ElementType for the `as` keyword: `~.c.i.p.impl.source.tree.LeafPsiElement (AS_KEYWORD) "as"`

Is it possible to check that instead? Seems slightly more robust than text matching.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRule.kt
```diff
@@ -112,5 +126,43 @@ class NoUnusedImportsRule : Rule("no-unused-imports") {
...
+            imports.addAll(listOf(element.split(" as ").first().replace("`", "").trim(), element.split("as").last().trim()))
```
shashachu: should that second param be checking for `" as "` too?

sowmyav24: Pulled the `" as "`  into a variable so as to avoid misses in check

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRule.kt
sowmyav24: That would help. But in this function, the import is sent as a String. The presence of " as " keyword comes in the subsequent node visits.

PR #561, opened 2019-08-15
Replace editorconfig parser to ec4j.

Tapchicoma:
Replace custom editorconfig parser to https://github.com/ec4j/ec4j.


Tapchicoma:
Not sure if we still support maven builds. If yes, I will add ec4j dependency to maven as well.

PR #566, opened 2019-08-27
Remove maven dependency resolver.

### LINKED ISSUE #451: Remove MavenDependencyResolver

Tapchicoma:
Also has to remove classpath check as jar does not contain any dependecies.

Classpath loading support for Java 9+ could be implemented in some another PR (possibly via #495).

Closes #451


shashachu:
great, thank you!

cc @bethcutler

Issue #582, opened 2019-09-16
Customize console output color?


AdamMc331:
I was looking at the console logs of one of my builds recently that had KtLint errors, and at first they didn't stand out to me. I have the color option on, but noticed that when we color things, they're gray by default: https://github.com/pinterest/ktlint/blob/master/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporter.kt#L56

Against a black console, this doesn't really stand out to me. Other code quality tools like Detekt will output these using a sharper color like yellow. 

I'm curious if we could change the color to something that stands out a little more against a dark console, or if the output color is something that could be customized? 

An example of how hard this was to read:

<img width="1077" alt="Screen Shot 2019-09-16 at 9 35 05 AM" src="https://user-images.githubusercontent.com/9515997/64966796-7192ed00-d86d-11e9-9661-a773a05fd5c4.png">

Comparison to Detekt:

<img width="1111" alt="Screen Shot 2019-09-16 at 9 35 22 AM" src="https://user-images.githubusercontent.com/9515997/64966831-80799f80-d86d-11e9-9569-f4e3cce55ed7.png">

shashachu:
Well it wouldn't be very plain then, would it? :)

I don't have strong feelings about this either way - you could easily write another custom reporter with nice colors, or this could be customizable...maybe by passing a color to the `--color` flag? e.g. `ktlint --color="(255, 0, 0)"` or something

AdamMc331:
I like that idea! I notice we already have a bunch of colors defined here: https://github.com/pinterest/ktlint/blob/master/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/internal/Color.kt

Ideally I could reference these, and maybe we can do `--color="RED"` and parse the color from that name. If no name is supplied, it could just default to the DARK_GRAY` we already use. 

I don't know much about how KtLint works now and it may take me a while to figure out how to implement this but I will try. If anyone sees this and knows a quick way to the solution feel free to take it from me. :)

AdamMc331:
Here's my take on that: https://github.com/pinterest/ktlint/pull/585

Tapchicoma:
Closed via #585

PR #584, opened 2019-09-17
Add support for custom output colors

AdamMc331:
## Summary

I recently submitted this issue because the console output colors were difficult to read against a dark terminal: #582 

For a proposed solution, I added a new flag called `outputColor` that end users can supply any color name found here: https://github.com/pinterest/ktlint/blob/master/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/internal/Color.kt

If not supplied, or an invalid name is supplied, we default to DARK_GRAY, which was the hardcoded value from before. 

## How This Was Tested

I've added unit tests on `PlainReporterProvider` which verify the color values passed in the map of strings get mapped to the correct values. I also added a new test on `PlainReporter` which verifies the color output is what we expect. 

I also tested this manually. Here it is with just `--color` set:

Screen Shot 2019-09-17 at 12 20 09 PM

And here is one with `--color --outputColor="RED"`:

Screen Shot 2019-09-17 at 12 20 34 PM

### DIFF
--- a/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporter.kt
+++ b/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporter.kt
@@ -13,7 +13,8 @@ class PlainReporter(
     val out: PrintStream,
     val verbose: Boolean = false,
     val groupByFile: Boolean = false,
-    val color: Boolean = false,
+    val shouldColorOutput: Boolean = false,
+    val outputColor: Color = Color.DARK_GRAY,
     val pad: Boolean = false
 ) : Reporter {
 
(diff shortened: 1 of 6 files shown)


AdamMc331:
I would have loved to add something like `--color="RED"` instead, but given the way command line arguments I couldn't quite figure it out. 

I started by changing the property type of color to `String`, but then if I just put `--color` as an argument, it told me I didn't have a value. 

If I supplied a default value, say `DARK_GRAY`, then when I ran without any command line arguments, DARK_GRAY is applied too, which is not the expected behavior. 

If anyone knows how we can tell the difference between:

- no arguments
- --color
- --color="RED"

I think that would be much more user friendly, but I couldn't figure it out so I thought I would try and see how you felt about this solution.

AdamMc331:
Update: I figured out how to handle the above, going to close this and reopen when I've resolved this case.

PR #585, opened 2019-09-17
Added support for customizing the output color for #582.

AdamMc331:
# Summary

This resolves the issue discussed here, that the default DARK_GRAY background color was hard to read against dark terminals: #582 

# Usage

I made sure this was a non breaking change by allowing the current usage of KtLint to behave the way it does now. For example:

- ./ktlint 
  - Will output without color as it does now
- ./ktlint --color
  - Will color the output in DARK_GRAY as it does now
- ./ktlint --color --color-name="RED"
  - Will color the output in RED, or whatever was supplied.

# How it was tested

I've added unit tests to `PlainReporterTest` to validate the colored output. I also added `PlainReporterProviderTest` which ensures the input for a color maps to the expected output. 

I also ran this manually. Here is KtLint with no command args:

Screen Shot 2019-09-17 at 2 07 40 PM

Here it is with `--color`:

Screen Shot 2019-09-17 at 12 20 09 PM

Here it is with `--color --colorName="RED"`:

Screen Shot 2019-09-17 at 12 20 34 PM

### DIFF
--- a/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporterProvider.kt
+++ b/ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporterProvider.kt
@@ -2,6 +2,7 @@ package com.pinterest.ktlint.reporter.plain
 
 import com.pinterest.ktlint.core.Reporter
 import com.pinterest.ktlint.core.ReporterProvider
+import com.pinterest.ktlint.reporter.plain.internal.Color
 import java.io.PrintStream
 
 class PlainReporterProvider : ReporterProvider {
@@ -13,9 +14,14 @@ class PlainReporterProvider : ReporterProvider {
             out,
             verbose = opt["verbose"]?.emptyOrTrue() ?: false,
             groupByFile = opt["group_by_file"]?.emptyOrTrue() ?: false,
-            color = opt["color"]?.emptyOrTrue() ?: false,
+            shouldColorOutput = opt["color"]?.emptyOrTrue() ?: false,
+            outputColor = getColor(opt["color_name"]),
             pad = opt["pad"]?.emptyOrTrue() ?: false
         )
 
     private fun String.emptyOrTrue() = this == "" || this == "true"
+
+    private fun getColor(colorInput: String?): Color {
+        return Color.values().firstOrNull { it.name == colorInput } ?: throw IllegalArgumentException("Invalid color parameter.")
+    }
 }
--- a/ktlint-reporter-plain/src/test/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporterProviderTest.kt
+++ b/ktlint-reporter-plain/src/test/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporterProviderTest.kt
@@ -1,0 +1,68 @@
+package com.pinterest.ktlint.reporter.plain
+
+import com.pinterest.ktlint.reporter.plain.internal.Color
+import java.io.PrintStream
+import java.lang.System.out
+import org.junit.Assert.assertEquals
+import org.junit.Assert.fail
+import org.junit.Test
+
+class PlainReporterProviderTest {
+    @Test
+    fun testNoColorNameProvided() {
+        try {
+            PlainReporterProvider().get(
+                out = PrintStream(out, true),
+                opt = mapOf()
+            ) as PlainReporter
+
+            fail("Expected IllegalArgumentException.")
+        } catch (iae: IllegalArgumentException) {
+            // Expected case
+        } catch (e: Exception) {
+            fail("Expected IllegalArgumentException but was: $e")
+        }
+    }
+
+    @Test
+    fun testEmptyColorNameProvided() {
+        try {
+            PlainReporterProvider().get(
+                out = PrintStream(out, true),
+                opt = mapOf("color_name" to "")
+            ) as PlainReporter
+
+            fail("Expected IllegalArgumentException.")
+        } catch (iae: IllegalArgumentException) {
+            // Expected case
+        } catch (e: Exception) {
+            fail("Expected IllegalArgumentException but was: $e")
+        }
+    }
+
+    @Test
+    fun testValidColorNameProvided() {
+        val plainReporter = PlainReporterProvider().get(
+            out = PrintStream(out, true),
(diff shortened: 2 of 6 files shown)


ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
```diff
@@ -128,9 +128,10 @@ class KtlintCommandLine {
 
     @Option(
         names = ["--color"],
-        description = ["Make output colorful"]
+        description = ["Make output colorful. To specify the color, use --COLOR=RED"],
```
rom4ek: I think it's better to specify lowercase `--color` here, because picocli is case sensitive. You'll end up with `UnmatchedArgumentException` otherwise.

ktlint-reporter-plain/src/main/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporterProvider.kt
```diff
@@ -13,9 +14,16 @@ class PlainReporterProvider : ReporterProvider {
...
+            color = getColor(opt["color"]),
             pad = opt["pad"]?.emptyOrTrue() ?: false
         )
 
     private fun String.emptyOrTrue() = this == "" || this == "true"
+
+    private fun getColor(colorInput: String?): Color? {
+        return when (colorInput) {
+            "null" -> null
+            else -> Color.values().firstOrNull { it.name == colorInput } ?: Color.DARK_GRAY
```
rom4ek: Not sure, maybe it's better not to fallback to gray color here, but behave like the value wasn't supplied? Or maybe even throw an exception that the argument is not matched.

AdamMc331:
@rom4ek Thanks for the feedback! I think those were good changes and I've made the updates. 

Let me know what you think of this. Also if you want me to rebase and squash these commits into one I'd be happy to once we resolve the reviews.

rom4ek:
@AdamMc331 not that I pretended to review, but lgtm, I think it's a great improvement :)

shashachu:
@AdamMc331 thanks for the PR! Can you see what happens if you pass in a color for `./ktlint --color=RED printAST`? Assuming it's not supported but wondering if maybe we should error out, or print a warning about it.

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
```diff
@@ -128,9 +128,10 @@ class KtlintCommandLine {
 
     @Option(
         names = ["--color"],
-        description = ["Make output colorful"]
+        description = ["Make output colorful. To specify the color, use --color=\"RED\""],
+        arity = "0..1"
     )
-    var color: Boolean = false
+    var color: String? = null
```
Tapchicoma: Changing type here will be a breaking change. 

I would propose to introduce additional parameter `--console-main-color` (short `-C`), that will be strongly typed `Color`: https://picocli.info/#_enum_types with default value `Color.NO_COLOR`. Additionally deprecate `--color` parameter.

AdamMc331: @Tapchicoma Do you think I may still run into the problem mentioned here where if I did `--console-main-color printAST` it would be interpreted as the param name? https://github.com/pinterest/ktlint/pull/585#issuecomment-534561994

I can test later to be sure, I'm still learning picocli.

Tapchicoma: This particular problem comes from `arity = "0..1"` that allows to not have option value: https://picocli.info/#_arity. In the case above `printAST` is parsed as optional option value and not as subcommand.

My proposal for `--console-main-color` is to have mandatory option value (basically `arity = "1"`)

AdamMc331: Okay. Should be easy enough. I'm a little iffy about deprecating `--color`. The other alternative could be to introduce a new parameter, `--color-name` but then I think that's kind of annoying to ask the user to specify two parameters and do `--color --color-name RED`. 

Which would you prefer?

Tapchicoma: Maybe yes, let's leave `--color` option that would just enable output coloring with default color (`RED`?) and `--color-name` that could be used to change color value.

AdamMc331:
Okay I've updated this to use a second parameter after all, which should be a non breaking change but provide people with this additional functionality. 

What is the rule when it comes to changelog/README? Should I update those files to account for this change too, or will that happen closer to the release process?

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
```diff
@@ -132,6 +133,12 @@ class KtlintCommandLine {
     )
     var color: Boolean = false
 
+    @Option(
+        names = ["--color-name"],
+        description = ["Customize the output color"]
+    )
+    var colorName: String = Color.DARK_GRAY.name
```
Tapchicoma: Would be nice to use here enum type, but it will require a lot of changes :(

ktlint-reporter-plain/src/test/kotlin/com/pinterest/ktlint/reporter/plain/PlainReporterProviderTest.kt
```diff
@@ -0,0 +1,68 @@
...
+            PlainReporterProvider().get(
+                out = PrintStream(out, true),
+                opt = mapOf()
+            ) as PlainReporter
+
+            fail("Expected IllegalArgumentException.")
+        } catch (iae: IllegalArgumentException) {
+            // Expected case
+        } catch (e: Exception) {
+            fail("Expected IllegalArgumentException but was: $e")
```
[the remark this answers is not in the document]
AdamMc331: Ooo whoops. Want me to fix?

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
AdamMc331: I agree and thought the same but I wasn't sure if I could reference an enum as a command line argument? I assumed it had to be a string. If you know something I don't - let me know and I can try to adjust.

AdamMc331:
Whoops replied while merging happened. Let me know if anything was glaring and we can talk about it as a future follow up. Thanks for reviewing!

PR #587, opened 2019-09-24
Add angle bracket whitespace rule

### LINKED ISSUE #580: Spacing around generics
It doesn’t complain where I would prefer it to.

```
  var a: Map< Int, String> = mapOf()
  var b: Map<Int, String > = mapOf()
  var c: Map <Int, String> = mapOf()
```

christinalee:
Fixes #580

Adds a lint for extra whitespace around type parameters which resolves the previously added failing tests.


ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/SpacingAroundAngleBracketsRule.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.pinterest.ktlint.core.Rule
+import com.pinterest.ktlint.core.ast.ElementType.GT
+import com.pinterest.ktlint.core.ast.ElementType.LT
+import com.pinterest.ktlint.core.ast.ElementType.TYPE_ARGUMENT_LIST
+import com.pinterest.ktlint.core.ast.ElementType.WHITE_SPACE
+import com.pinterest.ktlint.core.ast.nextLeaf
+import com.pinterest.ktlint.core.ast.prevLeaf
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+
+class SpacingAroundAngleBracketsRule : Rule("angle-brackets-rule") {
```
shashachu: nit: please leave a small comment about what this rule does.

shashachu:
@Tapchicoma want to take a look at this one? Feel free to merge it in if it looks good to you.

ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/SpacingAroundAngleBracketsRule.kt
```diff
@@ -0,0 +1,54 @@
...
+        node: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit
+    ) {
+        if (node.elementType == TYPE_ARGUMENT_LIST) {
+            val openingBracketChild = node.nextLeaf { it.elementType == LT }
+            if (openingBracketChild != null) {
+                // Check for rogue spacing before an opening bracket:  Map <String, Int>
+                val beforeAngleBracket = openingBracketChild.prevLeaf()
+                if (beforeAngleBracket != null && beforeAngleBracket.elementType == WHITE_SPACE) {
```
rom4ek: I think this can be simplified to `beforeAngleBracket?.elementType == WHITE_SPACE`, also in other such `ifs`. If you don't have special conventions for that, of course.

rom4ek:
Just popped into my mind, what if the argument list is huge and we want to split the types per line? E.g., this would work with the current implementation:
```
ClassWithGenerics<String,
    Int,
    Boolean,
    ...,
    Boo>
```

But this would already fail:
```
ClassWithGenerics<
    String,
    Int,
    Boolean,
    ...,
    Boo
>
```

So if you want to allow the latter case, then we need to add another check for a newline.

Tapchicoma:
@rom4ek added valid point about covering writing type parameters per line. I've checked kotlin and android style guides and don't found anything there. IDEA allows following formatting:
```kotlin
class AngleTest<
    S,
    G,
    Z,
    B
    > {
    
}
```

Also would be nice to add tests for nested angle brakets cases, for example:
```kotlin
class AngleTest<B : Map<Int, List<String>>>
```

christinalee:
Sounds good on all counts. Will update.

Tapchicoma:
@christinalee hi, any updates on this PR?

romtsn:
Closing this in favour of #769

PR #621, opened 2019-10-17
#445 - clean up repo by removing maven

### LINKED ISSUE #445: Remove Maven, Use Gradle

jaredsburrows:
@shashachu 
@Tapchicoma


shashachu:
Thanks @jaredsburrows! I verified that after updating to 0.36.0-SNAPSHOT (which has this fix https://github.com/pinterest/ktlint/pull/617), I don't have to make any changes to my gradle files, so it should be safe to start making releases via gradle.

.editorconfig
```diff
@@ -13,9 +13,5 @@ insert_final_newline = true
 [*.{java,kt,kts,scala,rs,xml,kt.spec,kts.spec}]
 indent_size = 4
 
-[{Makefile,*.go}]
```
shashachu: This seems unrelated?

jaredsburrows: Reverted.

.announce
```diff
@@ -1,44 +0,0 @@
-#!/bin/bash -e
```
Tapchicoma: I think this file is used to update version in README.md and on the ktlint.github.io site.

Tapchicoma:
Also maven contains step to automatically create a release notes on Github using chandler tool.

jaredsburrows:
@Tapchicoma Should we port the maven step/command over to Gradle?

Tapchicoma:
> @Tapchicoma Should we port the maven step/command over to Gradle?

Yes, would be nice. Then Maven could be removed after test release with Gradle.

.announce
jaredsburrows: Reverted

jaredsburrows:
@Tapchicoma Is there a plugin for this?

Tapchicoma:
As far as I know [chandler](https://github.com/mattbrictson/chandler) itself is not maintained anymore. Propose replacement is https://github.com/apps/release-drafter app. No idea though how could it be integrated with Gradle.

shashachu:
I'm okay without chandler, but the maven steps also include a deploy to GitHub which I'd like to not lose. I had to manually upload and generate the signature.

msfjarvis:
[This plugin](https://github.com/BreadMoirai/github-release-gradle-plugin) seems to work just fine, might be worth evaluating for the use case here.

Tapchicoma:
@msfjarvis thank you, will take a look :+1:

Tapchicoma:
@jaredsburrows could you resolve conflict in `pom.xml` file and update PR with latest changes from `master`?

Chandler tool could be replaced with plugin proposed by @msfjarvis and I add in a separate PR. So this PR could be merged as is.

jaredsburrows:
@Tapchicoma done!

PR #623, opened 2019-10-17
Try to discover `.git` instead of hardcoding.

erik:
Very minor convenience improvement.

Previously we assumed that we were always in the root of the repo for
installing Git hooks. Instead, call `rev-parse --show-toplevel` to
discover the root of the repo, which lets us call
`--install-pre-{commit,push}-hook` from anywhere inside a Git repository.


shashachu:
Thanks for the fix!

PR #626, opened 2019-10-20
Fix version in README.md

rom4ek:
Not sure why it wasn't up-to-date, maybe the `.announce` script wasn't used for the latest version?


Tapchicoma:
Thank you!

.cc @shashachu

PR #630, opened 2019-10-23
Added utf-8 BOM removal

### LINKED ISSUE #272: UTF-8 with BOM results in "Not a valid Kotlin file (expecting a top level declaration)"
When running `ktlint` (0.27.0) on some kotlin file, I get this error:

```kt
package demo 
 
fun main(args : Array<String>) { 
  println("Hello, world!") 
}
```

Error : 

```sh
$ ktlint --color test.kt --debug --print-ast
[DEBUG] Analyzing /tmp/test.kt
Exception in thread "main" com.github.shyiko.ktlint.core.ParseException: Not a valid Kotlin file (expecting a top level declaration)
	at com.github.shyiko.ktlint.Main$printAST$1.invoke(Main.kt:495)
	at com.github.shyiko.ktlint.Main.printAST(Main.kt:504)
	at com.github.shyiko.ktlint.Main.main(Main.kt:262)
```

Raibaz:
This fixes #272 by adding explicit removal of the UTF-8 BOM from the content of the files being parsed; it also introduces a minor refactoring that reduces code duplication.

Note that this also removes the BOM from the output when formatting files; it shouldn't be a big deal, as UTF-8 files can work without it and it is in fact suggested not to include it.

### DIFF
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
@@ -420,13 +420,12 @@ object KtLint {
                 .forEach { (e, corrected) -> params.cb(e, corrected) }
         }
 
-        if(!mutated) {
+        if (!mutated) {
             return params.text
         }
 
-        return if(hasUTF8BOM) UTF8_BOM else "" + // Restore UTF8 BOM if it was present
+        return if (hasUTF8BOM) UTF8_BOM else "" + // Restore UTF8 BOM if it was present
             rootNode.text.replace("\n", determineLineSeparator(params.text, params.userData))
-
     }
 
     private fun determineLineSeparator(fileContent: String, userData: Map<String, String>): String {
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
@@ -126,7 +126,7 @@ object KtLint {
      * @throws RuleExecutionException in case of internal failure caused by a bug in rule implementation
      */
     fun lint(params: Params) {
-        val normalizedText = params.text.replace("\r\n", "\n").replace("\r", "\n")
+        val normalizedText = normalizeText(params.text)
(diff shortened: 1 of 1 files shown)


Tapchicoma:
After reading some discussions over internal, I would say removing UTF-8 BOM symbol on file format is not a nice idea.

Tapchicoma:
Actually I could not reproduce original issue with latests ktlint. @Raibaz have you succeeded in it?

Raibaz:
Why do you think it's not a good idea?

Apparently, the BOM is not mandatory for UTF-8 files and messes up with non-UTF-8 applications (including ktlint) that don't expect the non-ASCII characters at the beginning of the file and try to parse them as ASCII characters, resulting in bad parsing of the file.

My opinion is that removing it from the start of the file is safe, as the bytes in the BOM cannot be there for any other reason that would need them to be there, and their presence is just preventing ktlint from operating as expected.

Do you have any other ideas in mind on how to handle UTF-8 files with BOM? I guess it can probably be dealt with relatively easily when doing just file validation, but when fixing style violations it will likely be trickier, as the whole text content of the file is going to be manipulated.

Tapchicoma:
For example, I've read following issue: https://github.com/editorconfig/editorconfig/issues/297, where some people complain about removing BOM support.

Generally, I would say that it is not responsibility of ktlint to remove BOM on file format. BOM itself does not relate to Kotlin code style and it may happen that people added it intentionally.

> Do you have any other ideas in mind on how to handle UTF-8 files with BOM? I guess it can probably be dealt with relatively easily when doing just file validation, but when fixing style violations it will likely be trickier, as the whole text content of the file is going to be manipulated.

Current approach for removing it in `lint()` method is ok as it is not destructive. For `format()` method BOM could be saved in some String field that would be added back on returning formatted file as string here: https://github.com/pinterest/ktlint/blob/2e7d67cb562c4b4e69c48d4f265a3ec0c1660a56/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt#L413

Raibaz:
Right, makes sense, I updated my PR accordingly.

Tapchicoma:
@Raibaz could you fix code style? Other then that your PR looks good to me.

PR #638, opened 2019-11-07
Implement a rule to lint enum entry name

### LINKED ISSUE #635: Enforce enum names
Hi and thanks for this awesome tool!

I haven't found how to enforce enum names though (e.g. I'd like all the enum names to be camel case with the first uppercase letter). I wonder if I've missed something or simply it's not implemented yet.

From: https://kotlinlang.org/docs/reference/coding-conventions.html#property-names

> For enum constants, it's OK to use either uppercase underscore-separated names (enum class Color { RED, GREEN }) or regular camel-case names starting with an uppercase first letter, depending on the usage.

Thanks in advance!

Monchi:
Resolve #635 

Implement an experimental rule to lint enum entry name.
This rule enforces enum entry names to 
1. contains no lower case letter if it contains under score.
2. starts with upper case if it contains lower case letter.

So, "ENTRY_NAME" and "EntryName" are valid, "Entry_Name", "entry_name" and "entryName" are invalid.

Currently auto-format will convert entry name to all upper case, but this behavior is highly debatable.

#### Before format
```kotlin
enum class FirstEnum {
    enum_entry
}
enum class SecondEnum {
    entry
}
```

#### formatted
```kotlin
enum class FirstEnum {
    ENUM_ENTRY
}
enum class SecondEnum {
    ENTRY
}
```

### DIFF
--- a/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/EnumEntryNameCaseRule.kt
+++ b/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/EnumEntryNameCaseRule.kt
@@ -1,0 +1,59 @@
+package com.pinterest.ktlint.ruleset.experimental
+
+import com.pinterest.ktlint.core.Rule
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.CompositeElement
+import org.jetbrains.kotlin.psi.KtEnumEntry
+
+class EnumEntryNameCaseRule : Rule("enum-entry-name-case") {
+
+    companion object {
+        const val ERROR_MESSAGE = "Enum entry name should be uppercase underscore-separated names like \"ENUM_ENTRY\" or upper camel-case like \"EnumEntry\""
+    }
+
+    override fun visit(
+        node: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit
+    ) {
+        if (node !is CompositeElement) {
+            return
+        }
+        val enumEntry = node.psi as? KtEnumEntry ?: return
+        val name = enumEntry.name ?: return
+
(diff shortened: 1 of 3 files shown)


Monchi:
I forgot an important thing. This format might causes compile error because enum entry identifier will changed.
Maybe formatting should be disabled and just lint.

Tapchicoma:
I would propose to disable formatting for now. 

Generally it is interesting question how to handle such cases that makes code non-compilable. Maybe via some additional configuration flag - `allowBreakingFormatting` :thinking:

shashachu:
Agree that we should disable formatting if it could cause compile errors.

Monchi:
@Tapchicoma @shashachu 
Thank you for your feedback. I disabled formatting for now, but leave formatting logic.
Also, created an issue https://github.com/pinterest/ktlint/issues/656 about supporting `allowBreakingFormatting` option.

ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/EnumEntryNameCaseRule.kt
```diff
@@ -0,0 +1,59 @@
...
+
+        if (name.containsLowerCase()) {
+            // In case of lower camel case like "enumName", or all lower case like "enumname"
+            if (!name.startsWithUpperCase()) {
+                emit(
+                    node.startOffset,
+                    ERROR_MESSAGE, false
+                )
+
+                if (autoCorrect) correct(enumEntry, name)
```
Tapchicoma: This will make breaking format on code change. Maybe you forgot to push disabling formatting?

Monchi: `canBeAutoCorrected` is false, so `autoCorrect` never be `true`.

Tapchicoma: Sorry, missed that :+1:

Issue #608, opened 2019-10-09
AnnotationOnSeparateLine rule on steroids 


winstonvk:
_Duplicate issue from [here](https://github.com/arturbosch/detekt/issues/2001)_

Could you please add field `excludedAnnotations:` to `AnnotationOnSeparateLine` rule?

## Expected Behavior of the rule
example: `excludedAnnotations: 'retrofit2.http.Query,retrofit2.http.Body'`
So, this annotation won't be forced to move on separate line:
```
    @GET("/foo")
    fun bar(
        @Query("id") id: String
    ): Single<MyResponse>
```

Tapchicoma:
For you solution would be to disable "experimental:annotation" rule. 

If still want to to have this rule active, but exclude some annotations from it, you could copy [the code](https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt) of this rule and create your custom one.

Tapchicoma:
In next release "experimental:annotation" rule will allow to have annotation on the same line.

PR #642, opened 2019-11-11
Allow single-line annotations in parameters/arguments

### LINKED ISSUE #556: AnnotationRule formatting uncertainties
I have a following Retrofit interface:

```Kotlin
interface UserService {

    fun updateMarketingConsent(
        @Path("userId") userId: String,
        @Path("context") context: String,
        @Body body: MarketingConsent
    ): Completable

    fun showUser(@Path("id") userId: String, @Query("include") include: String? = null, @QueryMap fields: Map<String, String> = emptyMap()): Single

    fun logout(@Path("userId") userId: String): Completable
}
```

Running ktlint in `--format` mode gives the following output:

```Kotlin
interface UserService {

    fun updateMarketingConsent(
        @Path("userId")
        userId: String,
        @Path("context")
        context: String,
        @Body body: MarketingConsentStructure
    ): Completable

    fun showUser(
        @Path("id")
        userId: String,
        @Query("include") include: String? = null, @QueryMap fields: Map<String, String> = emptyMap()
    ): Single

    fun logout(
        @Path("userId")
        userId: String
    ): Completable
}
```

Imo, the latter is way less readable. I went through [the official styleguide](https://kotlinlang.org/docs/reference/coding-conventions.html#annotation-formatting), but I think an annotation with arguments should be allowed for method params (even though not sure that it's applicable for all possible cases).

Also in the `showUser` function, the last two parameters should be placed on different lines (so the with-argument annotation followed by no-argument annotation). Cause otherwise the `parameter-list-wrapping` rules kicks in. This is most likely a bug in formatter, right?

rom4ek:
Resolve #556 and #608. 

As I mentioned in the issue in the [official styleguide](https://kotlinlang.org/docs/reference/coding-conventions.html#annotation-formatting) there's nothing about single-line annotations on method/constructor parameters. Also extended it a little, to allow the same thing in arguments' annotations (e.g. when you have a `@Retention(SOURCE) @Target(EXPRESSION)` for the annotation class)

### DIFF
--- a/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt
+++ b/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt
@@ -2,6 +2,8 @@ package com.pinterest.ktlint.ruleset.experimental
 
 import com.pinterest.ktlint.core.Rule
 import com.pinterest.ktlint.core.ast.ElementType.MODIFIER_LIST
+import com.pinterest.ktlint.core.ast.ElementType.VALUE_ARGUMENT_LIST
+import com.pinterest.ktlint.core.ast.ElementType.VALUE_PARAMETER_LIST
 import com.pinterest.ktlint.core.ast.children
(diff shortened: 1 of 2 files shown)


Tapchicoma:
Yes, I see it is a valid code style. Thank your for contribution :+1:

PR #646, opened 2019-11-15
Prevent invalid commands in CLI and fail the execution

### LINKED ISSUE #629: The CLI silently fails when attempting to use an invalid command
There is no response for invalid commands, nor any indication via the exit code:

```
$ ktlint -V
0.35.0

$ ktlint foo

$ echo $? 
0  
```

**Expected Behaviour:**

A typo or invalid command should generate an error message and return with a non-zero exit code.

rom4ek:
Fixes #629

### DIFF
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
@@ -52,6 +52,14 @@ fun main(args: Array<String>) {
         .addSubcommand(ApplyToIDEAGloballySubCommand.COMMAND_NAME, ApplyToIDEAGloballySubCommand())
         .addSubcommand(ApplyToIDEAProjectSubCommand.COMMAND_NAME, ApplyToIDEAProjectSubCommand())
     val parseResult = commandLine.parseArgs(*args)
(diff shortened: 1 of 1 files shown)


rom4ek:
Ah ok, that's actually a little bit trickier than I thought, cause the main function gets the file paths as a list of patterns in free form (could be simply `src`, `src/`, `src/**/*.kt[s]` or a file path)
https://github.com/pinterest/ktlint/blob/6864374fac6de26516277814359e32c070d04342/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt#L210

Therefore it's hard to distinguish what is a valid input and what's not. In theory we could check for `patterns.flieSequence().count() == 0`, and if it is, then throw an exception about the invalid command. Not sure how to proceed here.

Tapchicoma:
Linked issue is kind of invalid, as you already noticed - last arguments goes into file patterns.

I've tested locally and `ktlint` fails is I pass argument with a typo in it, for example `./ktlint -b` will print Error: Unmatched argument [-b] and then general help message.

rom4ek:
Yep, I'm aware of unmatched argument behaviour. Maybe as an improvement, there could be a message like "ktlint has not found any files to inspect" in case of empty `patterns`. But agree, there's no need to fix this thing so I'd rather close this and the issue as well

PR #628, opened 2019-10-21
Fix error on formatting annotation for primary constructor

Monchi:
Resolve #625 

Fix error on formatting annotation for primary constructor.

#### Current
Formatting following code causes the error.
```kotlin
class Foo @Inject internal constructor()
```
```stacktrace
com.pinterest.ktlint.core.RuleExecutionException: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at com.pinterest.ktlint.core.KtLint$format$1.invoke(KtLint.kt:384)
	at com.pinterest.ktlint.core.KtLint$format$1.invoke(KtLint.kt:41)
	at com.pinterest.ktlint.core.KtLint$visitor$2$2.invoke(KtLint.kt:279)
	at com.pinterest.ktlint.core.KtLint$visitor$2$2.invoke(KtLint.kt:41)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:536)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:537)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:537)
	at com.pinterest.ktlint.core.KtLint.access$visit(KtLint.kt:41)
	at com.pinterest.ktlint.core.KtLint$visitor$2.invoke(KtLint.kt:278)
	at com.pinterest.ktlint.core.KtLint$visitor$2.invoke(KtLint.kt:41)
	at com.pinterest.ktlint.core.KtLint.format(KtLint.kt:367)
	at com.pinterest.ktlint.internal.FileUtilsKt.formatFile(FileUtils.kt:79)
	at com.pinterest.ktlint.KtlintCommandLine$run$1.invoke(Main.kt:240)
	at com.pinterest.ktlint.KtlintCommandLine$run$4$1.call(Main.kt:306)
	at com.pinterest.ktlint.KtlintCommandLine$run$4$1.call(Main.kt:115)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.substring(String.java:1929)
[...]

PR #651, opened 2019-11-21
Fix invalid github link to post an issue

GotoFinal:


GotoFinal:
@Tapchicoma ah, can I ask when there will be new release? as there is a bug that is showing up in every project I want to use ktlint with, and I see that it is already fixed in #628

Tapchicoma:
@GotoFinal can't say exact day, but probably next week

PR #702, opened 2020-02-20
Add line and column to ParseException message

sharedprophet:
For build environments where normal exception printing is all we can see, this info is getting lost...


sharedprophet:
Any idea when this will be released?

Tapchicoma:
@sharedprophet we are aiming to make new release this week

PR #707, opened 2020-02-27
Add baseline functionality to ktlint

joshafeinberg:
Adds in a new flag, `baseline`, that allows you to specify a baseline to check again when running your report.

The functionality is similar to [Android lint's baseline](https://developer.android.com/studio/write/lint#snapshot). It creates a snapshot file of the current errors in the project on the first run using the checkstyle reporter. On all following runs it checks against that baseline file and only reports errors that do not exist inside that file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,9 @@ $ ktlint --reporter=plain?group_by_file
 # print style violations as usual + create report in checkstyle format 
 $ ktlint --reporter=plain --reporter=checkstyle,output=ktlint-report-in-checkstyle-format.xml
 
+# check against a baseline file
+$ ktlint --baseline=ktlint-baseline.xml
+
 # install git hook to automatically check files for style violations on commit
 # Run "ktlint installGitPrePushHook" if you wish to run ktlint on push instead
 $ ktlint installGitPreCommitHook
@@ -288,6 +291,8 @@ task ktlint(type: JavaExec, group: "verification") {
     args "src/**/*.kt"
     // to generate report in checkstyle format prepend following args:
     // "--reporter=plain", "--reporter=checkstyle,output=${buildDir}/ktlint.xml"
+    // to add a baseline to check against prepend following args:
+    // "--baseline=ktlint-baseline.xml"
     // see https://github.com/pinterest/ktlint#usage for more
 }
 check.dependsOn ktlint
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ task ktlint(type: JavaExec, group: LifecycleBasePlugin.VERIFICATION_GROUP) {
   description = "Check Kotlin code style."
   classpath = configurations.ktlint
   main = 'com.pinterest.ktlint.Main'
-  args '*/src/**/*.kt'
+  args '*/src/**/*.kt', '--baseline=ktlint-baseline.xml'
 }
 
 allprojects {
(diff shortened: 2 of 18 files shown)


shashachu:
I wonder if it would be cleaner to make your own reporter rather than rely on checkstyle. Seems like a non-obvious dependency that changing checkstyle could potentially break this functionality.

Also - it looks like the `--baseline` param both generates or loads the baseline file? Maybe less confusing to break them into separate commands.

shashachu:
We'd also want to build a set of tests.

joshfeinberg-ah:
> I wonder if it would be cleaner to make your own reporter rather than rely on checkstyle. Seems like a non-obvious dependency that changing checkstyle could potentially break this functionality.

Fair point, just wasn't sure if that would be too large of a change. I'll add in a new reporter tonight.

> Also - it looks like the `--baseline` param both generates or loads the baseline file? Maybe less confusing to break them into separate commands.

That is how it works with Android lint so I copied that model. If it exists then it loads otherwise it generates so that it can be run against the next time. I'm not sure if it makes sense to have two flags when you wouldn't know if you had a baseline file or not though happy to add in if you prefer

> We'd also want to build a set of tests.

I'll add some tonight as well where possible

joshafeinberg:
> I wonder if it would be cleaner to make your own reporter rather than rely on checkstyle. Seems like a non-obvious dependency that changing checkstyle could potentially break this functionality.

Fair point, just wasn't sure if that would be too large of a change. I'll add in a new reporter tonight.

> Also - it looks like the `--baseline` param both generates or loads the baseline file? Maybe less confusing to break them into separate commands.

That is how it works with Android lint so I copied that model. If it exists then it loads otherwise it generates so that it can be run against the next time. I'm not sure if it makes sense to have two flags when you wouldn't know if you had a baseline file or not though happy to add in if you prefer

> We'd also want to build a set of tests.

I'll add some tonight as well where possible

joshafeinberg:
So now there is an interesting issue - I can't have test resources that fail because they don't follow `ktlint` but I need them not to so that the tests pass. Any thoughts @shashachu?

jeffmcnd:
@shashachu how does this PR look now? This is a much needed feature!

LionZXY:
Any update? Need this feature

satorikomeiji:
Currently this prevents our team from adopting ktlint, the feature is critical for large projects with huge poorly formatted legacy codebase

emrisb:
I implemented ktlint in old multi module project with baseline, but it generates a baseline file for each module. Is there any way to generate one baseline file for all modules?

joshfeinberg-ah:
> I implemented ktlint in old multi module project with baseline, but it generates a baseline file for each module. Is there any way to generate one baseline file for all modules?

Not in how this is intended to work. Feel free to file a feature request

eboudrant:
Great feature @joshfeinberg. I have a question, if you run ktlint with the baseline and the `-F` option, is it supposed to fix the code noted in the baseline? I would expect it to not fix code in baseline but it seems it fix all code. That would be great for having a command to format the new code only.

joshafeinberg:
> Great feature @joshfeinberg. I have a question, if you run ktlint with the baseline and the `-F` option, is it supposed to fix the code noted in the baseline? I would expect it to not fix code in baseline but it seems it fix all code. That would be great for having a command to format the new code only.

The baseline functionality does not affect the formatter option at all. You can configure ktlint to just run on files currently in your git stage and that should solve this issue. Feel free to file a feature request on ktlint if you have a reason to specifically want a feature to not change stuff in the baseline

eboudrant:
I see. We use git stage. The reason was to prevent reformatting an entire file when you only touch few lines.

PR #714, opened 2020-03-12
Add file annotations rule

### LINKED ISSUE #689: Format file annotations
This should be flagged:
```kotlin
@file:JvmName package foo.bar
```
and reformatted to
```kotlin
@file:JvmName

package foo.bar
```
I tried to create a PR myself to fix this but I can't find a neat way to handle all the cases.

The problem is that the file annotations are treated as normal annotations but they are different. They have its own rules in the [coding conventions](https://kotlinlang.org/docs/reference/coding-conventions.html#file-annotations).

t-kameyama:
Fixes #689

### DIFF
--- a/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt
+++ b/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt
@@ -37,7 +37,8 @@ class AnnotationRule : Rule("annotation") {
             "Multiple annotations should not be placed on the same line as the annotated construct"
         const val annotationsWithParametersAreNotOnSeparateLinesErrorMessage =
             "Annotations with parameters should all be placed on separate lines prior to the annotated construct"
-        const val fileAnnotationsShouldBeSeparated = "File annotations should be separated from packages with a blank line"
+        const val fileAnnotationsShouldBeSeparated =
+            "File annotations should be separated from file contents with a blank line"
     }
 
     override fun visit(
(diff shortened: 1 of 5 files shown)


ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt
```diff
@@ -30,6 +37,7 @@ class AnnotationRule : Rule("annotation") {
             "Multiple annotations should not be placed on the same line as the annotated construct"
         const val annotationsWithParametersAreNotOnSeparateLinesErrorMessage =
             "Annotations with parameters should all be placed on separate lines prior to the annotated construct"
+        const val fileAnnotationsShouldBeSeparated = "File annotations should be separated from packages with a blank line"
```
romtsn, line 40: Following up from #747, I think the wording should be a bit different, cause there's not necessarily a package coming after a file annotation (e.g. in `.kts` files). So maybe change it to `from file contents` or something like that.

ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt
```diff
@@ -106,6 +115,29 @@ class AnnotationRule : Rule("annotation") {
...
+        if (node.elementType == FILE_ANNOTATION_LIST) {
+            val lineNumber = node.lineNumber()
+            val next = node.nextSibling {
+                !it.isWhiteSpace() && it.textLength > 0 && !(it.isPartOfComment() && it.lineNumber() == lineNumber)
+            }
+            val nextLineNumber = next?.lineNumber()
+            if (lineNumber != null && nextLineNumber != null) {
+                val diff = nextLineNumber - lineNumber
+                if (diff < 2) {
+                    emit(0, fileAnnotationsShouldBeSeparated, true)
```
romtsn, line 128: Hm, what if there's a license header above the file annotation, then I guess `0` offset is not correct anymore? According to [codestyle](https://kotlinlang.org/docs/reference/coding-conventions.html#file-annotations), there can be something on top of the annotation.

PR #725, opened 2020-04-14
KtLint: Make functions needed to calculate line / column public

sschuberth:
In order to avoid duplicating code, make these functions public for use
by tools that use KtLint programmatically and need to convert an offset
to line / column, like detekt.


shashachu:
@sschuberth Apologies for the delay. Can you try rebasing and re-pushing your PR? looks like something went wrong with the checks.

sschuberth:
Done @shashachu.

PR #727, opened 2020-04-20
Fix autocorrect for single line if/else expressions

### LINKED ISSUE #174: 'If' should force curly braces
As stated in the [Android Kotlin Style Guide under **Formatting**](https://android.github.io/kotlin-guides/style#formatting), even if an `If` statement have only one line, it should be surrounded by a curly brace.

In my project, my code: 
```
if (foo)
    bar = "hello world"
```
was deemed acceptable by ktlint. Is there anything I can do short of writing my own rule? Thanks in advance.

csgulley:
This fixes the autocorrect functionality in MultiLineIfElseRule that was partially implemented but commented out. It should resolve #174.

### DIFF
--- a/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/MultiLineIfElseRule.kt
+++ b/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/MultiLineIfElseRule.kt
@@ -7,6 +7,7 @@ import com.pinterest.ktlint.core.ast.ElementType.LBRACE
 import com.pinterest.ktlint.core.ast.ElementType.RBRACE
 import com.pinterest.ktlint.core.ast.ElementType.THEN
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.PsiWhiteSpaceImpl
 import org.jetbrains.kotlin.psi.KtBlockExpression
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/MultiLineIfElseRule.kt
```diff
@@ -22,14 +27,31 @@ class MultiLineIfElseRule : Rule("multiline-if-else") {
...
+                    autocorrect(node)
+                }
             }
         }
     }
+
+    private fun autocorrect(node: ASTNode) {
+        node.treeParent.replaceChild(node.treePrev, PsiWhiteSpaceImpl(" "))
+        val previousChild = node.firstChildNode
+        KtBlockExpression(null).apply {
```
rom4ek, line 43: hm, does this support indentation? For instance, if I had a nested if blocks, like:
```
if (condition1) {
    if (condition2) return true
}
```
I believe the `addChild(previousChild)` will not indent the `return true` block properly.. as well as the following RBRACE

csgulley: I just pushed a change that should indent the code properly.

ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/MultiLineIfElseRule.kt
```diff
@@ -22,14 +28,33 @@ class MultiLineIfElseRule : Rule("multiline-if-else") {
...
-                }*/
+                if (autoCorrect) {
+                    autocorrect(node)
+                }
             }
         }
     }
+
+    private fun autocorrect(node: ASTNode) {
+        val bodyIndent = node.treePrev.text
```
rom4ek, line 42: I think this still won't cover all the use cases.. so if you have a `if (true) return 0` on the same line, than it should be
```
if (true) {
    return 0
}
```
but this will format it to
```
if (true) {
 return 0
}
```

I guess you better take a look at the [IndentationRule](https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/IndentationRule.kt) and try to replicate its behaviour in regards to if/else indentation.

Or, I'm wondering, if it's possible to just call `IndentationRule().format(KtBlockExpression.text)` ? Never seen that the rules can invoke other rules though.. perhaps, it would be more applicable to extract that logic from the IndentationRule to some common place, so other rules can reuse that for indentation when needed.. what are your thoughts @shashachu @Tapchicoma ?

csgulley: This rule allows single line if statements so it would not change `if (true) return 0`. One of the test cases verifies this. Are there other cases where the resulting indentation would be wrong?

rom4ek: True, sorry, I totally missed that 🤦 Can you maybe add one more test just to make sure (although it shouldn't change anything), with multiline condition within `if` block? Like
```
        if (i2 > 0 &&
            i3 < 0
        ) {
            return 2
        }
```

romtsn:
@csgulley could you rebase so the checks can pass again?

romtsn:
Sorry for coming up with edge-cases again, but could you also add a test with `else if` branch in place? Don't think it will change something either, but good to have that case tested. After that I think we're good to merge this one.

romtsn:
@csgulley I've updated the handling in case of `else if` block present, I hope it's fine for you. I will merge it, once figure out the CI issues. Thanks a lot for implementing this!

Issue #728, opened 2020-04-20
Formatter breaks class with trailing comment


scottdweber:
Given a class that is defined like this:

```kotlin
class CoolThing (
    val thing1: String,
    val thing2: String
) // this is a comment
{
    // ...
}
```

## Expected Behavior
When running the ktlint formatter, output should be:

```kotlin
class CoolThing (
    val thing1: String,
    val thing2: String
) {
    // this is a comment
    // ...
}
```

## Observed Behavior
Actual output (0.36.0) is a class that no longer compiles:

```kotlin
class CoolThing (
    val thing1: String,
    val thing2: String
) // this is a comment {
    // ...
}
```

## Your Environment
* Version of ktlint used: `0.36.0`
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): `org.jlleitschuh.gradle:ktlint-gradle:9.2.1`
* Version of Gradle used (if applicable): `5.6`
* Operating System and version: `macOS`

shashachu:
@scottdweber could you do me a favor and just run `lint` (not format) with `--verbose` to see which rule it is that's making that change? (I'm unsure how to pass that flag with the plugin, but I assume it's supported).

scottdweber:
Looks like it is `curly-spacing`

romtsn:
Fixed by #730

PR #730, opened 2020-04-25
Move EOL comment after left curly brace so code remains compilable

csgulley:
This is a potential fix for #728. When pulling a left curly brace up to previous line, move EOL comment after curly brace.

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundCurlyRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundCurlyRule.kt
@@ -1,6 +1,5 @@
 package com.pinterest.ktlint.ruleset.standard
 
-import com.pinterest.ktlint.core.KtLint
 import com.pinterest.ktlint.core.Rule
 import com.pinterest.ktlint.core.ast.ElementType.AT
 import com.pinterest.ktlint.core.ast.ElementType.CLASS_BODY
@@ -20,7 +19,6 @@ import com.pinterest.ktlint.core.ast.ElementType.RPAR
 import com.pinterest.ktlint.core.ast.ElementType.SAFE_ACCESS
 import com.pinterest.ktlint.core.ast.ElementType.SEMICOLON
 import com.pinterest.ktlint.core.ast.isPartOfString
-import com.pinterest.ktlint.core.ast.isRoot
 import com.pinterest.ktlint.core.ast.nextLeaf
 import com.pinterest.ktlint.core.ast.prevLeaf
 import com.pinterest.ktlint.core.ast.upsertWhitespaceAfterMe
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundCurlyRuleTest.kt
```diff
@@ -209,4 +209,71 @@ class SpacingAroundCurlyRuleTest {
...
+                """
+                class MyClass()// a comment
+                {
+                    val x = 0
+                }
+                """.trimIndent()
+            )
+        ).isEqualTo(
+            """
+                class MyClass() {
```
rom4ek, line 226: I'm wondering if it actually should be 
```
class MyClass() { // a comment
    ...
}
```

csgulley: Yes I think leaving the comment on the same line as the curly brace is preferable to moving it to next line. I just pushed the code.

shashachu:
@csgulley sorry for the delay - can you try rebasing an re-pushing your diff? the CI build didn't complete.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundCurlyRule.kt
```diff
@@ -67,6 +70,18 @@ class SpacingAroundCurlyRule : Rule("curly-spacing") {
                 ) {
                     emit(node.startOffset, "Unexpected newline before \"${node.text}\"", true)
                     if (autoCorrect) {
+                        val eolCommentExists = prevLeaf.prevLeaf()?.let {
+                            it is PsiComment && it.elementType == EOL_COMMENT
```
romtsn, line 74: Now when I'm thinking more about it - would `/* a comment */` also have `EOL_COMMENT` type actually? I think this should also be allowed.. sorry for not spotting it earlier

csgulley: I'm not sure that end-of-line and block comments should be handled the same here. The original bug, #728, addressed the fact that an end-of-line comment caused the output you get from running ktlint to no longer be compilable Kotlin code which I think most people would find undesirable regardless of how the code is styled. In the case of a block comment, the code generated from the current ktlint will still be compilable. If we are going to handle block comments, do we also need to handle various styles of multi-line comments that might be found? If you think we should handle block comments can you provide examples on the desired before/after code for each case? I'd be happy to make the changes but I'm not 100% clear on what the goal is.

romtsn: True, I think you're right, handling different styles of comments is probably an overkill here.. great, then it looks good 👍

romtsn:
Thank you!

PR #740, opened 2020-05-13
AnnotationRule: don't report if a comment is placed after the last annotation

### LINKED ISSUE #736: Comment after annotations are interpreted as the `annotated construct`.
## Expected Behavior
No lint-error should be given

## Observed Behavior
When placing a comment after the last annotation, I get the error

    Multiple annotations should not be placed on the same line as the annotated construct

## Steps to Reproduce
Create the following class

```
import org.junit.Ignore
import org.junit.Test
import org.junit.Assert.fail

class SomeTest {
  @Test
  @Ignore // foobar
  fun testThisShouldBeIgnored() {
    fail("This should not have ran")
  }
}
```
Run ktLint, it will give the message `Multiple annotations should not be placed on the same line as the annotated construct`. When you remove `// foobar` or place it on its own line, no error is shown.

## Your Environment

* Version of ktlint used: `0.36.0`
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): 
`org.jlleitschuh.gradle.ktlint`, version `9.2.1`
* Version of Gradle used (if applicable): `Gradle 6.2.1`
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04 LTS
Release:        20.04
Codename:       focal
$ uname -r
5.4.0-29-generic
```
* Link to your project (if it's a public repository): private code

t-kameyama:
Fixes #736

### DIFF
--- a/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt
+++ b/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/AnnotationRule.kt
@@ -75,7 +75,7 @@ class AnnotationRule : Rule("annotation") {
         }
 
         val multipleAnnotationsOnSameLineAsAnnotatedConstruct =
-            annotations.size > 1 && !whiteSpaces.last().textContains('\n')
+            annotations.size > 1 && !whiteSpaces.last().textContains('\n') && doesNotEndWithAComment(whiteSpaces)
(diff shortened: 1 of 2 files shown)


shashachu:
@t-kameyama thanks for the PR. Can you confirm that 
```
@Annotation1 // comment 1
@Annotation2
class MyClass {
}
```

passes with no errors as well? I believe it should based on the bug report but I'd like to confirm. Apologies if this is already covered in test cases.

t-kameyama:
Added a test.

shashachu:
thank you!

PR #749, opened 2020-05-29
Introduce import layout configuration for ImportOrderingRule

### LINKED ISSUE #527: Alphabetical import ordering enabled by default is a problem in Android Studio/IntelliJ
In our project `ktlint 0.34.0` now reports a lot of incorrect import ordering, although we use a default import layou of IDE. All our developers have "Optimize imports" enabled by default, but this still creates an order like the following (which seems to be default one of Kotlin IDE plugin):

```
import ru.example.a
import ru.example.b
import ru.example.c
import javax.inject.Inject
```

Notice that the last one comes in non alphabetical order. 
The unfortunate thing is that this cannot be changed.

Import order editing is there only for Java CodeStyle, kotlin plugin seems not to provide it. Correct me if I'm wrong.

If you think that this is still something that should be default, then please close.

But currently I suspect that many Android Studio/IntelliJ users will be disabling this rule, otherwise they'll have an endless conflict between IDE autoformatter/ktlint.

romtsn:
Fix #527


.editorconfig
```diff
@@ -13,6 +13,9 @@ insert_final_newline = true
 [*.{java,kt,kts,scala,rs,xml,kt.spec,kts.spec}]
 indent_size = 4
 
+[*.{kt,kts}]
+kotlin_imports_layout=ascii
```
romtsn, line 17: I'm not sure if we should use "idea" for ktlint itself.. If yes, then I'd prefer to do it in a separate PR afterwards, but can also do it here

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
```diff
@@ -1,53 +1,171 @@
...
+    }
+
     override fun visit(
         node: ASTNode,
         autoCorrect: Boolean,
         emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit
     ) {
+        if (node.isRoot()) {
+            val editorConfig = node.getUserData(KtLint.EDITOR_CONFIG_USER_DATA_KEY)!!
+            importsLayout = editorConfig.get("kotlin_imports_layout")
```
romtsn, line 76: I've decided to not put `ij_` in front, because it would look weird a bit if it was `ij_kotlin_imports_layout=idea`. But open for suggestions

.editorconfig
shashachu: I personally use IDEA but I rarely use the IDE's code formatting so I don't have a strong opinion.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
```diff
@@ -1,53 +1,171 @@
...
+                ?: "idea" // default to idea in case there's no such entry in .editorconfig
```
gabrielittner, line 77: For Android projects this default violates the style guide https://developer.android.com/kotlin/style-guide#import_statements

romtsn: It violates the style guide, but it complies with how IDEA/Android Studio actually sorts your imports, so I'd say it's a good middle-ground. You can always configure it to "ascii" if you absolutely want to comply with Android style guide.

Also I don't really see a point in defining a style guide, such as you cannot set up Android Studio to comply with it, because there's literally no such capability for Kotlin yet.

.editorconfig
Tapchicoma: I would say it is fine. And this property could be in the future not really IDEA specific.

README.md
```diff
@@ -82,6 +82,17 @@ max_line_length=off
 # Note that rules in any ruleset other than the standard ruleset will need to be prefixed 
 # by the ruleset identifier.
 disabled_rules=no-wildcard-imports,experimental:annotation,my-custom-ruleset:my-custom-rule
+
+# Defines the imports layout. There are predefined layouts like "ascii" or "idea", as well as a custom layout.
```
Tapchicoma, line 86: Nice documentation :+1:. Probably should mention that this will be available starting Kotlin 1.4-M3 release.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
Tapchicoma: Hm, I think anyway it should have `ij_` prefix in the `.editorconfig` so IDEA will pickup for codestyle? Correct me if I am wrong.

If yet, I would prefer using it here as well. Could rethink this approach in the future.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
Tapchicoma: Probably you could reuse `ANDROID_USER_DATA_KEY` to select default config `ascii` for android projects.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/internal/importordering/ImportLayoutParser.kt
```diff
@@ -0,0 +1,44 @@
...
+            var hasAlias = false
+            var withSubpackages = false
+            if (import.startsWith(ALIAS_CHAR)) {
+                import = import.substring(1).trim()
+                hasAlias = true
+            }
+            if (import.endsWith(WILDCARD_CHAR)) {
+                withSubpackages = true
+            }
+            imports += if (import == WILDCARD_CHAR) {
```
Tapchicoma, line 36: Nit: probably you could use transformations like `fold` on `importsList` to transform it to end result instead of using `forEach`

romtsn: True, it was a bit different in the initial implementation, but now I think it's doable via transformation operators.. probabl `map` fits best here, will change that 👍

README.md
romtsn: But I mean it's available for ktlint already now, so you can specify the custom layout for ktlint, there's just no such capability in Kotlin IDEA plugin yet.. or do you want to also mention about the kotlin plugin just in case?

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
romtsn: I could certainly do that, but I think this will result in 95% of android devs to set it back to `idea`, until Kotlin 1.4 is officially out (I guess only a small percentage is using milestone versions), otherwise it will lead to the same conflicts and frustration as it is now.. I'd rather change it after 1.4 is officially there - what do you guys think?

romtsn: Implemented default to `ascii` in case of android flag enabled

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
romtsn: As discussed over slack, added support for both names

README.md
Tapchicoma: You are right, we could mention it later when it will be publicly available.

PR #767, opened 2020-06-05
IndentationRule: Allow free-form text inside raw strings

### LINKED ISSUE #682: Formatter never satisfied with raw string indentation
### Source code

```kt
val y = 5
val x =
  """
    $y
  """.trimIndent()
```

### Expected behavior

`ktlint --experimental --format` is a no-op after at most one invocation.

### Actual behavior

ktlint always adds 6 additional space characters in front of `$y`, i.e. running ktlint `n` times will prepend `6*n` spaces to `$y`.

romtsn:
This PR is aiming to allow any indentation within raw strings for now, will come up with something better later (perhaps new indentation rules that could handle `trimIndent`, `trimMargin` and plain raw strings differently)

Fix #682


shashachu:
lgtm. seems like the whole point of raw strings is that you preserve the contents exactly.

Issue #754, opened 2020-05-31
Publish single-jar version of ktlint to maven central?


ivoanjo:
Hello there! First of all, thank you very much for the awesome work on Ktlint. It's been an invaluable tool to get rid of style nitpicks during code reviews for my team! :bow: :bow: :bow: 

Currently, ktlint is published as both:
* A single file version, available [in the releases tab](https://github.com/pinterest/ktlint/releases)
* A regular jar with dependencies, that gets pushed to maven central

I would like to request that a third option gets added: publishing the jar that the single file version uses to maven central.

Why? The single-file version massively simplifies dependency management, but downloading the latest version from github is always a bit awkward/different from how we deal with other dependencies (for instance -- it does not have checksums, as requested in #695).

Having the single-jar version in maven central would simplify a lot the integration/updates of ktlint we have on our team.

Thanks, and have a great day!

romtsn:
I think it's already published, if you check [here](https://repo1.maven.org/maven2/com/pinterest/ktlint/0.36.0/) the `all` version is exactly the one ktlint executable is using

ivoanjo:
Ah, interesting! I did not find it in my searches, but that should be enough for my setup, thank you :pray:

PR #774, opened 2020-06-09
Add checksum generation task for executable

### LINKED ISSUE #695: Provide checksum file along with release
Would it be possible to provide checksums of the `ktlint` executable along with each release?

edit: My use case is that I want to download the executable and use the checksum to verify it was downloaded successfully

romtsn:
Close #695. Uses a [plugin](https://github.com/gradle/gradle-checksum) from Gradle itself for generating checksums.


shashachu:
Thanks! I did see your comment on https://github.com/pinterest/ktlint/issues/754 and it sounds like that particular question was resolved, but although we are publishing the fat jar, we aren't publishing the executable jar to maven, which is what the MD5 is going to be generated against. Am I misunderstanding?

PR #779, opened 2020-06-11
Fix import ordering distinction

### LINKED ISSUE #766: Import ordering rule removing used imports
With the fix for #527 released in 0.37.0, looked to turn the import ordering rule on for our projects with idea format, and found the following:

## Expected Behavior
With import ordering rule turned on (to idea format) via `.editorconfig`, running `ktlint -F` version 0.37.0 against https://github.com/trib3/leakycauldron/blob/master/json/src/test/kotlin/com/trib3/json/ObjectMapperTest.kt should not reformat the file.

## Observed Behavior // Steps to reproduce
With import ordering rule turned on (to idea format) via `.editorconfig`, running `ktlint -F` version 0.37.0 against https://github.com/trib3/leakycauldron/blob/master/json/src/test/kotlin/com/trib3/json/ObjectMapperTest.kt results in `assertk` imports other than `assertk.all` being removed, which then makes the file fail to compile due to the missing imports.

With import ordering rule turned on via `.editorconfig`, running `ktlint -F` version 0.36.0 (no idea format exists in this version) against https://github.com/trib3/leakycauldron/blob/master/json/src/test/kotlin/com/trib3/json/ObjectMapperTest.kt results in the java/kotlin imports being moved up from the end into lexicographic order (as expected)

## Your Environment
* Version of ktlint used: 0.37.0 / 0.36.0
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task):  Use maven plugin in general, but command line for the minimal test cases above
* Version of Gradle used (if applicable): n/a
* Operating System and version: ubuntu linux 20.04
* Link to your project (if it's a public repository): https://github.com/trib3/leakycauldron

romtsn:
Another 🤦 
Finally fix #766

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Experimental SpacingAroundAngleBracketsRule ([#769](https://github.com/pinterest/ktlint/pull/769))
 - Checksum generation for executable Jar ([#695](https://github.com/pinterest/ktlint/issues/695))
 
+### Fixed
+- Imports with aliases no longer removed ([#766](https://github.com/pinterest/ktlint/issues/766))
+
 ## [0.37.1] - 2020-06-08
 
 Minor release to fix some bugs in the 0.37.0 release.
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


Tapchicoma:
@romtsn could you also update changelog?

PR #780, opened 2020-06-11
Fix githubRelease is not generating proper release information.

Tapchicoma:
Set `dryRun` to `true` and run `githubRelease` task, then check output of the task.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,16 +76,35 @@ String getGithubToken() {
   }
 }
 
+// Explicitly adding dependency on "shadowJarExecutable" as Gradle does not it set via "releaseAssets" property
+tasks.named("githubRelease") {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -76,16 +76,35 @@ String getGithubToken() {
...
+  dependsOn { project(":ktlint").tasks.named("shadowJarExecutable") }
+}
+
 githubRelease {
   token getGithubToken()
   owner "pinterest"
   repo "ktlint"
+  tagName project.properties['VERSION_NAME']
+  releaseName project.properties['VERSION_NAME']
+  releaseAssets project.files({
```
romtsn, line 90: there's also an `.asc` file and the sources zip - are they also being added?

Tapchicoma: if it will be in `shadowJarExecutable` folder - should pick up. If not - easy to add manually here

PR #783, opened 2020-06-14
Update to Gradle 6.5

Tapchicoma:
Additionally enabled new Gradle dependency verification mechanism: https://docs.gradle.org/current/userguide/dependency_verification.html

And some Gradle properties to speed-up local builds.


gradle.properties
```diff
@@ -13,3 +13,8 @@ POM_LICENCE_DIST=repo
 
 POM_DEVELOPER_ID=pinterest
 POM_DEVELOPER_NAME=Pinterest, Inc.
+
+# Gradle properties
```
romtsn, line 17: maybe also configureondemand?

Tapchicoma: Usually it is discouraged to enable it and most probably will gone in next major release as it will be replaced by instant-execution: https://docs.gradle.org/nightly/userguide/configuration_cache.html

shashachu:
nice! Thank you

Issue #680, opened 2020-01-02
Contradictory indentation errors with multiline type parameter list


bethcutler:
If a code snippet contains a multi-line type parameter list, ktlint may produce contradictory errors with the experimental indentation rule.

E.g. for this code snippet
```kotlin
object TestCase {
  inline fun <
    T1,
    T2,
    T3> create(
      t1: T1,
      t2: T2,
      t3: T3
  ) {
    // do things
  }
}
```

... ktlint produces this errors for each function parameter line (t1, t2, t3):
> Unexpected indentation (6) (should be 4)

If the indentation on the function parameter lines is "corrected" like so
```kotlin
object TestCase {
  inline fun <
    T1,
    T2,
    T3> create(
    t1: T1,
    t2: T2,
    t3: T3
  ) {
    // do things
  }
}
```

... ktlint is still not satisfied, now producing this error on each of those lines:
> Unexpected indentation (expected 6, actual 4)

Tapchicoma:
Probably could be covered by https://github.com/pinterest/ktlint/pull/587

AdamMTGreenberg:
@Tapchicoma Added PR: https://github.com/pinterest/ktlint/pull/804 to enforce this behavior. However, it is unclear if we want to make this an enforced rule:

```
object TestCase {
  inline fun <
    T1,
    T2,
    T3> create(
    t1: T1,
    t2: T2,
    t3: T3
  ) {
    // do things
  }
}
```
to
```
object TestCase {
  inline fun <
    T1,
    T2,
    T3> create(
      t1: T1,
      t2: T2,
      t3: T3
  ) {
    // do things
  }
}
```
since the default spacing seems to leave the indentation flat. 

If we want to leave it as is, then I think we can close out this issue.

PR #786, opened 2020-06-15
Fixes #620 by adding arguments to ParameterListWrappingRule.

### LINKED ISSUE #620: Is it possible to enforce parameters are on their own line?
I found myself in a scenario where I had multiple parameter on one line, with others on a different line, like this:

```kotlin
fun test(one: String, two: String, three: String, four: String) {
}

fun callTest() {
    test(
        "One",
        "Two", "Three", "Four"
    )
}
```

KtLint actually passes this, because there is a new line after the opening `(` and one before the closing `)`, but I still think it's weird to have one parameter on one line and three on the other. 

I'd really prefer it be an all or nothing thing:

```kotlin
// This
test("One", "Two", "Three", "Four")
// Or
test(
    "One",
    "Two",
    "Three",
    "Four"
)
```

Any thoughts on a way to enforce this, if it's possible?

AdamMc331:
Reference ticket #620  for details. This now considers anything that is `VALUE_ARGUMENT` in addition to `VALUE_PARAMETER` for the `ParameterListWrappingRule`. This means function calls will get the same treatment as class definitions for their arguments being on separate lines (if necessary). 

## Example

Before, ktlint considered this valid:

```
fun callTest() {
    test(
        "One",
        "Two", "Three", "Four"
    )
}
```

This is not clear, though, as multiple arguments on one line is confusing, if they're not _all_ on one line. Now it will auto format them to be on separate lines:

```
fun callTest() {
    test(
        "One",
        "Two", 
        "Three", 
        "Four"
    )
}
```

## One Concern

All of the tests pass, but the annotation one is a little confusing as for me it appears it wanted to format the annotation like this:

```
class A {
    fun f(
        // ...
        @Annotation(
            [
            "v1",
            "v2"
        ]
        )
        // ...
}
```

Some back and forth with @romtsn on the ticket right now though to see if that's an issue before we merge this.


romtsn:
hey @AdamMc331, are you planning to finish this up or should we take over?

AdamMc331:
Hey @romtsn Sorry for falling off on this again. I still have an issue formatting this if there's a argument list inside an annotation. I will clean up my git history and write a test to demonstrate the bug, but I'd definitely appreciate some help looking into it I'm at a loss.

romtsn:
I made it work, but there's another corner case, which we probably should consider and I'm looking into it now:
```Kotlin
class NoVarRuleTest : Spek({
    // this should not fail
})
```

AdamMc331:
:eyes: Exciting! Thanks for your help pushing this across the finish line.

romtsn:
I am a bit hesitant to release this in the standard ruleset, as although the change is not significant code-wise, it kind of has a big impact even on ktlint's codebase.. maybe we should move it to the experimental ruleset for the next release? Thoughts @Tapchicoma @shashachu @AdamMc331? Otherwise, it's ready to be merged.

AdamMc331:
I'm open to the experimental ruleset but that could also be the imposter syndrome talking. 😂 This is my first contribution to a ktlint ruleset and so I'm definitely not as familiar as the others tagged but I will support their opinion either way. 

If we go with the experimental ruleset, what is the process like for getting it moved out of experimental in the future?

shashachu:
Although it may aggravate some people, I don't think it's a big deal to leave this in the standard ruleset, as its functionality is more consistent with the IndentRule, which is quite aggressive about adding newlines.

shashachu:
@AdamMc331 does this PR also address https://github.com/pinterest/ktlint/issues/680?

romtsn:
@shashachu nope, that problem is specific to indentation rule

romtsn:
Thanks Adam!

AdamMc331:
Thanks everyone! :)

adamu:
I’m seeing lots of results for `ParameterListWrapping` after upgrading to detekt 1.12.0 that look either questionable or wrong. Could it be related to this PR? A couple of examples:

Questionable. Fails with `Argument should be on a separate line (unless all arguments can fit a single line)` and `Missing newline before ")"` .
```kotlin
foo(bar.apply {
    // stuff
})
```

Wrong? Fails `function(arg1, arg2, arg3)` with `Argument should be on a separate line (unless all arguments can fit a single line)`.
```kotlin
json(
    """
    {
        "array": [
            ${function(arg1, arg2, arg3)}
        ]
    }
    """.trimIndent()
)
```

romtsn:
The latter one looks pretty much like https://github.com/pinterest/ktlint/issues/859, the former one though should not be reported (at least the arguments part). Could you create a separate issue for the former one? Thanks

adamu:
> @adamu The latter one looks pretty much like #859, the former one though should not be reported (at least the arguments part). Could you create a separate issue for the former one? Thanks

Done! #861

PR #789, opened 2020-06-17
Fixed HtmlReporterTest failing on windows

petertrr:
Command `gradlew.bat test` fails on Windows with HtmlReporterTest.

Changed `\n` to `System.lineSeparator()` in HtmlReporter. Other newlines were generated by `println()`, and in the test expected data has separator explicitly changed to platform-specific.

Now `gradlew.bat test` succeeds on Windows.


romtsn:
Thanks for the fix, we are going to switch to Github Actions in the course of next weeks/months to also run tests on windows and prevent such pitfalls.

PR #798, opened 2020-06-26
Add spacing rule for UnaryExpression

boyan01:
Add a rule to check spacing in `KtUnaryExpression`.

Example:
```
i++ // valid
i ++ // invalid
```


boyan01:
I just noticed that already a pull request for unary op has been merged ： https://github.com/pinterest/ktlint/pull/428

but I can not find any related files in master branch.

Tapchicoma:
It was reverted in this commit: a08dfd1

ktlint-ruleset-standard/src/test/resources/spec/unary-op-spacing/format.kt.spec
```diff
@@ -0,0 +1,17 @@
+fun main() {
```
Tapchicoma, line 1: Could you also add a tests for `val g = 3 + i` or `val k = i - 3` or another similar "normal" math cases?

boyan01: of course. "math case" added in this commit

https://github.com/pinterest/ktlint/pull/798/commits/f719fb9e84a22d12570c920f814a88fdef59182c

boyan01:
Suddenly thought of a case.
```kotlin
var i = - /*abc*/ 1  // it this should be valid?
i = -/*abc*/1 // or it this should be valid
```
I try to format this in IDEA, neither line of code is formatted..

Tapchicoma:
I wonder who writes such code with inlined comments :sweat_smile: 

IMHO, second case should be proper one.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingInUnaryExpressionRule.kt
```diff
@@ -0,0 +1,32 @@
...
+        if (node.elementType == ElementType.PREFIX_EXPRESSION ||
+            node.elementType == ElementType.POSTFIX_EXPRESSION
+        ) {
+            val children = node.children().toList()
+
+            // ignore: var a = + /* comment */ 1
+            if (children.any { it.isPartOfComment() }) return
+
+            val whiteSpace = children.firstOrNull { it.isWhiteSpace() } ?: return
+            if (autoCorrect) {
```
romtsn, line 26: I think, I would put this block after `emit`, because otherwise in case of autoCorrect, the node does not contain the whitespace already and the message is gonna be misleading

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingInUnaryExpressionRule.kt
```diff
@@ -0,0 +1,32 @@
...
+                node.removeChild(whiteSpace)
+            }
+            emit(whiteSpace.startOffset, "Unexpected spacing in ${node.text.replace("\n", "\\n")}", true)
```
romtsn, line 29: Why do we need the `replace` part? windows? I think if we do it, this should be done on a different level than per rule

boyan01: `replace ` intent to make the output more tidy. 
for example:
```kotlin
++
i
```
`node.text` will contains a break line, so use replace to remove this line break.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingInUnaryExpressionRule.kt
boyan01: oops, nice catch! will fix it

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingInUnaryExpressionRule.kt
romtsn: gotcha, sounds a bit specific though (never seen someone writing unary operators this way), but fine 👍

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/StandardRuleSetProvider.kt
```diff
@@ -36,6 +36,7 @@ class StandardRuleSetProvider : RuleSetProvider {
         SpacingAroundOperatorsRule(),
         SpacingAroundParensRule(),
         SpacingAroundRangeOperatorRule(),
+        SpacingInUnaryExpressionRule(),
```
romtsn, line 39: Can you move this to the experimental ruleset, so we indicate that the rule is incubating? We can't just put new rules to the standard ruleset unfortunately, as they need some battle-testing beforehand.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingInUnaryExpressionRule.kt
```diff
@@ -0,0 +1,33 @@
+package com.pinterest.ktlint.ruleset.standard
+
+import com.pinterest.ktlint.core.Rule
+import com.pinterest.ktlint.core.ast.ElementType
+import com.pinterest.ktlint.core.ast.children
+import com.pinterest.ktlint.core.ast.isPartOfComment
+import com.pinterest.ktlint.core.ast.isWhiteSpace
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+
+class SpacingInUnaryExpressionRule : Rule("unary-op-spacing") {
```
romtsn, line 10: Same as in the previous PR for this rule - could you add a short comment summarizing what the rule does and, perhaps, some link to either kotlin or google styleguide if there's a mention of this.

romtsn:
Thank you!

PR #804, opened 2020-06-29
Unit test to enforce issue #680

AdamMTGreenberg:
Adding a test to enforce the rule so that the spacing of a subset of multi-line type parameter list is not flattened when an indentation seems appropriate #680


romtsn:
Thanks for the PR, but the original issue was about the indentation rule, so I believe the fix should go there

PR #806, opened 2020-06-29
Extending `visit` function's argument list from `Rule.kt` with parameters

akuleshov7:
### What's done:
1. Extended visit function with new arguments. It becomes very useful in case you will need to match or check any parameter in any rule.
For example - in case you will need to check a filename in a rule - you don't need to pass it through your code and change every visitor. Everything will be already passed through and you will need simply to add needed parameter to `KtLint.Params` data class.
2. Changed all rules and tests where visit is overridden

### DIFF
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
@@ -164,7 +164,7 @@ object KtLint {
             // (not just because not all nodes are leaves but because rules are free to emit (and fix!) errors at any position)
             if (!isSuppressed(node.startOffset, fqRuleId, node === rootNode)) {
                 try {
-                    rule.visit(node, false) { offset, errorMessage, canBeAutoCorrected ->
+                    rule.visit(node, false, params) { offset, errorMessage, canBeAutoCorrected ->
                         // https://github.com/shyiko/ktlint/issues/158#issuecomment-462728189
                         if (node.startOffset != offset && isSuppressed(offset, fqRuleId, node === rootNode)) {
                             return@visit
@@ -348,7 +348,7 @@ object KtLint {
                 // (not just because not all nodes are leaves but because rules are free to emit (and fix!) errors at any position)
                 if (!isSuppressed(node.startOffset, fqRuleId, node === rootNode)) {
                     try {
-                        rule.visit(node, true) { _, _, canBeAutoCorrected ->
+                        rule.visit(node, true, params) { _, _, canBeAutoCorrected ->
                             tripped = true
                             if (canBeAutoCorrected) {
                                 mutated = true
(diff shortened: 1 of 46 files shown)


akuleshov7:
This is very useful for cases when you need to check file naming or any other stuff that can be added to Ktlint params.
We use this thing for our rule sets.

akuleshov7:
This will not break anything in ktlint, but will give a chance for developers of rules to use your code instead of customized forks. 
Simply passing arguments to visitors and open 1 class for inheritance.

romtsn:
This will break downstream dependencies though (like [detekt](https://github.com/detekt/detekt/blob/master/detekt-formatting/src/main/kotlin/io/gitlab/arturbosch/detekt/formatting/FormattingRule.kt#L61)) in that the public API has changed. But I guess it's fine. cc @arturbosch

akuleshov7:
@romtsn yeah, I understand this, but the main problem is now for usage of the framework is that we are not able to extend your library to write some rules. For example - to check everything related to file naming. Or do other checks where parameters being passed to a visitor are needed. 

I am pretty sure that one day some one will need parameters being passed inside the visitors and at that time everything will be so mature that you won't be able to change, unfortunately. So better to do that now or never :)

akuleshov7:
@romtsn @arturbosch so what do you think about breaking your API a little bit? :)

Tapchicoma:
Your idea is interesting. While ktlint is not yet reached 1.x version, we try to introduce api changes with some deprecation strategy and keep it for few versions.

If you really want this, please consider forking and building your version of ktlint until this will be merged or introduced in some similar way.

akuleshov7:
@Tapchicoma - we have already done it and are using fork inside of our ruleset: https://github.com/cqfn/diKTat

Tapchicoma:
@akuleshov7 what info from `Ktlint.Params` beside file name are you missing in the rule implementation? I would assume `debug` and maybe `script` flags?

BTW file name is already available to the rules: https://github.com/pinterest/ktlint/blob/427e2ae6d0623259ba69a606121d426b192a5bcb/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt#L203

akuleshov7:
@Tapchicoma I am pretty sure that you will agree with me that passing normal object of a normal class with strong typing to a visitor is much better than:
1) to use a custom map with custom string values (as .putUserData works)
2) to store a file name in each and every node

Creating objects with strong typing is much better than mapping values and keys on some magic constants. 

Today we need filePath from Params, tomorrow we will need other fields from Params. 
So I guess it will be a normal approach to pass Params to all rules (visitors) through the code.

akuleshov7:
@Tapchicoma @romtsn guys. if we are not able to merge this one PR - can we at least open API for RuleSet class?

romtsn:
tbh I'm fine with both of these changes, so I wouldn't mind merging this in, but I'd like @Tapchicoma and @shashachu opinions

Tapchicoma:
Personally I don't want to break `Rule` class api right now. As we already discussed some parameters you want to have are already passed via user data. If you need some additional parameters, feel free to add them.

Generally I would agree that user data usage is not the best approach and we should drop it when we will introduce next iteration of `Rule` class with some migration strategy.

akuleshov7:
> Personally I don't want to break `Rule` class api right now. As we already discussed some parameters you want to have are already passed via user data. If you need some additional parameters, feel free to add them.
> 
> Generally I would agree that user data usage is not the best approach and we should drop it when we will introduce next iteration of `Rule` class with some migration strategy.

Yes, that is fine. We are absolutely fine with merging at least https://github.com/pinterest/ktlint/pull/894

PR #809, opened 2020-07-09
Publish snapshots build against kotlin dev versions.

Tapchicoma:
Adds support for `-PkotlinDev` flag. This flag enables building ktlint against Kotlin development versions.

Additionally allows to publish this project snapshots that are built against dev versions. To be able to do that have to migrate publishing from 'com.vanniktech.maven.publish' plugin to generic Gradle 'maven-publish'.

Kotlin release and dev versions should be updated now in two places - `build.gradle` and `settings.gradle`. Will simplify this in the future.


build.gradle
```diff
@@ -23,6 +22,10 @@ ext.deps = [
   'jimfs'         : 'com.google.jimfs:jimfs:1.1'
 ]
 
+if (gradle.ext.isKotlinDev) {
+  project.properties.put("VERSION_NAME", "0.38.0-kotlin-dev-SNAPSHOT")
```
stephanenicolas, line 26: Shouldn't the version number be dynamic?

Tapchicoma: `-SNAPSHOT` means it will be dynamic

build.gradle
```diff
@@ -23,6 +22,12 @@ ext.deps = [
   'jimfs'         : 'com.google.jimfs:jimfs:1.1'
 ]
 
+if (gradle.ext.isKotlinDev) {
+  allprojects { p ->
+    p.ext."VERSION_NAME" = "0.38.0-kotlin-dev-SNAPSHOT"
```
romtsn, line 27: I think I would be also in favour of having kotlin version in the version name in case of a snapshot. Otherwise you'd always need to look up in this file, which kotlin version has been used to build the snapshot. Or maybe we could have a badge in readme to show the kotlin-dev version?

settings.gradle
```diff
@@ -2,7 +2,24 @@ pluginManagement {
...
+  plugins {
+    def kotlinVersion = settings.hasProperty("kotlinDev") ? "1.4-M3" : "1.3.70"
+    id 'org.jetbrains.kotlin.jvm' version kotlinVersion
+    id 'com.github.breadmoirai.github-release' version '2.2.12'
+    id 'com.github.johnrengelman.shadow' version '5.0.0'
+    id 'org.gradle.crypto.checksum' version '1.1.0'
+  }
+}
+
+// Pass '-PkotlinDev' to command line to enable kotlin-in-development version
```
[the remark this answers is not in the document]
Tapchicoma: Already done, please check latest changes.

build.gradle
Tapchicoma: I would prefer to not add it in version name, as you need always check if it is changed.

Badge could be an option and definitely changelog updates.

build.gradle
```diff
@@ -41,6 +46,12 @@ task ktlint(type: JavaExec, group: LifecycleBasePlugin.VERIFICATION_GROUP) {
 allprojects {
   repositories {
     mavenCentral()
+    gradlePluginPortal()
+
+    if (gradle.ext.isKotlinDev) {
+      maven { url 'https://dl.bintray.com/kotlin/kotlin-eap' }
+      maven { url 'https://kotlin.bintray.com/kotlinx' }
```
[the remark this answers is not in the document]
Tapchicoma: Seems added while I was testing building against dev version and forgot to remove. Removed in https://github.com/pinterest/ktlint/pull/809/commits/3a8a398f373c9efed83f9a18c906fd0339037195

build.gradle
shashachu: @Tapchicoma agree here. Can we pull the main version name from the `gradle.properties` file or something? Otherwise we have to remember to update `0.38.0` in 2 places when we next release.

Tapchicoma: sorry, missed here main point. Will update code to pull it from `gradle.properties`

Tapchicoma: Fixed via https://github.com/pinterest/ktlint/pull/809/commits/8f4fa138bb1f2dd972b001d4fd1773f511c5afe0

PR #811, opened 2020-07-12
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


shashachu:
Hi @sullis, sorry I'm not familiar with Dependabot. Could you add a little more information about what it does and what you're trying to accomplish?

Tapchicoma:
@shashachu it is a Github service that automatically updates project dependencies versions. Though support of Gradle is quite limited, see https://github.com/dependabot/dependabot-core/issues/1164

Generally I am not sure if it is good for this project:
- unnecessary noise by creating PRs for each dependency update, that could lead to several updates for the same dependency until next release
- doesn't work with Gradle dependency verification - dependabot just bumps versions

Issue #803, opened 2020-06-28
Internal Error (no-multi-spaces)


PaulWoitaschek:
The following file produces an internal error using ktlint 0.37.2

```kotlin
  package my.company.android
```

PR #817, opened 2020-07-19
Fix internal error on "no-multi-spaces" rule

lazysasha:
Fixes https://github.com/pinterest/ktlint/issues/803

The following exception was produced in case when a program starts with multiple spaces:
```java
Caused by: java.lang.IllegalStateException: node.treePrev must not be null
	at com.pinterest.ktlint.ruleset.standard.NoMultipleSpacesRule.visit(NoMultipleSpacesRule.kt:17)
```

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoMultipleSpacesRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoMultipleSpacesRule.kt
@@ -18,12 +18,21 @@ class NoMultipleSpacesRule : Rule("no-multi-spaces") {
             // allow multiple spaces in KDoc in case of KDOC_TAG for alignment, e.g.
             // @param foo      stuff
             // @param foobar   stuff2
-            !(node.treePrev.elementType == KDOC_MARKDOWN_LINK && node.treeParent.elementType == KDOC_TAG)
+            !isKdocMarkdownNode(node)
         ) {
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoMultipleSpacesRule.kt
```diff
@@ -18,12 +18,21 @@ class NoMultipleSpacesRule : Rule("no-multi-spaces") {
             // allow multiple spaces in KDoc in case of KDOC_TAG for alignment, e.g.
             // @param foo      stuff
             // @param foobar   stuff2
-            !(node.treePrev.elementType == KDOC_MARKDOWN_LINK && node.treeParent.elementType == KDOC_TAG)
```
romtsn, line 21: You could've simply put a `?` after `treePrev` and `treeParent` instead of introducing a new method :)

Tapchicoma:
@romtsn strange, I will check it now

Tapchicoma:
Seems one of the Travis caches were corrupted, deleting it - helped.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoMultipleSpacesRule.kt
lazysasha: Agreed, I've updated the code in ef5130b30609ab4701da47f3fc2b87b386ee9adc

PR #848, opened 2020-08-22
Compile ktlint with backwards compatibility to kotlin 1.3

jeremymailen:
This is needed to run any ktlint code within gradle since it bundles kotlin 1.3.72 at the moment.
It should not have negative consequences since the compiled code will be forward compatible with kotlin 1.4.

First off, thank you for your heroic efforts @shashachu and @Tapchicoma getting the 0.38.0 updates out to the community.

I ended up still running into a wrinkle using this in kotlinter-gradle since we execute ktlint code inside the gradle runtime.
See discussion here: https://github.com/jeremymailen/kotlinter-gradle/issues/144#issuecomment-678582319

At the moment, gradle still bundles kotlin 1.3.72 and there doesn't seem to be a way to push that forward and it probably wouldn't be safe given gradle dsl kotlin.

So, in that context we'll need ktlint compiled for 1.3 compatibility. As far as I know this shouldn't cause any issues unless the ktlint project was planning to use some language features or optimizations only available in 1.4 without a backwards compatible form (are there any?).

Thanks again.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,6 +67,7 @@ allprojects {
 
     kotlinOptions {
       jvmTarget = '1.8'
+      apiVersion = '1.3'
     }
   }
 }


henrik242:
@shashachu Are you planning a 0.38.1 release with this change any time soon? 🙂

shashachu:
@henrik242 working on it right now :)

shashachu:
0.38.1 released. Thanks everyone for your patience, and thanks @jeremymailen for the detective work. :)

shashachu:
@jeremymailen do you know if we can remove this change now? We're now getting build warnings saying apiVersion 1.3 is deprecated. I presume gradle has updated its kotlin version?

Tapchicoma:
@shashachu unfortunately not yet. Only starting with 6.8 version Gradle has bundled Kotlin 1.4 and 6.8 is still pretty fresh release.

jeremymailen:
> @jeremymailen do you know if we can remove this change now? We're now getting build warnings saying apiVersion 1.3 is deprecated. I presume gradle has updated its kotlin version?

Yes, at least for `kotlinter-gradle` we can move to 1.4. We ourselves use that api version and require gradle 6.8.3 and up for newer versions.

PR #849, opened 2020-08-22
IndentationRule: Do not enforce raw strings opening quote to be on a separate line

### LINKED ISSUE #711: Raw strings forced on a separate line by indent rule
KtLint 0.36.0

Raw strings don't seem to play well with indent experimental rule. For example this:

```kotlin
description = """
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris lacus arcu, blandit non semper elementum,
    """.trimIndent(),
```

reports an error: `[experimental:indent] Missing newline after "="`

This (weird) formatting:
```kotlin
description =
"""
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris lacus arcu, blandit non semper elementum,
    """.trimIndent(),
```
reports wrong indentation for the first `"""`, however this:
```kotlin
description =
    """
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris lacus arcu, blandit non semper elementum,
    """.trimIndent(),
```
is automatically reformatted by IntellIJ to the previous variant (with first `"""` not indented)

romtsn:
Fixes #711 

@shashachu I saw you've introduced this change in https://github.com/pinterest/ktlint/pull/387/commits/b54ea782bc5ee776efc69830e7c903523fc89178 but I think we should just leave raw strings alone because IDEA also does nothing about them (at least I haven't found any pattern) neither there are any mentions about it in the style guides

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
-- 
+### Fixed
+- Do not enforce raw strings opening quote to be on a separate line ([#711](https://github.com/pinterest/ktlint/issues/711))
 
 ## [0.38.0] - 2020-08-21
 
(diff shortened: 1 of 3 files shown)


shashachu:
Haha I don't remember why I made that change to be honest. I do wonder if we might be able to differentiate between multiline raw strings and single line raw strings (where I agree the `ktlint` formatting is broken). But for now we can just revert that change.

Issue #701, opened 2020-02-18
IDE integration via .editorconfig


Tapchicoma:
I am proposing to deprecate current IntellijIDEA integration. Problem with it:
- writes IDEA specific configuration xml files that could not be used by other editors
- is detached from used rules and not configurable. Only generic style is written to configuration file.

Instead I am proposing to define IDE kotlin code style via `.editorconfig` file: https://blog.jetbrains.com/idea/2019/06/managing-code-style-on-a-directory-level-with-editorconfig/. This overrides styles defined in IDEA xml configuration.

Rules themselves should provide applicable `.editorconfig` options as a constructor parameter. Current constructor that takes only `id` as a parameter should be deprecated. Then ktlint will collect all active rules editorconfig options and either write it to provided file or print them to standard output.

Possibly other editors could pickup this editorconfig options and apply it to internal kotlin code style settings (needs further investigation).

Problems with such approach:
- So far all kotlin specific style has `ij_kotlin` prefix, that may limit other editors integration. Editorconfig generated by IDEA based on current ktlint codestyle scheme:
```editorconfig
[{*.gradle.kts,*.kts,*.kt}]
ij_continuation_indent_size = 4
ij_kotlin_align_in_columns_case_branch = false
ij_kotlin_align_multiline_binary_operation = false
ij_kotlin_align_multiline_extends_list = false
ij_kotlin_align_multiline_method_parentheses = false
ij_kotlin_align_multiline_parameters = false
ij_kotlin_align_multiline_parameters_in_calls = false
ij_kotlin_assignment_wrap = normal
ij_kotlin_blank_lines_after_class_header = 0
ij_kotlin_blank_lines_around_block_when_branches = 0
ij_kotlin_block_comment_at_first_column = true
ij_kotlin_call_parameters_new_line_after_left_paren = true
ij_kotlin_call_parameters_right_paren_on_new_line = true
ij_kotlin_call_parameters_wrap = on_every_item
ij_kotlin_catch_on_new_line = false
ij_kotlin_class_annotation_wrap = split_into_lines
ij_kotlin_code_style_defaults = KOTLIN_OFFICIAL
ij_kotlin_continuation_indent_for_chained_calls = false
ij_kotlin_continuation_indent_for_expression_bodies = false
ij_kotlin_continuation_indent_in_argument_lists = false
ij_kotlin_continuation_indent_in_elvis = false
ij_kotlin_continuation_indent_in_if_conditions = false
[...]

PR #874, opened 2020-08-29
Initial implementation of IDE integration via editorconfig

Tapchicoma:
This PR brings base implementation of IDE integration via `.editorconfig` file:
- adds new interface `UsesEditorConfigProperties` that rule should implement to provide default values to generated `.editorconfig` file. Just as an example on how it will be -  'final-newline' rule was updated to support it. In followup PR all standard rules should be updated to support it.
- Adds new method in `KtLint` object - `generateKotlinEditorconfigSection(params)` that generates `.editorconfig` Kotlin section based on existing `.editorconfig` and default rules values. CLI support will be added in followup PR.

Additionally I've introduced two annotations - `FeatureInAlphaState` and `FeatureInBetaState` to indicate that this feature is experimental and api may change.

Update:
- Added `EditorConfigTestRule` to simplify writing `.editorconfig` file content for tests
- Two new `Rule` extension methods for tests that additionally accepts linted file path, so KtLint will load located in the same folder `.editorconfig` file.

Part of #701 issue


ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/EditorConfigGenerator.kt
```diff
@@ -0,0 +1,66 @@
...
+                    }
+                }
+
+                acc
+            }
+            .filterValues { it != null }
+            .map {
+                "${it.key} = ${it.value}"
+            }
+            .joinToString(separator = "\n")
```
romtsn, line 64: I have a feeling that on windows it should be something like `\r\n`, but not sure

ktlint-core/src/test/kotlin/com/pinterest/ktlint/core/internal/EditorConfigGeneratorTest.kt
```diff
@@ -0,0 +1,138 @@
...
+import com.pinterest.ktlint.core.api.UsesEditorConfigProperties
+import java.nio.file.FileSystem
+import java.nio.file.Files
+import java.nio.file.Path
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+import org.junit.Test
+
+@OptIn(FeatureInAlphaState::class)
+internal class EditorConfigGeneratorTest {
```
romtsn, line 17: This will be probably also failing on windows because the path separators should be replaced with `File.separator` I think

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/api/UsesEditorConfigProperties.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Provide a list of code style editorconfig properties, that rule validates.
+     *
+     * @return list of properties, for example:
+     * ```
+     * listOf("indent_size", "ij_kotlin_else_on_new_line")
+     * ```
+     */
+    public val editorconfigProperties: List<String>
```
romtsn, line 32: Did you think of introducing something like
```
data class EditorConfigProperty<T>(
  val key: String,
  val value: T?,
  val defaultValue: T
)
```
then we'd only need to return a list of those in this interface. Ideally `EditorConfig` would also contain those

Tapchicoma: No, I haven't. Will think if it is applicable here :thinking:

ktlint-core/src/test/kotlin/com/pinterest/ktlint/core/internal/EditorConfigGeneratorTest.kt
shashachu: It was suggested a while back we start running CI tests on Azure as well but I haven't yet looked in to it.

romtsn: I think we wanted to move to github actions instead where we could also run tests on windows

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/api/UsesEditorConfigProperties.kt
Tapchicoma: @romtsn please check 9025377 commit.

I haven't introduce `<T>` generic as editorconfig loading becomes too complex (imho) and `T` is only used inside rule. Rule could convert String internally into any representation.

ktlint-core/src/test/kotlin/com/pinterest/ktlint/core/internal/EditorConfigGeneratorTest.kt
Tapchicoma: `EditorConfigLoader` already supports loading on Windows fs. This class does not uses passed `filePath` to output anything.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/EditorConfigGenerator.kt
Tapchicoma: Fixed via a3fbfe4  :+1:

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/api/UsesEditorConfigProperties.kt
Tapchicoma: @romtsn and here is another new approach that uses ec4j `PropertyType`: fc1ab0f

romtsn:
LGTM 👍 looks way cleaner now (esp. on a calling side). What if we have a custom EC property (like ij_kotlin_something) - we'd need to create our own PropertyType, right?

romtsn (approved):

Tapchicoma:
> What if we have a custom EC property (like ij_kotlin_something) - we'd need to create our own PropertyType, right?

Yes, for that custom `PropertyType` needs to be created.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/api/UsesEditorConfigProperties.kt
```diff
@@ -0,0 +1,52 @@
...
+ * See [com.pinterest.ktlint.core.KtLint.generateKotlinEditorconfigSection] documentation how to generate
+ * `.editorconfig` based on [com.pinterest.ktlint.core.Rule]s with this interface implementations.
+ */
+@FeatureInAlphaState
+public interface UsesEditorConfigProperties {
+
+    /**
+     * Provide a list of code style editorconfig properties, that rule uses in linting.
+     */
+    public val editorconfigProperties: List<EditorConfigProperty<*>>
```
gnawf, line 28: Nit: should be `editorConfigProperties`

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -25,13 +29,18 @@ public object KtLint {
...
+        Key<EditorConfigProperties>("EDITOR_CONFIG_PROPERTIES")
     public val DISABLED_RULES: Key<Set<String>> = Key<Set<String>>("DISABLED_RULES")
     private const val UTF8_BOM = "\uFEFF"
     public const val STDIN_FILE: String = "<stdin>"
 
     private val psiFileFactory: PsiFileFactory = initPsiFileFactory()
     private val editorConfigLoader = EditorConfigLoader(FileSystems.getDefault())
 
+    @OptIn(FeatureInAlphaState::class)
+    private val editorconfigGenerator = EditorConfigGenerator(editorConfigLoader)
```
gnawf, line 42: Nit: should be `editorConfigGenerator`

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -336,6 +363,36 @@ public object KtLint {
...
+     * Rule should implement [UsesEditorConfigProperties] interface to support this.
+     *
+     * @return Kotlin section editorconfig content. For example:
+     * ```properties
+     * final-newline=true
+     * indent-size=4
+     * ```
+     */
+    @FeatureInAlphaState
+    public fun generateKotlinEditorconfigSection(
```
gnawf, line 382: Nit: should be `generateKotlinEditorConfigSection`

ktlint-core/src/test/kotlin/com/pinterest/ktlint/core/internal/EditorConfigGeneratorTest.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.ec4j.core.model.PropertyType
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+import org.junit.Test
+
+@OptIn(FeatureInAlphaState::class)
+internal class EditorConfigGeneratorTest {
+    private val tempFileSystem = Jimfs.newFileSystem(Configuration.forCurrentPlatform())
+    private val editorConfigLoader = EditorConfigLoader(tempFileSystem)
+    private val editorconfigGenerator = EditorConfigGenerator(editorConfigLoader)
```
gnawf, line 20: Nit: should be `editorConfigGenerator`

ktlint-core/src/test/kotlin/com/pinterest/ktlint/core/internal/EditorConfigGeneratorTest.kt
```diff
@@ -0,0 +1,146 @@
...
+    @Test
+    fun `Should use default rule value if property is missing`() {
+        tempFileSystem.writeEditorConfigFile(
+            rootDir,
+            """
+            root = true
+            """.trimIndent()
+        )
+
+        val generatedEditorconfig = editorconfigGenerator.generateEditorconfig(
```
gnawf, line 34: Nit: should be `generatedEditorConfig`

Same thing with all the tests here

gnawf (changes requested):
Sorry I was reading the code and couldn't help but notice these. I mean we're using this awesome lib because we're all pedantic… right?

Tapchicoma:
@gnawf thank you for noticing it, will fix nits :+1:

Tapchicoma:
@gnawf should be fixed via https://github.com/pinterest/ktlint/pull/874/commits/30cfe18d7b5dbc8b08d03460b482a8849170b1a4

Issue #868, opened 2020-08-27
experimental:annotation crash during formatting


shashachu:
## Observed Behavior
```
com.pinterest.ktlint.core.RuleExecutionException: java.lang.IndexOutOfBoundsException: Wrong offset: 29566. Should be in range: [0, 29472]
	at com.pinterest.ktlint.core.KtLint$format$2.invoke(KtLint.kt:386)
	at com.pinterest.ktlint.core.KtLint$format$2.invoke(KtLint.kt:43)
	at com.pinterest.ktlint.core.KtLint$visitor$2$1.invoke(KtLint.kt:248)
	at com.pinterest.ktlint.core.KtLint$visitor$2$1.invoke(KtLint.kt:43)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:574)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:575)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:575)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:575)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:575)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:575)
	at com.pinterest.ktlint.core.KtLint.visit(KtLint.kt:575)
	at com.pinterest.ktlint.core.KtLint.access$visit(KtLint.kt:43)
	at com.pinterest.ktlint.core.KtLint$visitor$2.invoke(KtLint.kt:246)
	at com.pinterest.ktlint.core.KtLint$visitor$2.invoke(KtLint.kt:43)
	at com.pinterest.ktlint.core.KtLint.format(KtLint.kt:371)
	at com.pinterest.ktlint.internal.FileUtilsKt.formatFile(FileUtils.kt:79)
	at com.pinterest.ktlint.KtlintCommandLine.process(Main.kt:331)
	at com.pinterest.ktlint.KtlintCommandLine.access$process(Main.kt:116)
	at com.pinterest.ktlint.KtlintCommandLine$lintFiles$2$1.call(Main.kt:257)
	at com.pinterest.ktlint.KtlintCommandLine$lintFiles$2$1.call(Main.kt:116)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #881, opened 2020-09-01
Idea for resolution to Issue #868

AdamMTGreenberg:
If we look at the stack trace we see:
```
Caused by: java.lang.IndexOutOfBoundsException: Wrong offset: 29566. Should be in range: [0, 29472]
	at org.jetbrains.kotlin.com.intellij.openapi.editor.impl.LineSet.findLineIndex(LineSet.java:178)
	at org.jetbrains.kotlin.com.intellij.openapi.editor.impl.DocumentImpl.getLineNumber(DocumentImpl.java:991)
	at com.pinterest.ktlint.core.ast.PackageKt.lineNumber(package.kt:235)
	at com.pinterest.ktlint.ruleset.experimental.AnnotationRule.visit(AnnotationRule.kt:149)
	at com.pinterest.ktlint.core.KtLint$format$2.invoke(KtLint.kt:376)
```
which corresponds to `val lineNumber = node.lineNumber()` - this shouldn't cause an error, but it does, most likely because the `node` was modified before.

I might have been trying to do too much in the combination of the two `AnnotationRule` with the multi-line stuff so I split them up. I played around with the idea of just searching for whitespace instead of line numbers, but if we are truly dealing with an issue with other rules then they could modify the lines as well.


AdamMTGreenberg:
Fix proposed for Issue #868

AdamMTGreenberg:
Tested by creating a file `test.kt`:
```
package a.b.c

class Test {
    fun bloop() {
        asdfadsf(asdfadsf, asdfasdf, asdfasdfasdfads,
        asdfasdf, asdfasdf, asdfasdf)
    }

    @Blah
    val test: Int
}
```
and then running:
```
./gradlew shadowJarExecutable &&  ./ktlint/build/run/ktlint --experimental -F /path/to/test.kt`

Result:
```
package a.b.c

class Test {
    fun bloop() {
        asdfadsf(
            asdfadsf,
            asdfasdf,
            asdfasdfasdfads,
            asdfasdf,
            asdfasdf,
            asdfasdf
        )
    }

    @Blah
    val test: Int
}
```

**Comment by AdamMTGreenberg** [comment]
I was able to achieve the same results without needing the line numbers - so removed it's usage. My initial assumption about the reuse of `node` was largely incorrect.

However, feel that this should be split into its own Rule since it differs enough certain people might want it in its own class or ignore rule

File: ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ExperimentalRuleSetProvider.kt [hunk 1]
```diff
@@ -3,7 +3,7 @@ package com.pinterest.ktlint.ruleset.experimental
 import com.pinterest.ktlint.core.RuleSet
 import com.pinterest.ktlint.core.RuleSetProvider
 
-class ExperimentalRuleSetProvider : RuleSetProvider {
+public class ExperimentalRuleSetProvider : RuleSetProvider {
```
shashachu, line 6: Was this change necessary?

ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/MultiLineAnnotationRule.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
+import org.jetbrains.kotlin.psi.psiUtil.children
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+
+/**
+ * Ensures there are not multiple line breaks between annotations and declarations.
+ *
+ * https://kotlinlang.org/docs/reference/coding-conventions.html#annotation-formatting
+ */
+class MultiLineAnnotationRule : Rule("multi-line-annotation") {
```
shashachu, line 24: Maybe a more descriptive name would be 'annotation-spacing' or something similar?

ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ExperimentalRuleSetProvider.kt
AdamMTGreenberg: Linter was complaining in the IDE that it needed an explicit `public` declaration

ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/MultiLineAnnotationRule.kt
AdamMTGreenberg: Updated to match your suggestions

PR #887, opened 2020-09-04
NoSemicolonsRule: don't report when semicolon is before annotation/comment/kdoc and lambda

### LINKED ISSUE #825: More false positive: unnecessary semicolon
More false positive edge cases like this: https://github.com/Mygod/VPNHotspot/blob/e333c859dfe4e675005a56049c510e5fb4ff270c/mobile/src/main/java/be/mygod/librootkotlinx/RootServer.kt#L498

_Originally posted by @Mygod in https://github.com/pinterest/ktlint/issues/703#issuecomment-670790143_

t-kameyama:
Fixes #825


romtsn:
Thanks!

PR #888, opened 2020-09-05
StringTemplateRule: don't report when template expression is keyword

### LINKED ISSUE #883: Auto-formatting redundant curly braces in specific string template fails to compile
## Expected Behavior
After formatting the following code (which *does* compile), the result should also compile.
Input (Bug.kt):
```kotlin
fun main() {
    println("This is: ${null}")
}
```
ktlint output:
```kotlin
fun main() {
    println("This is: $null")
}
```

## Observed Behavior
Compilation of the formatted code fails with this error: 
> Bug.kt:2:25: error: keyword cannot be used as a reference

## Steps to Reproduce
1. Format the input file using `ktlint -F Bug.kt`
2. Compile it: `kotlinc Bug.kt`
3. Compilation fails

## Your Environment
* ktlint: 0.37.2
* kotlinc: kotlinc-jvm 1.4.0 (JRE 13.0.1+9)

Note: I don't think it's particularly useful to use `${null}` in a string template, I just ran across this while formatting a large project. 😅

t-kameyama:
Fixes #883

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/StringTemplateRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/StringTemplateRule.kt
@@ -9,6 +9,10 @@ import com.pinterest.ktlint.core.ast.ElementType.LONG_STRING_TEMPLATE_ENTRY
 import com.pinterest.ktlint.core.ast.ElementType.SUPER_EXPRESSION
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
+import org.jetbrains.kotlin.psi.KtBlockStringTemplateEntry
+import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/StringTemplateRule.kt
```diff
@@ -41,13 +46,15 @@ class StringTemplateRule : Rule("string-template") {
                 ) {
                     emit(dot.startOffset, "Redundant \"toString()\" call in string template", true)
                     if (autoCorrect) {
+                        entryExpression = (entryExpression as? KtDotQualifiedExpression)?.receiverExpression
```
romtsn, line 49: What is this line for?

t-kameyama: If this line is not exists, `testFormat` will fail because `entryExpression` is KtDotQualifiedExpression.

`callExpression` of `dotQualifiedExpression` will be removed in the next line, so we save `receiverExpression` in `entryExpression`.

romtsn:
Thanks!

PR #894, opened 2020-09-07
Opening class RuleSet so it can be inherited.

akuleshov7:
### What's done:
In some cases some custom rule sets need to have different implementation of RuleSet. For example in the diktat (https://github.com/cqfn/diKTat) project it is used to pass ruleset specific configuration in the rules (config files). This should not affect anything and simply gives developers a chance to customise rule sets more.


akuleshov7:
This is a continuation of https://github.com/pinterest/ktlint/pull/806. We don't want to break ktlint API, so we can workaround by opening RuleSet class.

Tapchicoma (approved):
LGTM

PR #905, opened 2020-09-12
NoUnusedImportsRule: fix false positive with imported sub classes are used in DOT_QUALIFIED_EXPRESSION

### LINKED ISSUE #902: Unused import issue with 38.1
## Expected Behavior
I expect the following to pass formatting.

```kotlin
package com.zak.test

import com.zak.test.MyClass.Intention.Intention1
import com.zak.test.MyClass.Intention.Intention2
import kotlinx.coroutines.CoroutineScope

class MyClass(
    scope: CoroutineScope
) : CoroutineScope by scope {
    private val lambda = {
        var currentState = State()

        for (intention in listOf(Intention1(), Intention2())) {
            when (intention) {
                is Intention1 -> currentState // import is used here
                is Intention2 -> currentState
            }
        }
    }

    fun someFunction() {
        Intention.Intention1() // notice `Intention.` prefix is used even though it is imported
    }

    class State

    sealed class Intention {
        class Intention1 : Intention()
        class Intention2 : Intention()
    }
}

```

## Observed Behavior
Fails with: `Unused import (no-unused-imports)`

If I remove the `someFunction()`, or define it like so:

```kotlin
fun someFunction() {
    Intention1() // import is used, which ktlint likes
}
```

## Steps to Reproduce
run ktlint over the above file ^^

## Your Environment
* Version of ktlint used: `0.38.1` (command line)
* AS `4.0.1` with AGP `3.6`
* Operating System and version: MacOS `10.15.6`

t-kameyama:
Fixes #902

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoUnusedImportsRule.kt
@@ -12,9 +12,12 @@ import com.pinterest.ktlint.core.ast.isRoot
 import com.pinterest.ktlint.core.ast.parent
 import com.pinterest.ktlint.core.ast.visit
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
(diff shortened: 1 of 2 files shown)


Tapchicoma (approved):
LGTM

PR #911, opened 2020-09-17
NoUnusedImportsRule: fix false positive when import directive has backticks and alias

### LINKED ISSUE #910: Unused import thrown when using `` in import
## Expected Behavior
I expect the following code to pass lint check.

```
import com.test.`object`.Producer as Producer1

class Consumer(producer1: Producer1<String>) {
    val x = 1
}
```

## Observed Behavior

But it fails with `ktlint-repro/src/main/kotlin/Consumer.kt:1:1: Unused import`

If I refactor ``com.test.`object`.Producer`` to a package name that does not require ` then ktlint passes.

## Steps to Reproduce
Run ktlint on the above file.

## Your Environment

* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): Gradle plugin "org.jlleitschuh.gradle.ktlint" version "9.4.0"
* Version of Gradle used (if applicable): 6.1.1
* Operating System and version: Mac OS
* Link to your project (if it's a public repository): N/A

t-kameyama:
Fixes #910


romtsn:
Thanks!

PR #923, opened 2020-09-22
Add 'generateEditorConfig' CLI subcommand.

Tapchicoma:
Initial implementation of `.editorconfig` Kotlin section generation via CLI subcommand. For now it just outputs generated content to stdout.

Part of #701 issue.

### DIFF
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
@@ -29,7 +29,12 @@ internal fun List<String>.fileSequence(): Sequence<File> {
         .map(Path::toFile)
 }
 
-internal fun List<String>.toFilesURIList() = map {
+/**
+ * List of paths to Java `jar` files.
+ */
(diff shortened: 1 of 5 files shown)


ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
```diff
@@ -28,9 +29,18 @@ internal fun List<String>.fileSequence(): Sequence<File> {
         .map(Path::toFile)
 }
 
+internal fun List<String>.toFilesURIList() = map {
```
romtsn, line 32: Maybe we should introduce some `typealias Rulesets` and have extension functions for that to avoid namespace pollution?

romtsn (approved):

ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
Tapchicoma: Good idea, will do :+1:

PR #947, opened 2020-10-22
Update README.md

PavelPolyakov:
When pasting this to `pom.xml` it corrupts it. Because, it's not allowed to have `--` inside of comments.
You can paste it, for example, here: https://www.xmlvalidation.com/ to check.

After pasting the suggested snippet Maven stops working, I don't see another way to cure it, let's just remove it then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,11 +198,6 @@ $ ktlint installGitPreCommitHook
                 <java taskname="ktlint" dir="${basedir}" fork="true" failonerror="true"
                     classpathref="maven.plugin.classpath" classname="com.pinterest.ktlint.Main">
                     <arg value="src/**/*.kt"/>
-                    <!-- to generate report in checkstyle format prepend following args: -->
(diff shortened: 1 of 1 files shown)


shashachu:
I think it's okay as-is; you can either remove the block on your own pom.xml or uncomment those blocks in which case it will work.

PavelPolyakov:
@shashachu 
Yes, you can fix this, but you say in your guide "Add the following snippet to pom.xml" ([link](https://ktlint.github.io/) and developers would expect that it should just work.
You don't say "Add the following snippet to pom.xml, but it's actually a broken xml, you need to figure this out yourself".

I'm disappointed you've just closed this without changing the docs, because I can't say that current UX was great for me. I tried to make it better for others.

Please, reconsider, this needs to be fixed somehow, so people don't waste extra time debugging this, like I did.

Regards,

PR #977, opened 2020-11-21
Adding new option disable-standard-rules

akuleshov7:
### What's done:
Adding special option `disable-standard-rules` that is extremely useful and needed when `ktlint` is used as a framework for static analysis. In some cases when `custom ruleset` is used - there is no need to use `standard ruleset` as it can lead to conflicts.

In this case we need a mechanism to disable standard rules. This change will not break any API and the default behavior will remain the same.

### DIFF
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
@@ -149,8 +149,10 @@ class KtlintCommandLine {
 
     @Option(
         names = ["--disabled_rules"],
-        description = ["Comma-separated list of rules to globally disable." +
-            " To disable standard ktlint rule-set use --disabled_rules=standard"]
+        description = [
+            "Comma-separated list of rules to globally disable." +
+            " To disable standard ktlint rule-set use --disabled_rules=standard"
+        ]
     )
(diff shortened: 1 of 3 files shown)


akuleshov7:
@Tapchicoma please have a look when you will have a minute. Thanks!

akuleshov7:
@romtsn Roma, may be you will have a chance to have a look?

romtsn:
@akuleshov7 looks reasonable (for your case), but what do you think of extending `disabled_rules` option so you can also pass a ruleset name there (not only rules name)? then we could just `--disabled_rules=standard` and not introduce a new cli option

akuleshov7:
> @akuleshov7 looks reasonable (for your case), but what do you think of extending `disabled_rules` option so you can also pass a ruleset name there (not only rules name)? then we could just `--disabled_rules=standard` and not introduce a new cli option

Рома, updated, please have a look. This looks a little bit less user-friendly, but works fine,

romtsn:
Looks fine to me, but PR needs a fix :)

akuleshov7:
> Looks fine to me, but PR needs a fix :)
Oh, I have violated code formatting in code formatter, lol :D

@romtsn Fixed

PR #984, opened 2020-11-30
Fix Ktlint CLI could skip checking some of explicetly passed files.

### LINKED ISSUE #940: Run Ktlint on a specific folder
Is there any way to specify the folder that ktlint will run on android like: ./gradlew location ktlint Or something like that?

Because in large projects, using ktlint and its format function would make pull request larger than expected. (Yes, with fixed codestyle, but for this situation, I want to select where ktlint will run).

### LINKED ISSUE #942: ktlint doesn't report issues for all files, if path of one file is a subpath of another
## Observed Behavior
If file paths are passed in the command line, and path of one file is a subpath of another file, then `ktlint` only reports issues of the file with the shortest path.

For example if there're 2 files with issues:
1. `/src/main/java/org/example/bugtest/SampleBadClass.kt`
2. `/src/main/java/org/example/bug/AnotherBadClass.kt`

then only the issues of the `AnotherBadClass.kt` are reported.

This only happens if one's file path is a substring of another's file path. In the example above: `/src/main/java/org/example/bug` is a substring of `src/main/java/org/example/bugtest`. The case would also reproduce on the following paths of the second file:
* `/src/main/java/org/example/bugtes/AnotherBadClass.kt`
* `/src/main/java/org/example/bugte/AnotherBadClass.kt`
* `/src/main/java/org/example/b/AnotherBadClass.kt`
 

If the file with the shortest path doesn't have issues or that file doesn't exist then no issues are reported at all:

1. `/src/main/java/org/example/bugtest/SampleBadClass.kt`
2. `/src/main/java/org/example/bug/NonExistingFile.kt`

## Expected Behavior
Issues from all files are correctly reported

### Files setup

`/src/main/java/org/example/bugtest/SampleBadClass.kt`:
```kotlin
package org.example.bugtest

// Unnecessary block ("{}")
// Unexpected blank line(s) before "}"
// Needless blank line(s)
object SampleBadClass {


}
```

`/src/main/java/org/example/bug/AnotherBadClass.kt`:
```kotlin
package org.example.bug

// unnecessary block {}
object AnotherBadClass {}
```

`/src/main/java/org/example/bug/AnotherGoodClass.kt`:
```kotlin
package org.example.bug

// no issues
object AnotherGoodClass {}
```

### Check results
`ktlint "src/main/java/org/example/bugtest/SampleBadClass.kt"` finds issues as expected:
```
/Users/…/src/main/java/org/example/bugtest/SampleClass.kt:3:20: Unnecessary block ("{}")
/Users/…/src/main/java/org/example/bugtest/SampleClass.kt:4:1: Unexpected blank line(s) before "}"
/Users/…/src/main/java/org/example/bugtest/SampleClass.kt:5:1: Needless blank line(s)
```

`ktlint "src/main/java/org/example/bugtest/SampleBadClass.kt" "src/main/java/org/example/bug/AnotherGoodClass.kt"` reports no issues (issues in `SampleBadClass` found before are now **not** reported):
```
```

`ktlint "src/main/java/org/example/bugtest/SampleBadClass.kt" "src/main/java/org/example/bug/AnotherBadClass.kt"` reports only issues from `AnotherBadClass.kt` (issues in `SampleBadClass` found before are again **not** reported):
```
/Users/…/src/main/java/org/example/bug/AnotherBadClass.kt:4:24: Unnecessary block ("{}")
```

The bug also reproduces if multiple wildcards are used:
`ktlint "src/main/java/org/example/bug/*.kt" "src/main/java/org/example/bugtest/*.kt"` only reports issues from `AnotherBadClass.kt` skipping `SampleBadClass.kt`:
```
/Users/…/src/main/java/org/example/bug/AnotherBadClass.kt:4:24: Unnecessary block ("{}")
``` 

## Additional info
ktlint gradle plugin runs `ktlint` exaclty like in the examples above, specifying a path to every single file. That currently leads to inconsistent reports between `ktlint` command line run (without file paths) and `gradlew ktlintCheck`.

## Your Environment

* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): command line
* Version of Gradle used (if applicable):
* Operating System and version: macOS 10.15.7

Tapchicoma:
Bug is in external dependency [klob](https://github.com/shyiko/klob). Seems it is unmaintined, probably worth to think about alternatives :thinking: 

Added a workaround for now.

Fixes #942


RedefineMike:
The workaround for this issue kills the ability to use negated paths ie "!**/generated/**" because it is only additive in the way it combines the paths. I'm looking at forking klob and fixing this now so both problems are solved if I can.

Issue #753, opened 2020-05-30
Update ImportOrderingRule once Kotlin plugin has support for imports_layout in editorconfig 


romtsn:
As a follow up for #749, there is a number of actions we need to perform after Kotlin plugin is released with support for imports layout via .editorconfig.

- [x] Possibly rename the `ij_kotlin_imports_layout` property so it matches the one from Kotlin plugin (unlikely though)
- [x]  Deprecate `idea` and `ascii` predefined styles and print a warning in case of their appearances
- [x]  Update editorconfig documentation and show how `idea` and `ascii` can be achieved by setting a custom import layout pattern
- [x]  Update how "alias" pattern is represented in the editorconfig, so Kotlin plugin understands it

romtsn:
https://youtrack.jetbrains.com/issue/KT-40353

PR #998, opened 2020-12-07
Align with kotlin plugin on how subpackages are represented

romtsn:
Am not sure why they use a double star to indicate "with subpackages" behavior, and also, even when I specified a full path without subpackages and exported the `.editorconfig` file in IDEA, it still had a single star in it:

image

```
ij_kotlin_imports_layout = *,java.**,javax.**,kotlin.**,java.nio.file.Paths.*,^
```
Maybe this means "all members of an import" - don't know.

As usual source of truth: https://github.com/JetBrains/intellij-kotlin/blob/master/idea/src/org/jetbrains/kotlin/idea/formatter/KotlinPackageEntryTableAccessor.kt


Tapchicoma:
Could you add a link to `KotlinPackageEntryTableAccessor` into `ImportOrderingRule`?

romtsn:
Added, but rather to `ImportLayoutParser` as it's a better place for that link I think

romtsn:
Part of #753

PR #1001, opened 2020-12-08
Fix Overlapping Paths / Negated Patterns Issues

### LINKED ISSUE #999: Exclusive / Negated Globs No Longer Work
## Expected Behavior
When I run ktlint with `**/*.kt` & `!**/generated/**`, all files in any subfolder of a folder named generated should be excluded.

## Observed Behavior
After many tests, and reviewing the code, negated patterns are completely ineffective.

## Steps to Reproduce
1. Make a folder called "generated" in your source code, with kt files inside.
2. Run ktlint with the `**/*.kt` & `!**/generated/**` patterns and note that files in the generated folder are still listed.

## Your Environment
Gradle, Linux

## Solution
This issue was originally caused by the workaround in issue #942 Instead of using the Glob.from(vararg pattern) method, it iterates each pattern, collecting all the matching files, which prevents the exclusion pattern from working.

I have fixed the original issue in the `klob` dependency so that the workaround is no longer necessary. I will submit a pull request of the correction to ktlint.

thecoden:
Fixes #999

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ ext.deps = [
     'stdlib'  : "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}",
     'compiler': "org.jetbrains.kotlin:kotlin-compiler-embeddable:${versions.kotlin}"
   ],
-  'klob'          : 'com.github.shyiko.klob:klob:0.2.1',
+  'klob'          : 'io.codetactics.klob:klob:0.2.1',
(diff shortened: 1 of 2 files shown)


thecoden:
It looks like the build failed due to the new klob dependency not being in maven central (it is in jcenter).

Tapchicoma:
Closing this PR in favor of #1061.

Thank you for your help, ideas proposal and review :clap:

Issue #1007, opened 2020-12-12
Ignore back ticked function names in max_line_length rule


paul-dingemans:
When writing unit tests, it is a common pattern enclose the function name enclosed between back-ticks. Those function names can however not be split into multiple lines. Therefore such function names should be excludes from the max_line_length rule.

```
@Test
fun `Should return something when passing something and something and the other thing is in this or that state`() {...}
```
_Originally posted by @k1dbl4ck in https://github.com/pinterest/ktlint/issues/659#issuecomment-664319682_
Would like to add that back ticked function names in tests should also be handled / ignored:

Tapchicoma:
I am not sure that ktlint should allow such behaviour by default. In the the end max line constraint exists to ensure that code is readable on width constraint text areas.

Probably could be optionally enabled via `.editorconfig`.

pauldingemans:
That is a fair point. I will make it optional.

romtsn:
Closed by #1008

PR #1008, opened 2020-12-12
Ignore tokens between backticks in rule max line length (#1007)

paul-dingemans:
Tokens starting and ending with a backtick should be ignored entirely when validating whether a line exceeds the maximum length of a line. Such tokens can not be spit into multiple lines. For example when writing JUnit test it is a common pattern to write the function names between back ticks instead of using the DisplayName annotation.

Currently I just ignore all ASTNodes starting and ending with a back tick. Such a node no longer counts in the calculation of the length of the line. This is not explicitly limited to method names neither to unit test only. However I do think this approach is valid. Tokens can only contain back ticked values whenever this is allowed by Kotlin. Such tokens can never be split in multiple lines.


paul-dingemans:
The identifiers between back ticks can now optionally be ignored. By default this option is disabled so the behaviour is not changed for existing users.

paul-dingemans:
@Tapchicoma Sorry that I reach out to you in this way but I have found no other way to do so. I am wondering what else I can do to get my PR's merged. I am aware that you are volunteer on this project. I am willingly to give a long term commitment on the project if you would appreciate that. But for that I need to know how my contributions will be merged more smoothly.

Tapchicoma:
@paul-dingemans sorry for missing your last comment in this PR.

As far as I know @romtsn want to review your PRs. 

Regarding #1038 PR, that I've started to review, could you split it into two:
 - fixing actual #1029 issue plus adding tests for `SuppressedRegionLocator`
 - further refactoring to KtLint you've introduced in that PR

romtsn:
hi @paul-dingemans, sorry for long-awaited review, I've just changed 3 things:
* Changed how we use editorconfig properties (see #701 and associated PRs for the new API - we shouldn't introduce new fields for the EditorConfig interface from now on)
* Renamed the custom property to start with `ktlint_` following your own suggestion in #1033, jus
* Changed README a bit, so we don't have a new section (which is already quite bloated)

I will take your frustration on me, as I planned to do the aforementioned changes way earlier, but then life happened. Your PRs are of course appreciated, just one wish would be to make your PRs smaller and target one specific issue/thing (I'm referring to your other PRs, this one looks good). thanks for the contribution!

PR #1009, opened 2020-12-13
Fix snapshot publishing

romtsn:
According to https://github.com/johnrengelman/shadow/issues/586 the fat jar is now automatically published, so we do not need to specify it explicitly (this was causing the issue of multiple artifacts with the same classifier). Tested locally with publishToMavenLocal task.

### DIFF
--- a/ktlint/build.gradle
+++ b/ktlint/build.gradle
@@ -19,10 +19,6 @@ shadowJar {
   mergeServiceFiles()
 }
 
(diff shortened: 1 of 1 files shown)


shashachu:
Looks fine to fix publishing but let's keep an eye on the original issue, as some people consider it a bug so it's possible the behavior will be reverted.

PR #1011, opened 2020-12-14
Fix import-ordering editorconifg generation

romtsn:
@Tapchicoma not sure why you converted the properties to `List<PatternEntry>` but writing to editorconfig didn't work properly, because:

- PatternEntry's `toString` method was messed up
- By default `List<PatternEntry>.toString()` returns something like `[*,kotlin.**,java.**,^]` - mind the braces. This is an incorrect value if we wanna write to the standard `ij_kotlin_imports_layout`, to properly generate it we would need to use `joinToString` but we don't know the type in the `getEditorConfigValue`..

So the easiest path I found is to revert properties to be strings, but not sure if it's the best approach. Maybe we could introduce some interface like `EditorconfigWriteable` which has a `toString` method and put it as an upper bound for generic `EditorConfigProperty<T : EditorconfigWriteable>`

### DIFF
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/api/UsesEditorConfigProperties.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/api/UsesEditorConfigProperties.kt
@@ -40,17 +40,31 @@ public interface UsesEditorConfigProperties {
     }
 
     /**
+     * Write the string representation of [EditorConfigProperty]
+     */
+    public fun <T> EditorConfigProperties.writeEditorConfigProperty(
+        property: EditorConfigProperty<T>,
+        isAndroidCodeStyle: Boolean
+    ): String {
+        return property.propertyWriter(getEditorConfigValue(property, isAndroidCodeStyle))
+    }
+
+    /**
      * Supported `.editorconfig` property.
      *
      * @param type type of property. Could be one of default ones (see [PropertyType.STANDARD_TYPES]) or custom one.
      * @param defaultValue default value for property if it does not exist in loaded properties.
      * @param defaultAndroidValue default value for android codestyle. You should set different value only when it
      * differs from [defaultValue].
+     * @param propertyWriter custom function that represents [T] as String. Defaults to the standard `toString()` call.
+     * You should override the default implementation in case you need a different behavior than the standard `toString()`
+     * (e.g. for collections joinToString() is more applicable).
      */
     public data class EditorConfigProperty<T>(
         public val type: PropertyType<T>,
         public val defaultValue: T,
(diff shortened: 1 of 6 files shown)


Tapchicoma:
@romtsn this experimental api is in alpha state so it is fine to find such issues and fix them :slightly_smiling_face: Thank you for finding this one :clap:  

Regarding the issue itself - current code should correctly parse `*,kotlin.**,java.**,^` values (without `[` and `]`), check [ImportLayoutParserTest](https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/importordering/ImportLayoutParserTest.kt). Ec4j library does not modify property value and passes it as-is into provided parser.

The real "issue" is in `.editorconfig` generator - it uses default `T.toString()` implementation to write back this fields. In this particular case  it is `AbstractCollection.toSting()` one.

My vision how it could be fixed:
- `EditorConfigProperty` will get new field `propertyWriter` that by default will use `toString()`:
```kotlin
    public data class EditorConfigProperty<T>(
        public val type: PropertyType<T>,
        public val defaultValue: T,
        public val defaultAndroidValue: T = defaultValue,
        public val propertyWriter: (T) -> String = { it.toString() }
    )
```
- `UsesEditorConfigProperties` interface will get new method:
```kotlin
    public fun <T> EditorConfigProperties.writeEditorConfigProperty(
        property: EditorConfigProperty<T>,
        isAndroidCodeStyle: Boolean
    ): String {
        return property.propertyWriter(getEditorConfigValue(property, isAndroidCodeStyle))
    }
```
- `EditorConfigGenerator` will use `writeEditorConfigProperty` new method  instead of `.toString()`

Then to fix `ImportOrderingRule` properties should provide non-default `propertyWriter`, something like `it.joinToString()`

Tapchicoma:
> not sure why you converted the properties to `List<PatternEntry>`

imho it makes rule code easier as you not deal with string of patterns, but already parsed list of pattern that it is easier to work with

romtsn:
> imho it makes rule code easier as you not deal with string of patterns, but already parsed list of pattern that it is easier to work with

Yeah, makes sense 👍 

> Regarding the issue itself - current code should correctly parse *,kotlin.**,java.**,^ values (without [ and ]), check ImportLayoutParserTest. Ec4j library does not modify property value and passes it as-is into provided parser.

That's true, it will correctly parse it. But when you try to write it back using `PatternEntry.toString()` method, it will not be the same (but that's because I did not implement the toString method correctly initially - already fixed)

As for your proposal, it looks good I will follow up on this soon 👍

romtsn:
@Tapchicoma updated accordingly 👍

Tapchicoma (approved):
LGTM!

PR #1017, opened 2020-12-17
Add clarification for IDEA setup in README

Woodz:
To clarify how the IntelliJ IDEA code style setting configuration process works, add an brief explanation about what each command does and clarify that the code samples are command line calls, not config file entries

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,9 @@ val ktlintFormat by tasks.creating(JavaExec::class) {
 
 ##### Option #1 (recommended)
 
+To change the code style config files in a single IDEA project
+
+Run ktlint executable with the appropriate flag:
 > (inside project's root directory)  
 
 ```sh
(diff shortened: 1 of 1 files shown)


romtsn:
Thanks for the PR, but I am not sure how it wasn't clear from the previous readme?

Woodz:
> Thanks for the PR, but I am not sure how it wasn't clear from the previous readme?

Given that the majority of users will not use the ktlint executable directly (instead would run via build tool plugin), it was not clear whether `ktlint applyToIDEAProject` was a shell command or a setting in a file. I misunderstood that this was a config to be used inside `.editorconfig` file, which obviously didn't work. It was only when I re-read the entire README and found the "Installation" section that mentioned that it could be run as a standalone binary that I understood that this was a shell command. I think part of the issue is that the `.sh` code highlighting in GitHub does not show that it is a shell script, so even though the previous README contained the full information needed to understand how to use it, the way GitHub presented it left some room for confusion.

Since I was silly and wasted half an hour getting confused, I really wanted to avoid it happening again, hence this PR :-)

romtsn:
Oh ok, I see.. Well, if that would help you, I am fine with merging this in 👍

Woodz:
> Oh ok, I see.. Well, if that would help you, I am fine with merging this in 👍

Thanks! Appreciate your time reviewing and merging this in :-)

Issue #548, opened 2019-07-29
Introduce codestyle set


Tapchicoma:
## Proposal

Ktlint should introduce codestyle-set concept. Codestyle-set will contain set of rules, that would be used by ktlint on linting/formatting the code.

Such approach:
- reduces the level of user confusion (see #533, #527, #544) by allowing changes in defaults to be more conservative
- still easy to consume
- is flexible. User can change codestyle-set to another or custom one. Custom codestyle-set can extend existing codestyle-set, rules can be reused in different codestyle-sets.
- allows without much effort introduce support for other Kotlin code styles in the future (like Google Kotlin codestyle).

## Implementation details

Each codestyle-set should have a unique name and contain list of rule ids. Codestyle-set could extend another codestyle-set by adding (removing (?)) rules to the parent one.

### Built-in

Ktlint should provide built-in codestyle-sets:
- "official" (default) provides official Kotlin styleguide and should works the same way as IDEA Kotlin code style formatting
- "official-experimental" - adds experimental rules on top of "official" codestyle-set
- "android" - provides official Android Kotlin styleguide support
- "android-experimental" - adds experimental rules on top of "android" codestyle-set

### `.Editorconfig`

User should be able to define and use custom codestyle-set, using `.editorconfig` file:
```editorconfig
[*.{kt,kts}]
codestyle-set-define=my-kotlin-style-name:final-newline,chain-wrapping,no-unit-return
codestyle-set-define=my-kotlin-style-name-extended:my-kotlin-style-name:no-unused-imports,comma-spacing
codestyle-set=my-kotlin-style-name
```

where `codestyle-set-define` has following format:
```editorconfig
codestyle-set-define=<set name>:<optional parent set name>:<list of rule ids separated by comma>
```

`.Editorconfig` can contain several `codestyle-set` entries, but only last one `codestyle-set` entry will be used by Ktlint.
Nested `.editorconfig` files could override `codestyle-set` entry used in parent `.editorconfig`.

### Command line

Usage of codestyle-set in command line looks similar to how it is done in `.editorconfig` file:
```bash
$ ktlint --codestyle-set-define=my-kotlin-style-name:official:final-newline,chain-wrapping,no-unit-return --codestyle-set=my-kotlin-style-name
```

## Deprecations

`--android` CLI flag should be replace by "android" codestyle-set

paul-dingemans:
> ## Deprecations
> `--android` CLI flag should be replace by "android" codestyle-set

[...]

PR #1023, opened 2020-12-19
Add option to enable specific experimental rules

paul-dingemans:
With command line flag "--experimental" it is possible to enable all experimental rules. I want to be able to enable specific experimental rules which work well for a repository from the .editorconfig in that repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 ### Added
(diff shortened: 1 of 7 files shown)


paul-dingemans:
Alternative to this implementation, is my proposal in #1033 to restructure the properties in the `.editorconfig`.

romtsn:
Thanks for the PR, but we'd like to move towards a more flexible solution (#548). Also, I'd suggest you open an issue upfront, so we could discuss it and you wouldn't need to spend your time implementing something that's not in line with the project roadmap.

PR #1024, opened 2020-12-19
Fix (custom) import rule in case no imports are found for the first group

paul-dingemans:
In my .editorconfig file I have defined a custom import order as follows:
```
kotlin_imports_layout = empty.group1.**,|,empty.group2.**,|,*,|,com.group1.**,|,com.group2.**
```

All files in which no imports are found for `empty.group1` fail on the `no-consecutive-blank-lines` rule. For example:
```
            import com.group2.b
            import com.group2.a
            import com.group1.b
            import com.group1.a
```

In the `import-ordering` rule should not add blank lines for groups until at least one import has been processed.

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
@@ -7,6 +7,7 @@ import com.pinterest.ktlint.core.api.FeatureInAlphaState
 import com.pinterest.ktlint.core.api.UsesEditorConfigProperties
 import com.pinterest.ktlint.core.ast.ElementType
 import com.pinterest.ktlint.core.ast.isRoot
+import com.pinterest.ktlint.core.ast.isWhiteSpace
 import com.pinterest.ktlint.ruleset.standard.ImportOrderingRule.Companion.ASCII_PATTERN
 import com.pinterest.ktlint.ruleset.standard.ImportOrderingRule.Companion.IDEA_PATTERN
 import com.pinterest.ktlint.ruleset.standard.internal.importordering.ImportSorter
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
```diff
@@ -216,7 +217,9 @@ public class ImportOrderingRule :
                     if (autoCorrect && canAutoCorrect) {
                         node.removeRange(node.firstChildNode, node.lastChildNode.treeNext)
                         sortedImportsWithSpaces.reduce { current, next ->
-                            node.addChild(current, null)
+                            if (!(current.isWhiteSpace() && node.firstChildNode == null)) {
```
romtsn, line 220: what is this if-check for? just tried it locally and it worked without

paul-dingemans: If the file does not contain imports for "empty.group1" then the current node will be a WhiteSpace node containing "\n\n" which would lead to a consecutive blank line violation. Now the node is only added when a previous import was already added or when the first node actually isn't empty.

romtsn: could you add a test case for this as well? as I mentioned I tried to remove this check locally and run the tests - they passed.

paul-dingemans: Strangely enough I can not reproduce the original problem anymore. The original code gives same output as the modified code. I propose to close the PR without merging it.

romtsn: ok, let us know if we need to reopen or you find a way to reproduce it

Issue #1019, opened 2020-12-17
ktlint format removing used imports


charlesritchea:
Using ktlint version 0.40.0

First off, the pattern of using imports 
Running ktlint format on the following:

```kotlin
import com.fasterxml.jackson.databind.JsonNode
import com.foo.core.api.common.RouteConstants
import com.foo.core.api.model.PredictionModel
import java.net.URI
import com.foo.core.api.model.PredictionGeometric2D as ModeledObject
import com.foo.core.api.model.PredictionModel as ParentModel

class PredictionGeometric2DModel : PredictionModel() {
```

Results in:
```kotlin
package com.foo.core.api.model

import com.fasterxml.jackson.databind.JsonNode
import com.foo.core.api.common.RouteConstants
import java.net.URI
import com.foo.core.api.model.PredictionGeometric2D as ModeledObject
import com.foo.core.api.model.PredictionModel as ParentModel

class PredictionGeometric2DModel : PredictionModel() 
```

Which won't compile because PredictionModel is no longer imported as PredictionModel only as ParentModel

Tapchicoma:
@romtsn could this be a bug in import-ordering rule?

romtsn:
@Tapchicoma nope, that's unused-imports it seems, have tried it.

@charlesritchea but does it even make sense to have the same import twice (with and w/o an alias) - what's the usecase for this?

charlesritchea:
My co-workers like to do it lol. Apparently it was so they could use the
same ParentObject or DomainObject boilerplate types then just modify the
imports. Let's just say IMHO ktlint is fixing the tip of the bikeshedding
iceberg, I've got people using extension functions in the same file as The
class is defined, and all kinds of weirdness. Anyway, does it matter why,
it's valid kotlin?

charlesritchea:
Preaching to the choir 😉

romtsn:
could you post the full class body, or a minimal reproducible sample (and also maybe run ktlint with --verbose flag)? I couldn't reproduce it with:

```Kotlin
import com.foo.core.api.model.PredictionModel
import com.foo.core.api.model.PredictionModel as ParentModel

class PredictionGeometric2DModel : PredictionModel() {
    lateinit var parent: ParentModel
}
```

charlesritchea:
I can't spend anymore time on this right now, feel free to close, If I get
around to a reproducible example I'll let you know

PR #1026, opened 2020-12-20
Fix string-template for dot-expression from which the  toString() is removed

### LINKED ISSUE #996: Internal error no-unused-imports
## Observed Behavior
Running ktlint with parameter "-F" produces an internal error "no-unused-imports" on enclosed sample code. Running *without* parameter "-F" actually produces a helpful message.

## Steps to Reproduce
Run "ktlint -F" on code below:
```
package com.example.ktlint.error

import mu.KLogging

class NoUnsedImportInternalErrorKtlint {

    fun getDrafts(val draftsIds: List<Long>) {
        logger.info { "Entering getDrafts: draftIds=[${draftsIds.toString()}]" }
    }

    private companion object : KLogging() {
    }
}
```

results in:
`
NoUnsedImportInternalErrorKtlint.kt:1:1: Internal Error (no-unused-imports). Please create a ticket at https://github.com/pinterest/ktlint/issues (if possible, provide the source code that triggered an error) (cannot be auto-corrected)
`

without parameter "-F", the result is:
`
NoUnsedImportInternalErrorKtlint.kt:8:65: Redundant "toString()" call in string template
`

After removing the redundant "toString" the problem is gone.

## Your Environment
Ktlint version 0.39.0
MacOs

paul-dingemans:
Given code example below:
```
                fun getDrafts(val draftsIds: List<Long>) {
                    println("draftIds=[${'$'}{draftsIds.toString()}]")
                }
```
Formatting with autocorrect of this code fails on the no-unused-imports rule as the dot-expression node still exists but the toString invocation was already removed.

Resolves #996

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/StringTemplateRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/StringTemplateRule.kt
@@ -6,7 +6,11 @@ import com.pinterest.ktlint.core.ast.ElementType.DOT
 import com.pinterest.ktlint.core.ast.ElementType.DOT_QUALIFIED_EXPRESSION
 import com.pinterest.ktlint.core.ast.ElementType.LITERAL_STRING_TEMPLATE_ENTRY
 import com.pinterest.ktlint.core.ast.ElementType.LONG_STRING_TEMPLATE_ENTRY
+import com.pinterest.ktlint.core.ast.ElementType.LONG_TEMPLATE_ENTRY_END
+import com.pinterest.ktlint.core.ast.ElementType.LONG_TEMPLATE_ENTRY_START
+import com.pinterest.ktlint.core.ast.ElementType.REGULAR_STRING_PART
 import com.pinterest.ktlint.core.ast.ElementType.SUPER_EXPRESSION
+import com.pinterest.ktlint.core.ast.children
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
 import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
 import org.jetbrains.kotlin.psi.KtBlockStringTemplateEntry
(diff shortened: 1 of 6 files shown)


paul-dingemans:
This might fix #1019 although this is not sure as the issue could not be reproduced.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/StringTemplateRule.kt
```diff
@@ -65,9 +69,19 @@ class StringTemplateRule : Rule("string-template") {
             ) {
                 emit(node.treePrev.startOffset + 2, "Redundant curly braces", true)
                 if (autoCorrect) {
-                    // fixme: a proper way would be to downcast to SHORT_STRING_TEMPLATE_ENTRY
-                    (node.firstChildNode as LeafPsiElement).rawReplaceWithText("$") // entry start
-                    (node.lastChildNode as LeafPsiElement).rawReplaceWithText("") // entry end
+                    val leftCurlyBraceNode = node.findChildByType(LONG_TEMPLATE_ENTRY_START)
+                    val rightCurlyBraceNode = node.findChildByType(LONG_TEMPLATE_ENTRY_END)
+                    if (node.children().count() == 3 && leftCurlyBraceNode != null && rightCurlyBraceNode != null) {
```
romtsn, line 74: Could you elaborate on why `count` is 3 exactly? I guess I am missing a bit of context :)

paul-dingemans: It is a bit of defensive programma. The block above is surrounded by the following if-statement:
```
            if (node.text.startsWith("${'$'}{") &&
                node.text.let { it.substring(2, it.length - 1) }.all { it.isPartOfIdentifier() } &&
                (entryExpression is KtNameReferenceExpression || entryExpression is KtThisExpression) &&
                node.treeNext.let { nextSibling ->
                    nextSibling.elementType == CLOSING_QUOTE ||
                        (
                            nextSibling.elementType == LITERAL_STRING_TEMPLATE_ENTRY &&
                                !nextSibling.text[0].isPartOfIdentifier()
                            )
                }
            ) {
```
I was not 100% sure that this statement really covers the situation of a prefix-node, content-node and a suffix-node. In case that also prefix-node, content-node-1, ...,  content-node-n, suffix-node can match this statement then I do think it is not a valid to prefix only the first node with the '$'. So line 74 guards the autocorrect to patterns which can be corrected safely.

Issue #1029, opened 2020-12-23
Placing `// ktlint-disable` at top of file crashes ktlint


AndrewReitz:
Step to Reproduce
Pick any kotlin file add at the very top add `// ktlint-disable`
Run ktlint lint

Following Stacktrace will be printed out

```
com.pinterest.ktlint.core.RuleExecutionException: kotlin.UninitializedPropertyAccessException: lateinit property importSorter has not been initialized
        at com.pinterest.ktlint.core.KtLint$lint$1.invoke(KtLint.kt:93)
        at com.pinterest.ktlint.core.KtLint$lint$1.invoke(KtLint.kt:23)
        at com.pinterest.ktlint.core.KtLint$visitor$2$1.invoke(KtLint.kt:206)
        at com.pinterest.ktlint.core.KtLint$visitor$2$1.invoke(KtLint.kt:23)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:224)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:225)
        at com.pinterest.ktlint.core.KtLint$visitor$2.invoke(KtLint.kt:204)
        at com.pinterest.ktlint.core.KtLint$visitor$2.invoke(KtLint.kt:23)
        at com.pinterest.ktlint.core.KtLint.lint(KtLint.kt:74)
        at org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerAction.lint(LintWorkerAction.kt:71)
        at org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerAction.lintKt(LintWorkerAction.kt:65)
        at org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerAction.access$lintKt(LintWorkerAction.kt:19)
        at org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerAction$execute$2$lintFunc$1.invoke(LintWorkerAction.kt:40)
        at org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerAction$execute$2$lintFunc$1.invoke(LintWorkerAction.kt:19)
        at org.jmailen.gradle.kotlinter.tasks.lint.LintWorkerAction.execute(LintWorkerAction.kt:47)
        at org.gradle.workers.internal.DefaultWorkerServer.execute(DefaultWorkerServer.java:63)
[...]

PR #1038, opened 2020-12-27
Fix ktlint-disable directive on package statement

paul-dingemans:
* Split unit tests for error suppression into multiple tests
* Extract the creation of the SuppressionHints to extension
  methods for readability and reuseability

## Description

According to the documentation the `ktlint-disable` can be appended as an EOL comment to indicate that all rules for this file should be disabled. However that didn't work
```
                package com.pinterest.ktlint // ktlint-disable

                import something.*
```
Adding it as block comment before the package statement resulted in an internal error on the import rule. I have not tried to fix that.

Now it is possible to:
* Disable all rules for this file by adding `// ktlint-disable` after the package statement
* Disable specific rules for this file by adding `// ktlint-disable rule-id` after the package statement
* After disabling a or all rules via the package statement, the rules can be enabled again.

### DIFF
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/SupressedRegionLocator.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/SupressedRegionLocator.kt
@@ -39,20 +39,12 @@ internal fun buildSuppressedRegionsLocator(
     return if (hintsList.isEmpty()) {
         noSuppression
     } else { offset, ruleId, isRoot ->
-        if (isRoot) {
-            val h = hintsList[0]
-            h.range.last == 0 && (
-                h.disabledRules.isEmpty() ||
-                    h.disabledRules.contains(ruleId)
-                )
-        } else {
-            hintsList.any { hint ->
-                (
-                    hint.disabledRules.isEmpty() ||
-                        hint.disabledRules.contains(ruleId)
-                    ) &&
-                    hint.range.contains(offset)
-            }
+        hintsList.any { hint ->
+            (
+                hint.disabledRules.isEmpty() ||
+                    hint.disabledRules.contains(ruleId)
+                ) &&
+                hint.range.contains(offset)
         }
     }
 }
@@ -77,7 +69,8 @@ private fun collect(
             if (text.startsWith("//")) {
                 val commentText = text.removePrefix("//").trim()
                 if (node.isDisabledOnPackageStatement()) {
-                    node.createBlockDisableSuppressionHint(commentText)
+                    parseHintArgs(commentText, "ktlint-disable")
+                        ?.let { hints -> SuppressionHint(IntRange(0, Int.MAX_VALUE), hints) }
                         ?.let { suppressionHint -> open.add(suppressionHint) }
                 } else {
                     node.createLineDisableSupressionHint(commentText)
@@ -124,7 +117,7 @@ private fun PsiElement.createLineDisableSupressionHint(commentText: String): Sup
 
 private fun PsiElement.createBlockDisableSuppressionHint(commentText: String): SuppressionHint? {
     return parseHintArgs(commentText, "ktlint-disable")
-        ?.let { hints -> SuppressionHint(IntRange(node.startOffset, node.startOffset), hints) }
+        ?.let { hints -> SuppressionHint(IntRange(node.startOffset, Int.MAX_VALUE), hints) }
 }
 
 private fun PsiElement.createBlockEnableSuppressionHint(commentText: String, open: java.util.ArrayList<SuppressionHint>): SuppressionHint? {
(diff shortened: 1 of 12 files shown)


ktlint-core/src/test/kotlin/com/pinterest/ktlint/core/ErrorSuppressionTest.kt
```diff
@@ -10,29 +10,71 @@ import org.junit.Test
...
-        class NoWildcardImportsRule : Rule("no-wildcard-imports") {
-            override fun visit(
-                node: ASTNode,
-                autoCorrect: Boolean,
-                emit: (offset: Int, errorMessage: String, corrected: Boolean) -> Unit
-            ) {
-                if (node is LeafPsiElement && node.textMatches("*") && node.isPartOf(ElementType.IMPORT_DIRECTIVE)) {
-                    emit(node.startOffset, "Wildcard import", false)
-                }
+    class NoWildcardImportsRule : Rule("no-wildcard-imports") {
```
Tapchicoma, line 13: Thank you for splitting this tests, much cleaner :+1:

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/SupressedRegionLocator.kt
```diff
@@ -73,32 +76,19 @@ private fun collect(
             val text = node.getText()
             if (text.startsWith("//")) {
                 val commentText = text.removePrefix("//").trim()
-                parseHintArgs(commentText, "ktlint-disable")?.let { args ->
-                    val lineStart = (
-                        node.prevLeaf { it is PsiWhiteSpace && it.textContains('\n') } as
-                            PsiWhiteSpace?
-                        )?.let { it.node.startOffset + it.text.lastIndexOf('\n') + 1 } ?: 0
-                    result.add(SuppressionHint(IntRange(lineStart, node.startOffset), HashSet(args)))
+                if (node.isDisabledOnPackageStatement()) {
```
Tapchicoma, line 79: KtLint readme [on this case](https://github.com/pinterest/ktlint#how-do-i-suppress-an-error-for-a-lineblockfile) is not very clear.
```
import package.* // ktlint-disable
```
Should disable any rule check on this particular line.

It doesn't make sense to disable all the rules for the whole file - better to either not pass it to KtLint or exclude it via GLOB pattern. To disable all the rules for the region, better to use:
```kotlin
/* ktlint-disable */
<some code>
/* ktlint-enable */
```

So, for the case `<some code> // ktlint-disable` please use `createLineDisableSupressionHint` and pass `emptySet()` as `disabledRules`.

Tapchicoma (changes requested):

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/SupressedRegionLocator.kt
paul-dingemans: I really like puzzles. But this review remark really confused me at first. I have the feeling that we were discussing different, albeit related, topics. Let's try to unravel this.

> KtLint readme [on this case](https://github.com/pinterest/ktlint#how-do-i-suppress-an-error-for-a-lineblockfile) is not very clear.
> 
> ```
> import package.* // ktlint-disable
> ```
> 
> Should disable any rule check on this particular line.
This code example is confusing for different reasons:
* There is no apparent reason why this statement should be excluded by ktlint as there is no rule violated.
* The word `package` after `import` took precedence for me. So I read this statement as being the package statement:
```
package com.example.mypackage // ktlint-disable
```
I have modified this.

> It doesn't make sense to disable all the rules for the whole file - better to either not pass it to KtLint or exclude it via GLOB pattern. 
It is only possible to follow this approach when you know which files need to be excluded from processing by ktlint. In bigger code bases with which you are not familiar you may just not know which files would need to be excluded. When you are able to specify it explicitly in the file itself, you can also document the reasoning for this. This approach allows to lint/format the entire code base without having to know the files which need to be excluded.

> So, for the case `<some code> // ktlint-disable` please use `createLineDisableSupressionHint` and pass `emptySet()` as `disabledRules`.
I have removed the code for the `ktlint-disable` directive on the *package* statement as I no longer needed it after fixing some other bugs.

I found out that certain bugs are caused by the fact that some rules perform a kind of initialization as soon as the rootNode is visited. If for some reasons, for example disabling a rule, the rootNode is not visited then the rule would not have been initialized correctly. I have split initialization and visiting.

Next problem was that the IndentationRule modifies the ASTNode hierarchy by inserting additional node. Disabling the indentation rule was not always recognized by the SupressedRegionLocator. I have fixed and simplified this locator.

The changes can best be reviewed per commit.

Tapchicoma (commented):
Thank you for your effort, but it seems too much changes trying to fix #1029 

This issue could be fixed by changing [SuppressedRegionLocator.kt#74](https://github.com/pinterest/ktlint/blob/adb930d499b687965d32476d2f881dcc5060a54b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/internal/SupressedRegionLocator.kt#L74) into:
```kotlin
        if (node is PsiComment && node.prevSibling != null) {
```

that will ignore all lines containing only `// ktlint-disabled` comment.

pauldingemans:
As stated in my previous comment, my changes also fix other problems. Please take that into account as well.

paul-dingemans:
PR is too far out of sync with current master.

PR #1041, opened 2020-12-29
Don't remove the equals sign for a default argument (`no-line-break-before-assignment`)

### LINKED ISSUE #1039: Default argument equals removed when at line start
When running with autoformat, the equals sign for a default argument is removed if it is at the line start (it's because of codegen).

### Reproducer
With 0.40.0, Kotlin 1.4.21, 
```kotlin
public fun decodeRaw2(littleEndian: Boolean?
= null): Int = 3
```
is turned into this:
```kotlin
public fun decodeRaw2(
    littleEndian: Boolean?
    null
): Int = 3
```
which doesn't compile.

t-kameyama:
## Description
Fixes #1039

## Checklist
- [x] tests are added


romtsn:
Squashed as the integration test is not really needed (or shouldn't be in there at the very least). Thanks!

PR #1042, opened 2020-12-29
Upgrade ec4j to latest version

### LINKED ISSUE #739: Syntactically invalid .editorconfig silently ignored
I am trying to setup my .editorconfig and it seems it silently ignores some internal errors and it makes the .editorconfig hard to debug.

```
# http://editorconfig.org
droot = true
[*]
indent_style = space
indent_size = 4
end_of_line = lff
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
[*.{kt,kts}]
disabled_rules = disabled_rules = no-wildcard-imports,import-ordering,no-consecutive-blank-lines,colon-spacing,curly-spacing,range-spacing,no-var,dump,chain-wrapping,comment-spacing,filename,final-newline,indent,max-line-length,modifier-order,no-blank-line-before-rbrace,no-line-break-after-else,no-line-break-before-assignment,no-semi,no-trailing-spaces,parameter-list-wrapping,colon-spacing,comma-spacing,curly-spacing,op-spacing
```

For instance, see the invalid `end_of_line` parameter - IntelliJ will highlight the incorrectness, but ktlint just skips it. Also for some reason, the `disabled_rules` is also ignored with no error as I keep getting:

```
/src/main/kotlin/net/goout/geoipinfo/GeoIpInfo.kt:10:1: Wildcard import (cannot be auto-corrected)
```

Also changing the `root = true` to invalid `droot = true` has no effect and yields not even a warning.

On the other hand, changing indent_size to 5 correctly yields (meaning the file is correctly loaded):

```
src/main/kotlin/net/goout/geoipinfo/Geoipinfo.kt:52:1: Unexpected indentation (4) (it should be 5) (cannot be auto-corrected)
```
 
Using `--debug` doesn't help much, it just says:

```
[DEBUG] Discovered .editorconfig (/Users/knyttlwork/Workspace/GoOut/GeoIpInfo) {charset=utf-8, end_of_line=lff, indent_size=4, indent_style=space, insert_final_newline=true, tab_width=4, trim_trailing_whitespace=true}
[DEBUG] Discovered .editorconfig (/Users/knyttlwork/Workspace/GoOut/GeoIpInfo) {charset=utf-8, end_of_line=lff, indent_size=4, indent_style=space, insert_final_newline=true, tab_width=4, trim_trailing_whitespace=true}
```
(why does the message appear twice?)

But it doesn't complain about anything invalid.

Version:
```
$ ktlint --version
0.36.0
```

goodhoko:
ec4j 0.3.0 removes the length limit of key, value and section name in editor config files.

Fixes #739.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ ext.deps = [
     'compiler': "org.jetbrains.kotlin:kotlin-compiler-embeddable:${versions.kotlin}"
   ],
   'klob'          : 'com.github.shyiko.klob:klob:0.2.1',
-  ec4j            : 'org.ec4j.core:ec4j-core:0.2.2',
+  ec4j            : 'org.ec4j.core:ec4j-core:0.3.0',
   'picocli'       : 'info.picocli:picocli:3.9.6',
   // Testing libraries
   'junit'         : 'junit:junit:4.13.1',


romtsn:
merging as this is the only diff in ec4j between version 0.2.2 and 0.3.0, so should not bring any problems. Thanks for following ec4j and contributing!

goodhoko:
@romtsn Are there any estimates on when this fix might get released? Is there anything we could help with to push the relrase forward? .)

PR #1054, opened 2021-01-12
Fix false positives when declaration has tail comments (`spacing-between-declarations-with-comments`)

### LINKED ISSUE #1053: [0.40.0] experimental:spacing-between-declarations-with-comments false positives with `//` comments on class properties, variables and enum values
## Expected Behavior
I expect ktlint not to fail nor modify the code below when autoformatting. According to the issue [linked in ktlint check](https://github.com/pinterest/ktlint/blob/09e27487ed3dd1c0c568da4796e30f744f71c809/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/SpacingBetweenDeclarationsWithCommentsRule.kt#L16), while Intellj indeed adds spacing between methods it doesn't between class properties, enum values, nor method variables.
Basically I expect `spacing-between-declarations-with-comments` not to report issues with a code that doesn't get autocorrected by IntelliJ.

## Observed Behavior
ktlint fails with 
```
SampleClass.kt:6:36: Declarations and declarations with comments should have an empty space between.
```
on every `"// false positive"` comment (below)

\+ when executed with `-F` it adds spaces in places where intellij doesn't
|before|after formatting with ktlint|
|---|---|
|image|image|

## Steps to Reproduce
Run ktlint on following code:
```kotlin
package com.doesnt.work.library

class SampleClass {

    private val public = "ok" // ok
    private val private = "not_ok" // false positive
    private val halfPublicHalfPrivate = "not_ok" // false positive

    /**
     * Doc 1
     */
    fun one() = 1

    /**
     * Doc 2
     */
    fun two() = 2
    fun three() {
        val public = "ok" // ok
        val private = "not_ok" // false positive
    }
}

enum class SampleEnum {
    One, // ok
    Two, // false positive
    Three, // false positive
}
```

## Your Environment
```groovy
configurations {
    ktlint
}

dependencies {
    ktlint "com.pinterest:ktlint:0.40.0"
}

task ktlint(type: JavaExec, group: "verification") {
    description = "Check Kotlin code style."
    classpath = configurations.ktlint
    main = "com.pinterest.ktlint.Main"
    args "src/**/*.kt", "--experimental"
}
```

* Version of ktlint used: 0.39.0 (works), 0.40.0 (doesn't work)
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): manually added Gradle tasks pasted above
* Version of Gradle used (if applicable): 6.8
* Link to your project (if it's a public repository):

t-kameyama:
## Description
Fixes #1053

## Checklist

- [x] tests are added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Fix indentation for function types after a newline (`indent`) ([#918](https://github.com/pinterest/ktlint/issues/918))
 - Don't remove the equals sign for a default argument (`no-line-break-before-assignment`) ([#1039](https://github.com/pinterest/ktlint/issues/1039))
 - Fix internal error in `no-unused-imports` ([#1040](https://github.com/pinterest/ktlint/issues/1040))
+- Fix false positives when declaration has tail comments (`spacing-between-declarations-with-comments`) ([#1053](https://github.com/pinterest/ktlint/issues/1053))
 
 ### Changed
 - Update Gradle shadow plugin to `6.1.0` version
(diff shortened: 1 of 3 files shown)


Tapchicoma (approved):
Thank you for fixing it :+1:

PR #1058, opened 2021-01-15
Fix formatting with comments (`colon-spacing`)

### LINKED ISSUE #1057: multiline expressions transformed to single line inside comment
## Expected Behavior

```
var x // check it out
: String = "hi"
```
should be transformed to
```
var x: String = "hi" // check it out
```

## Observed Behavior

```
var x // check it out
: String = "hi"
```
outputs to
```
var x // check it out: String = "hi"
```

## Your Environment

* Name and version: custom gradle task and command line

t-kameyama:
## Description
Fixes #1057

## Checklist

- [x] tests are added

### DIFF
--- a/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundColonRuleTest.kt
+++ b/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundColonRuleTest.kt
@@ -248,14 +248,48 @@ class SpacingAroundColonRuleTest {
             class X :
                 Y,
                 Z
-            class A // comment
-                : B
-            class A /*
+            class A : // comment
+                B
+            class A : /*
 
             */
-                : B
+                B
             val xmlFormatter:
                 String = ""
+            """.trimIndent()
+        )
+    }
+
+    // https://github.com/pinterest/ktlint/issues/1057
+    @Test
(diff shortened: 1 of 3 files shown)


ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/SpacingAroundColonRuleTest.kt
```diff
@@ -248,15 +248,49 @@ class SpacingAroundColonRuleTest {
...
+            SpacingAroundColonRule().format(
+                """
+                val x // comment
+                    : Int = 1
+                """.trimIndent()
+            )
+        ).isEqualTo(
+            """
+                val x: // comment
+                    Int = 1
```
Tapchicoma, line 292: Not sure it is the best way to format :thinking: 
Property type should be, imho, on the same line with property name. Probably formatted code could look like this:
```kotlin
val x: Int = // comment
    1
```

or like this:
```kotlin
// comment
val x: Int = 1
```

What do you think?

t-kameyama: > ```kotlin
> val x: Int = // comment
>     1
> ```
I think this is better.
Fixed in 9460b58.

Tapchicoma (approved):
LGTM!

PR #1093, opened 2021-03-03
Allow to add/replace loaded values from '.editorconfig' files

### LINKED ISSUE #1016: Expose configurability of ImportOrderingRule
## Expected Behavior
Dektet or other clients can configure `ImportOrderingRule` by invoking `UserDataHolder.putUserData(Key<T>, T)`

## Observed Behavior
https://github.com/pinterest/ktlint/pull/935 defines `PatternEntry` as the internal type, which makes it impossible for Detekt to generate the value to invoke `UserDataHolder.putUserData()`. Therefore,
Detekt now either need to use a ugly hack (https://github.com/detekt/detekt/pull/3281) or needs to bypass the `internal` visibility of `PatternEntry`.

ImportOrderingRule is now in `FeatureInAlphaState`, so I expect API or implementation to change in the future. But I would like to use this issue to discuss the way moving forward.

## Your Environment

Tapchicoma:
## Description

Add new parameter to add/replace loaded values from `.editorconfig` files.

Fixes #1016 

## Checklist

- [x] tests are added

### DIFF
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
@@ -2,10 +2,12 @@ package com.pinterest.ktlint.core
 
 import com.pinterest.ktlint.core.api.EditorConfigProperties
 import com.pinterest.ktlint.core.api.FeatureInAlphaState
+import com.pinterest.ktlint.core.api.UsesEditorConfigProperties
 import com.pinterest.ktlint.core.ast.visit
 import com.pinterest.ktlint.core.internal.EditorConfigGenerator
 import com.pinterest.ktlint.core.internal.EditorConfigLoader
 import com.pinterest.ktlint.core.internal.EditorConfigLoader.Companion.convertToRawValues
+import com.pinterest.ktlint.core.internal.EditorConfigOverridesMap
 import com.pinterest.ktlint.core.internal.KotlinPsiFileFactory
 import com.pinterest.ktlint.core.internal.LineAndColumn
 import com.pinterest.ktlint.core.internal.SuppressionLocator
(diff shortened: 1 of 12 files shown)


ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
```diff
@@ -51,10 +54,18 @@ public object KtLint {
...
+     * For possible keys check related [Rule]s that implements [UsesEditorConfigProperties] interface.
+     *
+     * For values use `PropertyType.PropertyValue.valid("override", <expected type>)` approach.
+     * It is also possible to set value into "unset" state by using [PropertyType.PropertyValue.UNSET].
+     *
      * @param isInvokedFromCli **For internal use only**: indicates that linting was invoked from KtLint CLI tool.
      * Enables some internals workarounds for Kotlin Compiler initialization.
      * Usually you don't need to use it and most probably it will be removed in one of next versions.
      */
+    @OptIn(FeatureInAlphaState::class)
```
Tapchicoma, line 68: I am thinking about introducing `ExperimentalParams` that will have `@FeatureInAlphaState` annotation.

`Params` will not have any experimental properties, but actually will create `ExperimentalParams` with default values for experimental properties.

Tapchicoma:
@chao2zhang could you take a look on this PR and check if exposed api looks good and useful for your case?

chao2zhang:
Thank you for mentioning me. Based on my static code reading, escalating visibility of `ImportOrderingRule.ideaImportsLayoutProperty` and `FinalNewlineRule.insertNewLineProperty` should help eliminate hacks in Detekt.

Later this week, I can check the compatibility by applying this PR to Detekt. But this PR should be good to merge from my perspective.

romtsn (commented):
LGTM 👍  the only thing is that detekt does not use `Params` class, therefore not sure how useful is this addition for them (they inject those directly into the nodes via `putUserData`)

romtsn (approved):

Issue #1102, opened 2021-03-11
Support SARIF format


chao2zhang:
Github has the capabilitiy to integrate third-party static analysis tools. The official documentation can be found [here](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning#about-third-party-code-scanning-tools) which requires static analysis tools to produce [SARIF] format.

This issue is already picked up in 
- Android Lint - See https://issuetracker.google.com/issues/167593732
- Detekt - https://github.com/detekt/detekt/issues/3045
- Checkstyle - https://github.com/checkstyle/checkstyle/issues/8905 This is working in progress for me.

I would like to seek your opinion on supporting SARIF format as first-class for ktlint. (We already have json format https://github.com/pinterest/ktlint/tree/master/ktlint-reporter-json).
If approved, I would be happy to contribute as well.

Tapchicoma:
So it is basically adding another reporter that will produce report  in `SARIF` format?

chao2zhang:
That's correct.

chao2zhang:
Because [SARIF is officially supported by Github code action](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/sarif-support-for-code-scanning), we could show the violations inline (https://github.com/chao2zhang/detekt/pull/9/files)
![image](https://user-images.githubusercontent.com/1175468/111238765-ff5fad80-85b4-11eb-9800-a7222ad8bf40.png)

Tapchicoma:
Then it makes sense to add such reporter. Feel free to contribute :+1:

Tapchicoma:
@chao2zhang do you plan to further improve SARIF format support or it is fine as it is?

chao2zhang:
We can close this issue since we have basic support. Further improvements can be done in separate issues:

- [ ]  SarifSchema210.runs[0].tool.driver.rules is empty for now. Ideally, this should be a list of rules that were active in the current run. I will address this in a follow-up PR since this requires refactoring.
- [ ]  I have not yet added other information like autoCorrected or autoCorrectible in the output.
- [ ]  Potentially we can also add github workflow setup in the documentation. There are other community projects that does this so this is optional to me.

akuleshov7:
@chao2zhang have you tested ktlint reports in GitHub? It looks like the calculation of URI paths is incorrect
Any examples, where it works?

akuleshov7:
[...]

PR #1113, opened 2021-03-18
Add basic SARIF support

chao2zhang:
## Description
This addresses #1102. SARIF website: https://sarifweb.azurewebsites.net/
- This makes ktlint to depend on `https://github.com/detekt/sarif4k`, which uses kotlinx-serialization to generate the JSON. Schema makes the code safer and less error-prone. If we want to avoid additional dependency, I could change the code to yield JSON using basic string formatting and concatenation. 
- The level in SARIF report is currently set to ERROR. I believe this fits with the ktlint's strategy of no configuration.

## Checklist
- [X] tests are added

## Follow-up
- [ ] `SarifSchema210.runs[0].tool.driver.rules` is empty for now. Ideally, this should be a list of rules that were active in the current run. I will address this in a follow-up PR since this requires refactoring.
- [ ] I have not yet added other information like autoCorrected or autoCorrectible in the output.
- [ ] Potentially we can also add github workflow setup in the documentation. There are other community projects that does this so this is optional to me.

## Test
- I ran `klint --reporter=sarif` on the `ktlint` itself. I also validated the output on `https://sarifweb.azurewebsites.net/Validation` to make sure that there is no basic warnings.


ktlint-reporter-sarif/src/main/kotlin/com/pinterest/ktlint/reporter/sarif/SarifReporter.kt
```diff
@@ -0,0 +1,97 @@
...
+        val sarifSchema210 = SarifSchema210(
+            schema = "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
+            version = Version.The210,
+            runs = listOf(
+                Run(
+                    tool = Tool(
+                        driver = ToolComponent(
+                            downloadURI = "https://github.com/pinterest/ktlint/releases/tag/$version",
+                            fullName = "ktlint",
+                            guid = "7e96e861-6db7-44ea-a1e4-56bb4554ffe9",
```
romtsn, line 78: hm, where is this coming from?

ktlint-reporter-sarif/src/main/kotlin/com/pinterest/ktlint/reporter/sarif/SarifReporter.kt
```diff
@@ -0,0 +1,97 @@
...
+                        physicalLocation = PhysicalLocation(
+                            region = Region(
+                                startLine = err.line.toLong(),
+                                startColumn = err.col.toLong()
+                            ),
+                            artifactLocation = workingDirectory?.let { workingDirectory ->
+                                ArtifactLocation(
+                                    uri = File(file).relativeTo(workingDirectory).path,
+                                    uriBaseID = SRCROOT
+
```
romtsn, line 54:
```suggestion
```

romtsn (approved):
Don't know a thing about sarif, but I guess you know what you're doing 👍

ktlint-reporter-sarif/src/main/kotlin/com/pinterest/ktlint/reporter/sarif/SarifReporter.kt
chao2zhang: guid is a unique identifier for the component: https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317539. I generated a random one for the time being.

It is optional, I can remove this.

shashachu:
@chao2zhang apologies for the delay and thanks for the contribution. the CI build is now fixed so if you rebase I can go ahead and merge your change.

chao2zhang:
Updated the PR, it looks like the PR is pending approval to kick off github actions

PR #1117, opened 2021-03-21
Detect filename is not compliant with PascalCase convention

bedla:
Hi,
I found that FilenameRule does not check PascalCase convention mentioned in https://kotlinlang.org/docs/coding-conventions.html#source-file-names
What do you think?
Thx
Ivos


shashachu:
hi @bedla thanks for the PR. Reading the Kotlin docs, my take is that filenames should *always* conform to Pascal case, regardless of the contents. Is that not also your understanding?

bedla:
Yep, from my understanding PascalCase is only viable name of the file.

I have rebased to current master.

bedla:
I have update PR to current master

paul-dingemans:
The code seems unnecessary complex to me. Basically, I would expect it do the following:
1. Determine the number of top level declarations
2. Determine the name of the first class
3. If one top level declaration was found and the first class name is not null than the name of file should be identical to the name of the class followed by ".kt". Otherwise check if the filename adheres to the the PascalCase convention.

paul-dingemans:
@bedla Do you plan to follow up on my remarks on this PR?

bedla:
Yes, I will. Sorry for being late, I had some rush in work.

bedla:
Just pushed requested changes. I have rewritten original logic to make it more explicit from my point of view. Pls, take a look. Thx

bedla:
Hmm, interesting "edge cases".

- `CommandLineExt` & `ByteArrayExt` - actually I like more this convention with `DescriptiveName+Ext.kt` ... could we extends original Kotlin's convention with exception to `Ext` suffix in this case? It seems reasonable to me.
- `ToolchainForTests` - I agree with rule that filename should be `Project.kt`, but on the other hand there should be possible to add exceptions from the rule.

What do you thing?

paul-dingemans:
> * `CommandLineExt` & `ByteArrayExt` - actually I like more this convention with `DescriptiveName+Ext.kt` ... could we extends original Kotlin's convention with exception to `Ext` suffix in this case? It seems reasonable to me.

> * `ToolchainForTests` - I agree with rule that filename should be `Project.kt`, but on the other hand there should be possible to add exceptions from the rule.

I think we should handle both cases similar. If a file does not contain any class and has filename ending with "Ext" or "Extension" and is following the PascalCase convention then there should be no violation about the file name. Would that be feasible?

bedla:
Ok, I will take a look at it. I have also created https://youtrack.jetbrains.com/issue/KT-51919 to know Kotlin creator's opinion on this edge case (I do not know if they will join discussion, but why not try :) ).

bedla:
Mr. Elizarov responded in Youtrack issue. I think that he is right.

But it means that should remove checks around `allOfSameReceiver`. And also his response invalidate what was mentioned in original code https://github.com/pinterest/ktlint/blob/0.45.2/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/FilenameRule.kt#L67 ... If I get it correctly, it is because of situation when we want to name file with same name like Receiver name, when we have only extensions in the file. But this one is not explicitly noted in lang-docs.

What do you think?
Do you know why exactly is that note at line 67?

paul-dingemans:
It looks like that https://github.com/pinterest/ktlint/blob/0.45.2/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/FilenameRule.kt#L67 is just the example of a file in which extension functions have the same receiver. There can be a good reason to group such functions together in one file, but I agree that we should not force that the file is named after the receiver. It would be enough to enforce the upper PascalCase convention here.

When applying above, a project can still use a filename like `SomeReceiverExt.kt` as it adheres to the upper PascalCase convention. But the other way around, a file which only contain extension functions on the `Foo` receiver does not necessarily to be named as `Foo.kt` but could be named as `RedundantBar.kt`.

paul-dingemans:
@bedla Do you have some time to finish up the issue? I expect a new release to be build in a couple of weeks.

bedla:
ahh, sorry I missed your previous comment.

bedla:
I have removed that all-top-level receivers exception, but it still fails on:

```
C:\Users\ivo.smid\IdeaProjects\ktlint\ktlint\src\main\kotlin\com\pinterest\ktlint\internal\ByteArrayExt.kt:1:1: Extension property hex should be declared in a file named hex.kt (cannot be auto-corrected) (filename)
C:\Users\ivo.smid\IdeaProjects\ktlint\ktlint\src\main\kotlin\com\pinterest\ktlint\internal\CommandLineExt.kt:1:1: Extension function printHelpOrVersionUsage should be declared in a file named printHelpOrVersionUsage.kt (cannot be auto-corrected) (filename)
```

So if I get it correctly, I should remove also this sub-rule for Extension functions & Extension property? What do you think?

vanniktech:
Is it desired that with this change, the following file:

```kotlin
package com.vanniktech.feature.boardmoney.data

import app.cash.sqldelight.adapter.primitive.IntColumnAdapter
import app.cash.sqldelight.db.SqlDriver
import com.vanniktech.boardmoney.QueryWrapper

internal fun createQueryWrapper(driver: SqlDriver) = QueryWrapper(
  driver = driver,
  boardMoneyPlayerAdapter = BoardMoneyPlayer.Adapter(
    colorAdapter = IntColumnAdapter,
  ),
)
```

gets linted with the following: 

> /Users/niklas/dev/GitHub/vanniktech/feature-board-money/src/commonMain/kotlin/com/vanniktech/feature/boardmoney/data/Db.kt:1:1: File 'Db.kt' contains a single top level declaration and should be named 'CreateQueryWrapper.kt' (cannot be auto-corrected) (filename)

That does not seem right to me.

vanniktech:
> If a Kotlin file contains a single class or interface (potentially with related top-level declarations), its name should be the same as the name of the class, with the .kt extension appended. It applies to all types of classes and interfaces. If a file contains multiple classes, or only top-level declarations, choose a name describing what the file contains, and name the file accordingly. Use [upper camel case](https://en.wikipedia.org/wiki/Camel_case) with an uppercase first letter (also known as Pascal case), for example, ProcessDeclarations.kt.

I'd agree for a Class / Interface, but not for a given function (as in my case).

PR #1129, opened 2021-03-30
Fix unused import when import is duplicated

analog-alex:
## Description

**ktlint** detects unused imports, such as:

```
import lib
import lib2 // <- ktlint unused import rule is violated

fun main() {
	lib.function();
}
```

However, when an import is duplicated -- and the duplicate import line is thus an unused import -- **ktlint** emits no rule violation.

Screen Shot 2021-03-30 at 2 17 05 PM

**IntelliJ** treats it as a duplicate import ☝️ 
 
E.g:

```
import lib
import lib // unused, but no violation

fun main() {
	lib.function();
```

This PR fixes that, emitting a rule violation when two imports have the same import path.

## Checklist

- [x] tests are added


shashachu:
@analog-alex our dependency verification errors are fixed now. you should be able to revert the changes to `verification-metadata.xml` and rebase and your builds should pass.

analog-alex:
> @analog-alex our dependency verification errors are fixed now. you should be able to revert the changes to `verification-metadata.xml` and rebase and your builds should pass.

Reverted!

Issue #659, opened 2019-11-26
Ignore line length in string literals


pkubowicz:
Let's have `.editorconfig'
```
[*.{kt,kts}]
max_line_length=40
```
and Kotlin file:
```
val string1 =
        "a very very very very very long string literal"

val string2 = "a very very very very very long string literal"
```

Ktlint 0.35.0 rejects both declarations:
```
LongLines.kt:2:1: Exceeded max line length (40) (cannot be auto-corrected) (max-line-length)
LongLines.kt:4:1: Exceeded max line length (40) (cannot be auto-corrected) (max-line-length)
```

I think at least line 2 should be accepted - it contains only whitespace and a string literal.

Tapchicoma:
Why it should be accepted? :thinking: 

You could split string literal using `+` sign. If you still want to have it long, add [disable](https://github.com/pinterest/ktlint#how-do-i-suppress-an-error-for-a-lineblockfile) block around it.

pkubowicz:
Since comments and raw strings are already not taken into account when checking line length.

Splitting string literal is risky - this string may be the expected value in a test and you may miss something if the string is contained in 2 lines:
```
assertThat(string).isEqualTo(
    "very very long" +
    "string but I missed space between lines"
)
```

Disabling blocks are very invasive as you need to open and close them, so you get 1:2 signal to noise ratio: 2 lines to satisfy ktlint in 1 line of code.

Tapchicoma:
Another way - would be support disabling particular rules via `@Suppress` annotation.

pkubowicz:
Supporting `@Suppres` would be nice in this case.

ansman:
I agree, string in exceptions, deprecations etc should be allowed to be of any length as it's very inconvenient to split them.

k1dbl4ck:
Would like to add that back ticked function names in tests should also be handled:

```
@Test
 fun `Should return something when passing something and something and the other thing is in this or that state`()` {...}
```

VishalNehra:
+1

florianmski:
Just adding my two cents for this issue that we hit in our current project.
We found out as mentioned here https://github.com/pinterest/ktlint/issues/659#issuecomment-664319682 that this is not practical for tests names and sometimes you do really need that extra 20 chars to have a clear test name.
I first set the `max_line_length` to `off` but it still seemed to auto format some of the code according to some hardcoded limit somewhere so I came up with the following which has been working great for the last few months or so, hope that helps:

[...]

PR #1149, opened 2021-05-19
Add ignoreSingleLineRawStringProperty

mideuger:
## Description

Allow single line raw strings to be ignored during MaxLineLengthRule execution.

Since multi-line raw strings are already ignored by default and that it is common to have long string literals, this may be a good compromise between disabling the rule and using multi-line raw strings.

## Checklist

- [x] tests are added


shashachu:
I do wonder if this even needs a configuration option. Maybe we should just ignore all raw strings, regardless of number of lines? As a principle for ktlint we try to have a minimal set of configuration options (although we've certainly violated that in the past.)

paul-dingemans:
In https://github.com/pinterest/ktlint/issues/659#issuecomment-844148789 it was mentioned that #1149 and #1150 are alternative solutions for the issue. As @shashachu mentioned, and I agree with her, we should not add a ktlint option for this. So closing down this alternative without merging.

PR #1150, opened 2021-05-19
Add ktlint_ignore_string option

mideuger:
## Description

Add ktlint_ignore_string option to max-line-length.
Allow ignoring strings length when checking max line length.

## Checklist

- [x] tests are added

### DIFF
--- a/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/MaxLineLengthRuleTest.kt
+++ b/ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/MaxLineLengthRuleTest.kt
@@ -4,6 +4,7 @@ import com.pinterest.ktlint.core.LintError
 import com.pinterest.ktlint.core.api.DefaultEditorConfigProperties.maxLineLengthProperty
 import com.pinterest.ktlint.core.api.FeatureInAlphaState
 import com.pinterest.ktlint.ruleset.standard.MaxLineLengthRule.Companion.ignoreBackTickedIdentifierProperty
+import com.pinterest.ktlint.ruleset.standard.MaxLineLengthRule.Companion.ignoreStringProperty
 import com.pinterest.ktlint.test.EditorConfigOverride
 import com.pinterest.ktlint.test.diffFileLint
 import com.pinterest.ktlint.test.lint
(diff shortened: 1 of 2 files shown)


ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/MaxLineLengthRuleTest.kt
```diff
@@ -90,6 +90,41 @@ class MaxLineLengthRuleTest {
...
+                """
+                fun main() {
+                    val longString = "this is a very very very very long long string on a single line"
+                }
+                """.trimIndent(),
+                userData = mapOf("max_line_length" to "40")
+            )
+        ).isEmpty()
+    }
+
```
paul-dingemans, line 109: Please add a test in which both properties are set. Experiment with a test with following scenario:
```
val `backTickedProperty` = "some very very very very long string on a single line"
```

paul-dingemans:
@mideuger Do you plan to work on this? If so, please merge master first in your local branch.

mideuger:
@paul-dingemans Sure. Let me know if there is something else I should do.

I also wonder, what is your opinion on the `totalLengthStringElements` function as it does not perfectly mimick `totalLengthBacktickedElements` ? I'm not sure if something like `elements.filter { it.isPartOfString() }.sumBy { it.textLength }` would be better.

paul-dingemans:
> @paul-dingemans Sure. Let me know if there is something else I should do.
> 
> I also wonder, what is your opinion on the `totalLengthStringElements` function as it does not perfectly mimick `totalLengthBacktickedElements` ? I'm not sure if something like `elements.filter { it.isPartOfString() }.sumBy { it.textLength }` would be better.

To be honest, it is quite hard to say what is best. Best thing what you can do is write unit tests which proof that the current solutions has drawbacks. After all, it only has effect in case the last element on the line is a string. So I guess that a line containing a long string followed by an end of line comment will not be properly detected. Next thing would be trying out your alternative solution. Also try to write tests for that solution and see if you can break it.

mideuger:
Just added a test showing that strings are ignored only when they are the last token on the line.
How would you like me to publish the other implementation and the corresponding tests ? Should I just open a new PR ?
Currently I have not much idea about what should be tested with the other implementation (except showing the difference with the actual implementation, ie. all strings would be ignored, everywhere).

paul-dingemans:
I am getting more and more doubts about the functionality of ignoring the content of strings for the maxline length rule. See examples and reasoning below:

```
// Suppose that max line length ends at the hashtag sign #

fun fooBar(
    // Next statement is longer than max line length. It
    // can, but should not be wrapped on multiple lines.
    // So this line should be allowed to exceed the max
    // line length.
    // End of line at hashtag                            #
    `baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar` : String
    /* Do not force this to wrap as follows:
    `baaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar`
        : String
     */
) {
    // Next statement is longer than max line length. It
    // can not be wrapped on multiple lines as. If a
    // linebreak would be added after the var keyword,
    // there would still be some additional indentation
    // be required.
    // So this line should be allowed to exceed the max
    // End of line at hashtag                            #
    var `foooooooooooooooooooooooooooooooooooooooooooooooo` = "fooooooo"

    // Next statement is longer than max line length. It
    // can be wrapped on multiple lines. So this line
    // should not be allowed to exceed the max line
    // length.
    // End of line at hashtag                            #
    `foooooooooooooooooooooooooooooooooooooooooooooooo` = "fooooooo"
    /* It should be changed manually to:
    `foooooooooooooooooooooooooooooooooooooooooooooooo` =
        "fooooooo"
     */

    // Next statement is longer than max line length. It
    // can be wrapped on multiple lines. So this line
    // should not be allowed to exceed the max line
    // length.
    // End of line at hashtag                            #
    var foo1 = "fooooooooooooooooooooooooooooooooooooooooooooooo"
    /* It should be changed manually to:
    var foo1 =
        "fooooooooooooooooooooooooooooooooooooooooooooooo"

     */

    // Next statement is longer than max line length. It
    // can (technically, but it feels artificial) be
    // wrapped on multiple lines. So this line should
    // not be allowed to exceed the max line length.
    // End of line at hashtag                            #
    var foo2 = "fooooooooooooooooooooooooooooooooooooooooooooooooooooooo"
    /* It should be changed manually to below (note that
    strings starting at position 0 are ignored by indent
    rule):
    var foo2 =
"fooooooooooooooooooooooooooooooooooooooooooooooooooooooo"
     */

    // Next statement is longer than max line length. It
    // can (technically, but it feels artificial) be
    // wrapped on multiple lines. So this line should
    // not be allowed to exceed the max line length.
    // End of line at hashtag                            #
    var foo3 = "fooooooooooooooooooooooooooooooooooooooooooooooooooooooooo"
    /* It should be changed manually to below (note that
    strings starting at position 0 are ignored by indent
    rule):
    var foo3 =
"""
fooooooooooooooooooooooooooooooooooooooooooooooooooooooooo
""".trimIndent()
     */

    // Next statement is longer than max line length. It
    // can not be wrapped on multiple lines as the length
    // of the string content is longer than the max line
    // length. So this line should be allowed to exceed
    // the max line length.
    // End of line at hashtag                            #
    var foo4 = "foooooooooooooooooooooooooooooooooooooooooooooooooooooooooo"

    // Next statement (including the EOL comment) is 
    // longer than max line length. It can and should be
    // rewritten by moving the comment to a separate line.
    // So this line should not be allowed to exceed the
    // max line length.
    // End of line at hashtag                            #
    var foo5 = "foooooooooooooooooooooooooooooooooooooooo" // Bar
}
```

I can not (yet) figure out how to express the conditions in which the rule should ignore the content of long strings when deciding whether a violations should or should not be ignored. This means that we also can not explain it to our users, and as of that will result in future bug reports and modifications. So I think we just need to clarify those conditions first.

The use case for the backticked identifiers was originally to support unit tests with long test names between backticks:
```
@Test
fun `Given that we have a test description that is longer than the max line length, then do not report a violation`() {
    // ....
}    
```
For communication purposes, it might be better if we apply this exception to function name identifiers only.

paul-dingemans:
Hi @mideuger,
Do you want to proceed with this PR, or do you agree that we seem to be stuck?

mideuger:
Hi @paul-dingemans , this is indeed more tricky than I first thought. I agree that we can close this PR until we find a proper solution.

PR #1180, opened 2021-07-01
Fix false positive with elvis operator and comment (`chain-wrapping`)

### LINKED ISSUE #1055: False positive for chain-wrapping with elvis operator and comment
## Expected Behavior

In IntelliJ, there is a quick fix to shorten the following code using the `?:` operator:

```kotlin
val abc = getAbc()
if (abc == null) {
    // Comment
    return Xyz()
}
```

into

```kotlin
val abc = getAbc()
    ?: // Comment
    return Xyz()
```

As this was how IntelliJ transformed it, it should be considered good style.

## Observed Behavior

Instead klint then complains about this with `chain-wrapping    myfile.kt:130    Line must not end with "?:"`

## Steps to Reproduce

See above

## Your Environment
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): API via [code-assert](https://github.com/nidi3/code-assert/)
* Operating System and version: Win10 x64 20H2

t-kameyama:
## Description

Fixes #1055

## Checklist

- [x] tests are added


Tapchicoma (approved):
:+1:

PR #1199, opened 2021-08-06
Support absolute paths for globs

### LINKED ISSUE #1131: [Question] What is the reason to drop support for absolute paths in file patterns?
I'm seeing the new exception at https://github.com/pinterest/ktlint/blob/7e2c7fd96fba4ac246639900d0c96c286afeb0fb/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt#L89, as well as a note in release notes that ktlint dropped support for absolute paths in patterns starting from 0.41.0. I understand that this is because of migration to `PathMatcher`, but I don't quite understand the reason for this particular restriction.
Is it a permanent decision or can this behavior be altered in future releases? Or does it break something in expected workflow?

petertrr:
## Description
* Support absolute paths
* Correctly expand tilde in globs (previously was leading to concatenation of absolute paths twice)
* Support unescaped backslashes on Windows (was working in previous versions, but was changed after migration to `PathMatcher`)

This closes #1131 and also two more minor issues I've discovered along the way.
I found out that `PathMatcher` works OK with absolute paths in globs, and based on discussion in #1131 decided that it's no more a design decision to support only relative paths. So I attempted to implement this fix. Will gladly hear any feedback.

## Checklist
- [x] tests are added


akuleshov7:
This should be useful. Finally we will make ktlint great again with absolute paths

PR #1203, opened 2021-08-07
Fix false positive with lambda argument and call chain (`indent`)

### LINKED ISSUE #1202: Indentation regression with inline lambda in function call
Note: this is a regression compared to 0.41.0 that is still present in the 0.42.1 patch release where some indentation bugs were already fixed.

`ktlint -F` turns the following:

```kt
class Foo {
    fun bar() {
        val foo = bar.associateBy({ item -> item.toString() }, ::someFunction).toMap()
    }
}
```

into

```kt
class Foo {
    fun bar() {
        val foo = bar.associateBy({ item -> item.toString() }, ::someFunction).toMap()
        }
    }
```

t-kameyama:
## Description

Fixes #1202

## Checklist

- [x] tests are added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


matejdro:
@t-kameyama Any chance we can get a release for this?

Thanks.

Issue #1229, opened 2021-09-06
Allow rule to define that is should be run after a specific other rule


paul-dingemans:
I want to create custom rule (to indent the content of multiline raw string literals) that only should be executed after the 'indent' rule has been executed. Reason for this is that I do no want to add logic to my custom rule for which the indentation rule is responsible.

KtLint currently offers the Rule.Modifier below:
```
Modifier {
        /**
         * Any rule implementing this interface will be given root ([FileASTNode]) node only
         * (in other words, [visit] will be called on [FileASTNode] but not on [FileASTNode] children).
         */
        interface RestrictToRoot
        /**
         * Marker interface to indicate that rule must be executed after all other rules (order among multiple
         * [RestrictToRootLast] rules is not defined and should be assumed to be random).
         *
         * Note that [RestrictToRootLast] implements [RestrictToRoot].
         */
        interface RestrictToRootLast : RestrictToRoot
        interface Last
    }
 ```

I could create my custom rule without this PR. Due to the current implementation of KtLint this rule would always be run after the standard indent rule. I consider this however to be a bad approach because my rule would not work correctly if the KtLint implementation would change the order in which the rule sets are loaded.

Also internally in KtLint standard rules, it could be beneficial if rule can depend on other rules being executed before. For example the current indentation rule could be split into two distinct rules. A new rule could be created to rearrange the nodes (i.e. insert new lines when needed) while the current indentation rule would only change the indentation of line. For this it needs to be guaranteed that the rearrange rule is executed before the indentation rule.

Currently, each rule in KtLint extends the abstract class Rule and optionally implements a marker interface like shown below:
```
// Current situation: it is not possible to specify that rule1 has to be executed after the 'indent' rule.
class Rule1 :
    Rule("rule-1"),
    Rule.Modifier.RestrictToRootLast {
    // implementation omitted for brevity
}
```

It is not possible to pass down information via the Rule.Modifier interfaces. Neither can those interfaces be changed to a normal class as it would result in multiple parent classes for the rule. It is possible to add an additional parameter in the Rule class to pass down some information. This has some disadvantages though:
  - It violates the Single Responsibility Principle because a Rule should not have any knowledge about the order in which rules are executed.
  - From a maintainability perspective it is difficult to design the structure of this parameter with respect to future wishes regarding rule order execution.

[...]

PR #1230, opened 2021-09-11
Conditional ordering of rules

paul-dingemans:
## Description

For backgrounds see issue: https://github.com/pinterest/ktlint/issues/1229. 

Description below is identical to the commit messages in the order they are applied. Only the last commits is new functionality. All preceding commits are preparations.

### Extract VisitorProvider from Ktlint class

The visitor is created once for each file that has to be linted/formatted. It is now divided into two distinct phases.

The setup phase is executed whenever the VisitorProvider is created. This provider is created only once per invocation of the KtLint command. In the setup phase, the rules are analysed and the order in which they have to be executed is determined. This order is identical for all nodes. As the phase is executed only once, it can now containing logging about the order in which the rules are executed. Also it opens up possibilities for more complex ordering rules using annotations and reflection (for example executing a rule after a specific rule).

The visitor creation phase actually creates the visitor. In this phase all enabled rules will be executed. Note that disabling of rules depends on the '.editorconfig' which is active for a given rootNode.

### Split test into multiple tests

Split DisabledRulesTest into multiple tests.

### Move sorting on ruleSetId from RuleSetsLoader to VisitorProvider

The order in which the rule sets are discovered is not relevant for the loader. Ordering rules by rule set however is important within the scope of the VisitorProvider though.

Also add order that experimental rules takes precedence over custom rules. Finally add secondary sorting on ruleId so that the order of the rules as defined by the RuleSetProviders is not relevant anymore.

### Restrict usage of Rule.Modifier to initialisation phase in which the order of the rules is determined

Rule.Modifier is no longer used during actual creation of the visitor. When creating the visitor the rules are already ordered. This opens up the new possibilities to add other rule ordering mechanisms.

Note that this introduces *functional* changes below with regard to the order of execution of the rules:
- Rules without Rule.Modifier and rules with Rule.Modifier.RestrictToRoot are now handled as one group instead of two distinct groups.
- Rules with Rule.Modifier.Last and rules with Rule.Modifier.RestrictToRootLast are now handled as one group instead of two distinct groups.

### Add annotations RunAsLateAsPossible and RunOnRootNodeOnly as replacements for the Rule.Modifier interfaces

Rules in the standard and experimental rule sets already are converted to use the annotations instead of the Rule.Modifier interfaces. The Rule.Modifiers interface are kept to support a grace period in which maintainers of custom rule sets should migrate as well.

### Add annotation RunAfterRule and alter the rule execution order (NEW FUNCTIONALITY)

Ensures that the rule annotated with RunAfterRule appears in the rule execution order after a specified other rule

## Checklist

- [X] tests are added


shashachu:
thanks for the PR @paul-dingemans. It's an interesting idea! I feel that in theory rules should 'just work' no matter what order they're run - i.e. either they'll have work to do, or not, and if you have conflicting rules, that's sort of your problem. Similarly, you should not depend on some particular rule running before yours in order for your rule to work. But in practice maybe this ends up not being the case. I don't think I have a big problem with this PR, but I don't want to see us introducing a lot of strict ordering requirements in our standard or even experimental set of rules.

paul-dingemans:
I do agree that dependencies between rules should be avoided when possible. But when I have too choose between a rule dependency versus implementing similar logic in multiple rules than I prefer the dependency. 

For example the rule for wrapping arguments also does some indenting but it conflicts with indenting of the indent rule which is confusing when running lint without format. I think it would be better if indentation is solely reported by the the indent rule.

Within the standard and experimental rules we can be strict on using rule dependencies. With this functionality we do empower custom ruleset builders to make their own decisions. But with great power comes great responsibility 😉

romtsn:
Yeah the wrapping rule should just do wrapping, and the indentation should be delegated to the indent rule. We'd need to think of a good way to have rule dependencies.

To me, the downside in this approach is that we have to introduce a ~3MB kotlin-reflect dependency, if possible it would be cool to avoid this.

paul-dingemans:
> To me, the downside in this approach is that we have to introduce a ~3MB kotlin-reflect dependency, if possible it would be cool to avoid this.
I didn't thought about this. I have replaced the annotations with a new sealed class VisitorModifier. Now the kotlin-reflect is no longer needed. I did like the annotation approach a little more, but I am also happy with the new approach.

Issue #1127, opened 2021-03-29
Weird indentation for multiline strings in annotations


Fryie:
I hope this hasn't been discussed yet, but I couldn't find anything related to this specific case.

Applying ktlint version 0.40.0 (I think the problem still occurs with v 0.41.0) to the following file gives an IMHO suboptimal result:
```kotlin
interface UserRepository : JpaRepository<User, UUID> {
    @Query("""
        select u from User u
        inner join Organization o on u.organization = o
        where o = :organization
    """)
    fun findByOrganization(organization: Organization, pageable: Pageable): Page<User>
}
```

## Expected Behavior
Ideally, the indentation is kept as in the original, or at least an indentation is chosen that aligns the ending triple quotes with the annotation or with the beginning triple quotes.

## Observed Behavior
The result is
```kotlin
interface UserRepository : JpaRepository<User, UUID> {
    @Query(
        """
        select u from User u
        inner join Organization o on u.organization = o
        where o = :organization
    """
    )
    fun findByOrganization(organization: Organization, pageable: Pageable): Page<User>
}
```
which looks sub-optimal. *Note:* The problem doesn't occur with some earlier versions of ktlint (e.g. `0.35.0`).

## Steps to Reproduce

Copy code to file, `ktlint -F test.kt`.

## Your Environment

* Version of ktlint used: 0.40.0
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task):
* Version of Gradle used (if applicable):
* Operating System and version: macOS Big Sur
* Link to your project (if it's a public repository):

PR #1262, opened 2021-10-30
Fix indentation of the closing quotes of a multiline string literal

paul-dingemans:
Fix indentation below:

```
val bar = """
    some text
         some more text
"""
```

to

```
val bar = """
    some text
         some more text
    """
```

However, above might be a debatable choice. Therefore I left following TODO in a test:

```
        // TODO: What is the proper way to place the opening and closing quotes of a multi line string value? Note that
        //  the actual content of the multi line string is never been changed as required by ktlint maintainers
        //     val foo =
        //        """
        //        line1
        //        line2
        //        """
        //  or
        //     val foo = """
        //        line1
        //        line2
        //        """
        //  or
        //     val foo = """
        //     line1
        //     line2
        //     """
        //  or
        //     val foo = """
        //        line1
        //        line2
        //     """
        // First option would be most in line with function parameter. Second option is also acceptable. Third option
        // (previous behavior) is not logical as there is no continuation kind of indent visible. Last option is not in
        // sync with function parameter.
```
I would like your opinion about it, so that I can finalize this PR.

This changes also includes the fix for https://github.com/pinterest/ktlint/issues/1127 as it has a comparable problem with the closing quotes.


ktlint-reporter-json/src/test/kotlin/com/pinterest/ktlint/reporter/json/JsonReporterTest.kt
```diff
@@ -67,37 +67,40 @@ class JsonReporterTest {
...
-				"line": 2,
-				"column": 20,
-				"message": "A single thin straight line",
-				"rule": "rule-2"
-			}
-		]
-	}
-]
-""".trimStart().replace("\n", System.lineSeparator())
+            [
```
romtsn, line 70: I think it's totally unreadable now, can we do something about that? is this how the indent rule works now?

paul-dingemans: No, luckily this is not how the rule works. 

After I aligned the closing quotes with the start quotes,  all indentation tabs in the file were automatically replaced with spaces. That was unexpected and confusing. It also difficult to see in the test results what was wrong.

Instead of "{TAB}" we can also use "\t" if you prefer that.

romtsn: I'd like to not have any of this tbh... Does JsonReport emit tabs instead of spaces? I'd rather change it there to emit spaces, same for CheckstyleReporter

paul-dingemans: Yeah sure, no problem to replace the tab indentation with space indentation.

romtsn:
Hm, I'm a bit confused, the description doesn't look relevant to the PR itself, or does it? Also, what part of the diff actually fixes the closing quote? I only see changes relevant to the opening quote if it's placed on a newline so far...

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/IndentationRule.kt
```diff
@@ -345,7 +345,7 @@ class IndentationRule : Rule("indent"), Rule.Modifier.RestrictToRootLast {
         val treeParent = n.treeParent
         if (treeParent.elementType == STRING_TEMPLATE) {
             val treeParentPsi = treeParent.psi as KtStringTemplateExpression
-            if (treeParentPsi.isMultiLine() && treeParentPsi.isFollowedByTrimIndent() && n.treePrev.text.isNotBlank()) {
+            if (treeParentPsi.isMultiLine() && n.treePrev.text.isNotBlank()) {
```
romtsn, line 348: Why did we remove the check for `trimIndent`?

paul-dingemans:
> Hm, I'm a bit confused, the description doesn't look relevant to the PR itself, or does it? Also, what part of the diff actually fixes the closing quote? I only see changes relevant to the opening quote if it's placed on a newline so far...

Sorry about the confusion. I should have stated more clearly that I need input before being able to solve this entirely. I mentioned a debatable choice in the initial description. In one of the other discussion about indentation, you mentioned that you like to align the closing quotes with the opening quotes. This PR was aiming to solve that in the end. Currently the following is allowed:
```
val foo = """
     line 1
     line 2
     """
```

How should the closing quotes be aligned in this case? Aligning would mean, the following.
```
val foo = """
     line 1
     line 2
               """
```
I can not see this as an acceptable solution.

Another solution, would be to wrap the opening quotes:
```
val foo = 
     """
     line 1
     line 2
     """
```
This is my preferred solution, but it can be considered as breaking with previous ktlint version.

The removal of the trimIndent function is possible since we decided that indentation in multiline raw string literals is never changed. Therefore it is not relevant whether this literal is followed by trimIndent, trimMargin, anything else or nothing at all. It has no relation with the goal of the PR and should have been requested in a separate PR.

If you answer my previous question about the placement of the closing quotes, then I will split this PR in separate PR's with a single change in each.

PR #1272, opened 2021-11-03
Recognize 'ij_kotlin_packages_to_use_import_on_demand' from EditorConfig

asmadsen:
## Description

Add configuration option for allowing certain wildcard imports.
This is to make developing with libraries like ktor or kotlin-react more pleasant. These libraries rely heavily on extension functions, which in turn adds a lot of imports.

## Checklist

- [x] tests are added


paul-dingemans:
We should not introduce a new configuration option for this. It is better to use property "ij_kotlin_packages_to_use_import_on_demand" as this has the same purpose in the IntelliJ preferences.In this way we keep IntelliJ preferences and Ktlint configuration in sync.

paul-dingemans:
Hi @asmadsen ,

Do you plan to work on this?

> We should not introduce a new configuration option for this. It is better to use property "ij_kotlin_packages_to_use_import_on_demand" (mapping to field "Packages to use Import with '*' on the import tab of the Kotlin Code Style preferences) as this has the same purpose in the IntelliJ preferences. In this way we keep IntelliJ preferences and Ktlint configuration in sync.
> 
> Note that there is sublte different in the value of this variable which relates to checkbox "With subpackages". This should be taken into account:
> 
> ```
> ij_kotlin_packages_to_use_import_on_demand=com.example.** # with subpackages
> ij_kotlin_packages_to_use_import_on_demand=com.example.* # without subpackages
> ```

asmadsen:
@paul-dingemans Hi, sorry for the late reply.
I'll take a look at it, hopefully I'll be able to update the PR during the weekend.

PR #1273, opened 2021-11-04
Add modifiers and types to adopt explicit api mode

Goooler:
## Description

https://kotlinlang.org/docs/whatsnew14.html#mixing-named-and-positional-arguments

## Checklist

- [ ] tests are added


romtsn:
thanks for the PR, but I think it doesn't make sense to make everything `public` just to get rid of the warnings... we should take a more conscious approach here because not everything needs to be exposed as public API

romtsn:
we'll revisit it with proper visibility modifiers

Goooler:
CC @paul-dingemans.

Goooler:
Link #1373.

paul-dingemans:
> Link #1373.

I was aware of this issue. It should be resolved so that our API user are explicitly informed and that we get rid of those compiler warnings. But as @romtsn mentioned, we should do it a bit more thoughtful and per case decide whether we want to define something as public or that it should be an internal.

PR #1279, opened 2021-11-06
Prevent false warning about missing custom rule set

### LINKED ISSUE #1269: Custom gradle tasks with custom ruleset results in warning
Running ktlint in gradle with a custom task (ie no plugin) AND a custom rule set results in the following warning:
[WARNING] Number of custom rule sets loaded does not match the expected number of rule sets to be loaded.
          Expected to load 0 custom rule set(s) for rules:
          Actually loaded 1 custom rule set(s) with names: XXXX
          One or more of the specified jars does not provide the custom rule set correctly. Check following:
            - Does the jar contain an implementation of the RuleSetProvider interface?
            - Does the jar contain a resource file with name "com.pinterest.ktlint.core.RuleSetProvider"?
            - Is the resource file located in directory "src/main/resources/META-INF/services"?
            - Does the resource file contain the fully qualified class name of the class implementing the RuleSetProvider interface?

## Expected Behavior
Loading custom rule sets should NOT result in a warning unless the custom rules failed to meet the other rules.

## Observed Behavior
Custom rules are loaded and executed, just an invalid warning is being displayed.

## Steps to Reproduce
Load custom rule set using:
```
val ktlint by configurations.creating

dependencies {
    ktlint("com.pinterest:ktlint:0.43.0") {
        attributes {
            attribute(Bundling.BUNDLING_ATTRIBUTE, objects.named(Bundling.EXTERNAL))
        }
    }
    ktlint(project(":custom-ktlint-ruleset"))
}
```

## Your Environment

* Version of ktlint used: 0.43.0
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): custom gradle tasks
* Version of Gradle used (if applicable): 7.2
* Operating System and version: OSX 12.0.1
* Link to your project (if it's a public repository): n/a

## Additional Notes
This did not occur in 0.42.1 or earlier versions. Looking through the new RuleSetLoader, it appears the bug is being caused because the expected number of custom rulesets is being counted from the number of `--ruleset` parameters but the custom gradle tasks provides them via the classpath so the actual vs expected count doesn't get calculated correctly.

paul-dingemans:
## Description

Prevent false warning about missing custom rule set

Check JAR files one by one whether a custom rule set has been provided and print warning when it is missing.

Closes #1269

## Checklist

- [ ] tests are added


romtsn:
I'm not sure if you've seen my https://github.com/pinterest/ktlint/pull/1226#discussion_r733515074, do you think this check is still worth having, even though it can be wrong?

paul-dingemans:
Compared to #1226 the check has been improved. In #1226 the number of Jars was compared with the number of custom rulesets loaded. That check indeed was not foolproof. Now I have improved the check. Now it checks for each added JAR whether ar least one custom ruleset provider is found.

Tapchicoma:
@paul-dingemans could you add a CLI test for this warning?

paul-dingemans:
@Tapchicoma CLI tests have been added to check that the warning regarding the missing ruleset provider are only printed when required.

I have refactored the BaseCLITest in order to make it work:
* Arguments were separated by comma (default) instead of space
* Arguments can now refer to the BASE_DIR_PLACEHOLDER constant so that the jars that are stored in the same directory as the CLI-test can be reference correctly in the command line arguments
* Improved the output of the assertion in case the assertion fails

paul-dingemans:
The fix typo commit is a kind of fake commit which was needed to retrigger the build. Previously, only one of three required checks was executed. I was not able to restart the missing checks. I think this was caused by the fact that this branch originally targetted another branch than "master". After changing the target, only the "Validate Gradle Wrapper" check was fired.

Issue #1255, opened 2021-10-20
Unnecessary newline before "(" for nullable lambda definitions.


ggajews:
## Expected Behavior
No error reported

## Observed Behavior
ktlint is formatting following line
`var changesListener: ((width: Double?, depth: Double?, length: Double?, area: Double?) -> Unit)? = null`
like this
```
var changesListener: (
        (
            width: Double?,
            depth: Double?,
            length: Double?,
            area: Double?
        ) -> Unit
    )? = null
```

but at the same time it reports an issue:
`xxx.kt:21:9: Unnecessary newline before "("`

## Your Environment

* Version of ktlint used: 0.42.1
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): `org.jlleitschuh.gradle:ktlint-gradle:10.2.0`

paul-dingemans:
Closed by #1284

paul-dingemans:
In this case the parameter-list-wrapping rule is a bit too aggressive. Because the max-line-length is exceeded, this rule decides to move each parameter to a separate line but this should not have been applied to the parameters of the lambda as well. So the code should have been wrapped as follows:
```
var changesListener: (
    (width: Double?, depth: Double?, length: Double?, area: Double?) -> Unit
)? = null
```
This formatting is also accepted by the default IntelliJ IDEA formatting.

It becomes more interestingly when the lambda parameter would still exceed the max-line-lenght. In that case, I suggest to not wrap the rule (this strategy is used in other rules as well). Reason for this is that the default IntelliJ IDEA formatting is not able, to format this properly. For example:
```
var changesListener: (
    ( width: Double?,
      depth: Double?,
      length: Double?,
      area: Double?
    ) -> Unit
 )? = null
 ```
 will be changed to code below by default IntelliJ IDEA:
 ```
var changesListener: (
    (
    width: Double?,
    depth: Double?,
    length: Double?,
    area: Double?
) -> Unit
)? = null
```

Issue #1119, opened 2021-03-24
Wrong parameter list wrapping


vRallev:
## Expected Behavior
IntelliJ formats a top-level function like that:
```kotlin
private fun <T, U : LayerRendering> StatelessWorkflow<
  AuthenticatorInput, AuthenticatorOutput, LayeredScreen<PosLayering>
  >.RenderContext.renderChildToLayeredScreen(
  workflow: Workflow<T, Unit, U>,
  props: T
): LayeredScreen<MainAndModal> {
  ...
}
```

## Observed Behavior
KtLint complains about:
```
file.kt:4:3: Unexpected indentation (expected 4, actual 2) (parameter-list-wrapping)
file.kt:5:3: Unexpected indentation (expected 4, actual 2) (parameter-list-wrapping)
file.kt:6:1: Unexpected indentation (expected 2, actual 0) (parameter-list-wrapping)
```

Changing the function to:
```kotlin
private fun <T, U : LayerRendering> StatelessWorkflow<
  AuthenticatorInput, AuthenticatorOutput, LayeredScreen<PosLayering>
  >.RenderContext.renderChildToLayeredScreen(
    workflow: Workflow<T, Unit, U>,
    props: T
  ): LayeredScreen<MainAndModal> {
  ...
}
```
will trigger these errors:
```
file.kt:4:1: Unexpected indentation (4) (should be 2) (indent)
file.kt:5:1: Unexpected indentation (4) (should be 2) (indent)
file.kt:6:1: Unexpected indentation (2) (should be 0) (indent)
```

## Steps to Reproduce
See sample above.

## Your Environment

* Version of ktlint used: 0.41.0
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): custom Gradle task
* Operating System and version: Mac / Linux
* `indent_size = 2`

ileasile:
Maybe the same bug:

`build.gradle.kts`:
```kotlin
tasks.test {
    systemProperties = mutableMapOf(
        "junit.jupiter.displayname.generator.default" to
            "org.junit.jupiter.api.DisplayNameGenerator\$ReplaceUnderscores",

        "junit.jupiter.execution.parallel.enabled" to
            doParallelTesting.toString() as Any,
        "junit.jupiter.execution.parallel.mode.default" to
            "concurrent",
        "junit.jupiter.execution.parallel.mode.classes.default" to
            "concurrent"
    )
}
```

[...]

Issue #1045, opened 2021-01-03
adding disabledRules=["experimental:argument-list-wrapping"]  triggers IndexOutOfBoundsException


phileo:
## Steps to Reproduce
```
brew install ktlint
git clone https://github.com/ricknout/android-mdc-theming
cd app/src/main/java
```

then run the following from command line:
`ktlint -F --experimental --android --verbose --debug  --disabled_rules=final-newline,parameter-list-wrapping,argument-list-wrapping,no-wildcard-imports,max-line-length,experimental:package-name
`

Tapchicoma:
I could not reproduce to get exactly `IndexOutOfBoundsException`, but I got another one:
```kotlin
[DEBUG] Checking android-mdc-theming/app/src/main/java/com/ricknout/mdctheming/theming/ThemingActivity.kt
[DEBUG] Internal Error (experimental:multiline-if-else)
com.pinterest.ktlint.core.RuleExecutionException: java.lang.ClassCastException: org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.CompositeElement cannot be cast to org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
        at com.pinterest.ktlint.core.KtLint$format$1.invoke(KtLint.kt:320)
        at com.pinterest.ktlint.core.KtLint$format$1.invoke(KtLint.kt:27)
        at com.pinterest.ktlint.core.KtLint$visitor$2$2.invoke(KtLint.kt:247)
        at com.pinterest.ktlint.core.KtLint$visitor$2$2.invoke(KtLint.kt:27)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:225)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:226)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:226)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:226)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:226)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:226)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:226)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:226)
        at com.pinterest.ktlint.core.ast.PackageKt.visit(package.kt:226)
[...]

Issue #897, opened 2020-09-10
Infix function confuses the indent rule


androa:
## Expected Behavior

I'm writing a DSL for expressing some rules, and use a lot of `infix` functions.

If write the following:
```
val rules = "rule1".all(
    "subrule1".all(
        date1 notBefore date2,
    )
).then(
    "subrule2".all(
        date3 notBefore date4
    )
) or "rule1".all(
    date4 notBefore date6
)
```
everything works as expected. If I use the infix `then` ktlint becomes confused
```
val rules = "rule1".all(
    "subrule1".all(
        date1 notBefore date2,
    )
) then(
    "subrule2".all(
        date3 notBefore date4
    )
) or "rule1".all(
    date4 notBefore date6
)
```

## Observed Behavior
ktlint can't decide on how much it wants to indent the three last lines. It just keeps sending me around fixing it back and forth from 0 -> 4 -> 8 -> 4 -> 8 -> 4 -> 8 and so forth

`ktlint --verbose` says:
```
File.kt:48:5: Unexpected indentation (expected 8, actual 4) (parameter-list-wrapping)
File.kt:49:1: Unexpected indentation (expected 4, actual 0) (parameter-list-wrapping)
```

## Your Environment
* Version of ktlint used: 0.38.1

abubics:
I'm in a similar situation, but the symptoms are a little more deterministic (so I hope the extra detail helps narrow anything down).

I'm using [Exposed](https://github.com/JetBrains/Exposed), which likewise has an infix-heavy DSL, and sometimes needs brackets for readability.

When trying a long chained expression like this:
```kotlin
AmendmentTable
  .innerJoin(PlanningSchemeAmendmentTable)
  .innerJoin(PlanningSchemeTable)
  .select {
    (PlanningSchemeTable.code eq schemeCode) and (
      (AmendmentTable.exhibitionStartDate eq null) or
        (AmendmentTable.exhibitionStartDate lessEq now())
    )
  }
```
the `infix fun and` causes ktlint issues with its closing bracket (wanting to be indented an extra step).

Changing it to
```kotlin
    (PlanningSchemeTable.code eq schemeCode).and(
```
(no longer infix) removes the reported violation.

Also, indenting it as advised removes the violation, unlike the above reported issue (where it alternates between errors).

## Your Environment
- ktlint version: 0.41.0? It's not clear from the gradle plugin, and doesn't appear in `gradle dependencies`.
[...]

PR #1284, opened 2021-11-10
Remove lint warning about unexpected indentation outside of indent rule

### LINKED ISSUE #1267: Split type arguments break 0.43.0
## Expected Behavior
I expect it to format it or accept it

## Observed Behavior
It fails with an error like this

```
Step 'ktlint' found problem in 'path/to/file.kt':
Error on line: 48, column: 7
Unexpected indentation (expected 4, actual 6)
java.lang.AssertionError: Error on line: 48, column: 7
Unexpected indentation (expected 4, actual 6)
        at com.diffplug.spotless.kotlin.KtLintStep$State.lambda$createFormat$0(KtLintStep.java:129)
        at jdk.proxy49/jdk.proxy49.$Proxy441.invoke(Unknown Source)
        at com.pinterest.ktlint.core.KtLint.format(KtLint.kt:428)
        at com.pinterest.ktlint.core.KtLint.format(KtLint.kt:352)

```

## Steps to Reproduce
this snippet fails for us

```kotlin
    fun setMessagingChannelDisplayNameMap(
      messagingChannelDisplayNameMap: Map<String, // This line
        String>
    ): Builder = apply {
      this.messagingChannelDisplayNameMap =
        messagingChannelDisplayNameMap
    }
```

## Your Environment

* Version of ktlint used: 0.43.0
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): Spotless
* Version of Gradle used (if applicable): N/A
* Operating System and version: N/A
* Link to your project (if it's a public repository): N/A

paul-dingemans:
## Description

Only the indentation rule should emit warnings about incorrect indentation to avoid
conflicting warnings from different rules about the indentation of the exact same
line. However, those other rules should still fix the indentation as good as they
can for the case that the indent rule is not run at all.

Closes #1267, #1119, #1045, #1255, #1319, #1320, #897, #1337

## Checklist

- [ ] tests are added


ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ArgumentListWrappingRule.kt
```diff
@@ -139,18 +139,19 @@ class ArgumentListWrappingRule : Rule("argument-list-wrapping") {
...
+                                    // Do not emit a warning about incorrect indentation. Final indentation is
+                                    // determined by the indent rule. Displaying, possibly conflicting warnings about
+                                    // the exact indenting position is really confusing for the user.
                                 } else {
                                     emit(child.startOffset, errorMessage(child), true)
                                 }
                                 if (autoCorrect) {
-                                    val finalIndent =
-                                        (if (cut > -1) spacing.substring(0, cut) else "") + intendedIndent
+                                    // Although no warning is displayed about incorrect indentation, an attempt is
```
shashachu, line 149: This seems strange that we wouldn't emit a warning on linting, but we'd change indentation on formatting.

paul-dingemans: Yes it is a bit unusual. But look at the number of issue we get about those warnings. So the warning has to be removed at minimum. Keeping the indent fix during format only has value in case the indent rule is disabled.

shashachu: We shouldn't have different behavior for linting vs formatting.

paul-dingemans: I am fine with removing the functionality to fix indentation inside the argument/parameter list wrapping rule in order to make linting and formatting consistent again. Some users might consider this a breaking change.

@shashachu please confirm whether that is the way to go.

romtsn: yes, let's disable autocorrection of indentation for now

paul-dingemans:
Like before a warning is emitted in case the current indentation is not as expected. However, the emitted warning now no longer reports the expected indentation but just that the argument/parameter should be on a separate line unless all of them fit on a single line.

romtsn:
hm, can we remove autocorrection of indentation entirely from these rules? let's just delegate this to the indent-rule completely and say that we do not support indentation alignment in these wrapping rules (even the names suggest that they are wrapping and not indenting :))

paul-dingemans:
> hm, can we remove autocorrection of indentation entirely from these rules? let's just delegate this to the indent-rule completely and say that we do not support indentation alignment in these wrapping rules (even the names suggest that they are wrapping and not indenting :))

I think this would give a bad user experience in case the indent rule would be disabled. More importantly is that it would have a big Ktlint-developer impact as this would also imply that the expected formatted code in unit tests would have to be de-indented. So, writing a test like below:
```
    @Test
    fun testFormatArgumentsWithNestedCalls() {
        assertThat(
            ArgumentListWrappingRule().format(
                """
                val x = test(
                    one("a", "b",
                    "c"),
                    "Two", "Three", "Four"
                )
                """.trimIndent()
            )
        ).isEqualTo(
            """
            one(
            "a",
            "b",
            "c"
            ),
            "Two",
            "Three",
            "Four"
            )
            """.trimIndent()
        )
    }
```

romtsn:
gotcha, let's merge it like that then!

paul-dingemans:
When checking all reported issues, I encountered a problem which I did not see before. The incorrect indentation warning was sometimes replace with another incorrect warning. No I have changed the rules to ignore the indentation if the the whitespace before the parameter, argument or closing parathesis contains a new line because this means that the wrapping was done correctly. Previously an incorrect warning was shown in case that whitespace contained a newline but was incorrectly indented according to that rule.

In order to keep the unit tests of the rules readable i have modified the lint/format utilities to accept a list of rules instead of a single rule. In this way, I can execute both the wrapping as well as the indentation rule together. I am going to need that functionality also for the new signature rewrite rule that I am developing, so I think it is a good addition to our testing toolkit.

I have, once more checked that all issues associated with this PR are now indedd resolved. If the build succeeds, I will merge to master.

PR #1285, opened 2021-11-15
Adding a new rule: Underscore numeric literals.

nexsja:
## Description

Added a new rule for numeric literals: numeric literals should be chunked in 3s and delimited with a `_`.

## Checklist

- [x] tests are added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## Unreleased
 
 ### Added
+- New `underscore-numeric-literal` rule
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


romtsn:
Thanks for the PR, but I think we cannot accept it as part of the standard ruleset (at best - experimental), as no such convention exists in https://kotlinlang.org/docs/coding-conventions.html. The rule also can be considered highly opinionated - we will discuss whether we introduce a new `contrib` ruleset, if yes, this can definitely be part of it

nexsja:
Fair enough, thanks for the feedback. :)

If you do decide to add this to any other ruleset - I don't mind finishing up whatever's necessary. Thanks! :)

shashachu:
@nexsja we've considered having some sort of repo or ruleset with community-contributed rules, but it doesn't actually make a lot of sense right now to host them all into a single ruleset, since it's very unlikely any developer would want all of them. We would need to introduce the idea of code style sets (https://github.com/pinterest/ktlint/issues/548) in order for something like this to make sense. But thanks for the contribution! I encourage you to host this rule in your own repo. :)

Issue #1245, opened 2021-10-06
Wrong indentation when use tab style


Keldmar:
When using tabs instead of spaces, I constantly get an error in the class constructor, if the parameters are specified vertically, in the example you can see that only the second class is validated, when using spaces there is no such problem

.editorconfig file
```
root = true
[*.{kt,kts}]

indent_style = tab
```

Sample class
```
package somepackage

internal class SomeClass(
				private val someField1: String,
	private val someField2: String,
	private val someField3: String
) {

	val s = 1

	fun someTest(
		someField1: String,
		someField2: String,
		someField3: String
	) {
		/* ignore */
	}
}

internal class SomeClass2(private val someField1: String, private val someField2: String, private val someField3: String) {

	val s = 1

	fun someTest(someField1: String, someField2: String, someField3: String) {
		/* ignore */
	}
}

```
![image](https://user-images.githubusercontent.com/26333514/136213461-90f5725f-508a-4339-8337-d9a4cee3eca5.png)

```
SomeClassTest.kt:4:1: Unexpected indentation (4) (should be 1)
SomeClassTest.kt:4:5: Unexpected indentation (expected 4, actual 4)
SomeClassTest.kt:5:2: Unexpected indentation (expected 4, actual 1)
SomeClassTest.kt:6:2: Unexpected indentation (expected 4, actual 1)
SomeClassTest.kt:12:3: Unexpected indentation (expected 5, actual 2)
SomeClassTest.kt:13:3: Unexpected indentation (expected 5, actual 2)
SomeClassTest.kt:14:3: Unexpected indentation (expected 5, actual 2)
```

paul-dingemans:
There are more known problems related to tab indentations. I have solved most of them a long time ago but it's PR (and several other unpublished PR's which are blocked by it) is till in the process of reviewing.

paul-dingemans:
@Keldmar  @VovaStelmashchuk The PR mentioned before was merged into version 0.43.0 recently. Can you please check whether the tab indentation problems are solved for you when running with that version? Please note, that we have several open problems with this release, but they are not related to tab indentation.

paul-dingemans:
Closing issue as the problem is resolved in a previous release. Please reopen if the issue is not yet solved for you.

bfichter:
I'm still experiencing this issue with 0.43.2

paul-dingemans:
> I'm still experiencing this issue with 0.43.2

@bfichter please add details for your specific problem and setup in a new issue.

PR #1289, opened 2021-11-26
Fixed parameter list wrapping when indent style is tab

ipinlnd:
## Description
When the `indent_style` is set to be `tab`, parameter wrapping doesn't consider that tab is 1 character and it expects the indention to be higher than just 1 character. 

Sample code in this issue: https://github.com/pinterest/ktlint/issues/1245

In this PR, I've added a new property to the `ParameterListWrappingRule` class named `indentCharacter` which would be a `" "` or `"\t"` based on the settings and also set the `indentSize` to be 1 in case of tabs. 

## Checklist

- [ x] tests are added

### DIFF
--- a/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ArgumentListWrappingRule.kt
+++ b/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ArgumentListWrappingRule.kt
@@ -1,5 +1,6 @@
 package com.pinterest.ktlint.ruleset.experimental
 
+import com.pinterest.ktlint.core.EditorConfig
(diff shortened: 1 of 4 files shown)


paul-dingemans:
Ok nice. But note that fixing indentation should primarily be done in the indent rule, which already should be fully tab compatible. I have submitted this https://github.com/pinterest/ktlint/pull/1284 for that.

Can you please apply the same fix in the argument list wrapping rule?

ipinlnd:
@paul-dingemans oh well then in that case, we don't need this PR at all. If your PR gets merged, then i'll close my PR, because that would solve this issue. Meanwhile, i'll make the fix for the argument list wrapping rule as well, thanks

romtsn:
Closing in favour of #1284

PR #1294, opened 2021-11-30
Add Gradle toolchain setup.

Tapchicoma:
## Description

- Unify Kotlin setup via compiler plugin and use jvm toolchain with configured language version 8.
- Add tasks to run additionally tests on JDK 11 - "testOnJdk11"
- Change CI setup to not use java versions matrix - JDK 11 will be auto-provisioned by Gradle itself
- Update Dokka to `1.6.0` version (previous version caused issues on CI run)

## Checklist

- [x] tests are added


buildSrc/build.gradle.kts
```diff
@@ -2,11 +2,25 @@ plugins {
...
+
 repositories {
     mavenCentral()
-    gradlePluginPortal()
+}
+
+// Pass '-PkotlinDev' to command line to enable kotlin-in-development version
+val kotlinVersion = if (project.hasProperty("kotlinDev")) {
+    logger.warn("Enabling kotlin dev version!")
+    "1.6.0"
```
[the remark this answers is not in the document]
Tapchicoma: For now  - no, `1.6.20-M1` should be out around next week

shashachu (approved):

buildSrc/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+    id("org.jetbrains.kotlin.jvm")
+}
+
+repositories {
+    mavenCentral()
+}
+
+kotlin {
+    jvmToolchain {
+        (this as JavaToolchainSpec).languageVersion.set(JavaLanguageVersion.of(8))
```
mateuszkwiecinski, line 13: Out of curiocity, as I'm not sure I get toolchains properly: Wouldn't you want to explicitly set target version? I can imagine at some point you would like to bump toolchain version to Java 17, but the output artifact should remain compatible with Java 8. 
If someone changes the toolchain version it will affect produced artifacts which is an unwanted behavior, right? 

Basically I understood you always want to compile the project with highest possible (java and kotlin) versions, but java target version and kotlin api version should be set to the lower library supports - is that wrong assumption?

Tapchicoma: > Wouldn't you want to explicitly set target version?

Toolchain by default sets target version to be equal to the toolchain one if it is not set explicitly.

>  If someone changes the toolchain version it will affect produced artifacts which is an unwanted behavior, right?

Right, but such change should go via PR review :wink:

> Basically I understood you always want to compile the project with highest possible (java and kotlin) versions, but java target version and kotlin api version should be set to the lower library supports

Initially I want to use JDK 1.8 to compile and add later tests for CLI that runs against latest JDK version. Generally it is also possible to use toolchain with JDK 17 + set target to "1.8". That change could be done at a later time.

mateuszkwiecinski: Great, that was the answer I was looking for. Thanks for explanation 👍

romtsn (approved):

PR #1301, opened 2021-12-05
Do not remove trailing comma in value argument list in annotation

### LINKED ISSUE #1297: The trailing comma experimental rule does not working properly
The trailing comma experimental rule is not working properly for this file:

https://github.com/mohsenoid/Rick-and-Morty/blob/c9fb8dc093c7986055e8892b8e126f60bf0247cc/app/src/main/kotlin/com/mohsenoid/rickandmorty/data/db/DbRickAndMorty.kt#L51

Android Studio works correctly and adds a comma, but Ktlint doesn't and even removes the comma added by Android Studio!

## Expected Behavior
Add a comma and do not remove the one from the Android Studio code styler

## Observed Behavior
Do not add a comma and also remove the one from the Android Studio code styler

## Steps to Reproduce
Take [this file](https://github.com/mohsenoid/Rick-and-Morty/blob/c9fb8dc093c7986055e8892b8e126f60bf0247cc/app/src/main/kotlin/com/mohsenoid/rickandmorty/data/db/DbRickAndMorty.kt) and enable both these rules in .editorconfig
```
ij_kotlin_allow_trailing_comma = true
ij_kotlin_allow_trailing_comma_on_call_site = true
```
execute this command:
```
ktlint -F --experimental "app/src/**/*.kt"
```

## Your Environment
* Version of ktlint used: CLI 0.43.0
* Operating System and version: Win 11
* Link to your project (if it's a public repository): https://github.com/mohsenoid/Rick-and-Morty/tree/294ec735fb8a59a4ca5f3abc7c556b713cd9c82a

paul-dingemans:
## Description

The code previously contained special handling for case below:
```
    @Annotation([
        "something",
    ])
```

This special case however was also triggered in case of:
```
    @Annotation(
       foo = [
        "something"
       ],
       bar = "some-bar"
   )
```

In this case the normal handling of value argument lists should be used once for the outer value argument list (containing elements foo and bar) and once for the inner value argument list of parameter foo (containing element "something"). So in example above, a trailing comma has to be added to both the outer and the inner value argument list.

Closes #1297

## Checklist

- [X] tests are added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 - Fix false positive in rule spacing-between-declarations-with-annotations ([#1281](https://github.com/pinterest/ktlint/issues/1281))
+- Do not remove trailing comma in annotation ([#1297](https://github.com/pinterest/ktlint/issues/1297))
 
 ### Changed
 - Update Kotlin version to `1.6.0` release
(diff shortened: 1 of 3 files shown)


mohsenoid:
@paul-dingemans Thanks for taking care of that. 💪

romtsn (approved):

PR #1307, opened 2021-12-12
Add build task experimental rules

paul-dingemans:
## Description

Ktlint should apply the dogfooding principal and only provide experimental rules that at least on the ktlint code base itself gives satisfiable results.

In the intial commit of the PR all experimental rules that causes a violation have been disabled. In following commits, rules are enabled one by one to make the consequences of enabling a rule clear. See commit messages for further details about choices made. Most important choices:
* experimental rules should always be enforced on the ktlint code base itself. If we are not happy with the effect of a rule on our own code base, we should not provide such a rule
* The experimental:argument-list-wrapping is the only rule which is not yet enabled as it breaks on some false
positives. This will be solved by https://github.com/pinterest/ktlint/pull/1284 after which this rule can and should be enabled as well.
* For now, it has been chosen to disallow trailing comma's instead of forcing them to be added. I am also fine with enabling the rules. But this results in a lot of changes (145 files). It will be best to change this directly after merging this PR into master.

## Checklist

- [ ] tests are added


build.gradle
```diff
@@ -48,6 +48,13 @@ task ktlint(type: JavaExec, group: LifecycleBasePlugin.VERIFICATION_GROUP) {
   args '**/src/**/*.kt', '--baseline=ktlint-baseline.xml', '--verbose'
 }
 
+task ktlint_experimental(type: JavaExec, group: LifecycleBasePlugin.VERIFICATION_GROUP) {
+  description = "Check Kotlin code style with experimental rules."
+  classpath = configurations.ktlint
+  main = 'com.pinterest.ktlint.Main'
+  args '**/src/**/*.kt', '--verbose', '--experimental'
```
romtsn, line 55: I think we don't need a new task for that, I'd just add `--experimental` to the task above. Otherwise, we'd also need to update the CI workflow

ktlint/src/test/resources/TestBaselineFile._kt
```diff
@@ -0,0 +1,5 @@
+// This class is stored on purpose in a file having extension "_kt". This prevents the file from being autocorrected by
```
romtsn, line 1: I think we can just exclude this file (and also `TestBaselineExtraErrorFile`) via negation from the task execution (like `!ktlint/src/test/resources/TestBaselineFile.kt`)

romtsn (approved):
2 minor things, but LGTM otherwise, thanks.

build.gradle
paul-dingemans: Yes indeed. I was not sure whether you was willing to integrate it in the default pipeline. But I definitely am in favor of this.

ktlint/src/test/resources/TestBaselineFile.kt
```diff
@@ -1,3 +1,5 @@
+// This class is stored on purpose in a file having extension "_kt". This prevents the file from being autocorrected by
+// ktlint when the ktlint code base itself is reformatted.
```
romtsn, line 2:
```suggestion

```

ktlint/src/test/resources/TestBaselineFile.kt
```diff
@@ -1,3 +1,5 @@
+// This class is stored on purpose in a file having extension "_kt". This prevents the file from being autocorrected by
```
romtsn, line 1:
```suggestion
```

ktlint/src/test/resources/TestBaselineExtraErrorFile.kt
```diff
@@ -1,3 +1,5 @@
+// This class is stored on purpose in a file having extension "_kt". This prevents the file from being autocorrected by
```
romtsn, line 1:
```suggestion
```

ktlint/src/test/resources/TestBaselineExtraErrorFile.kt
```diff
@@ -1,3 +1,4 @@
+// ktlint when the ktlint code base itself is reformatted.
```
romtsn, line 1:
```suggestion
```

romtsn:
Suggested edit:
```diff
diff --git a/ktlint/src/test/kotlin/com/pinterest/ktlint/BaselineTests.kt b/ktlint/src/test/kotlin/com/pinterest/ktlint/BaselineTests.kt
--- ktlint/src/test/kotlin/com/pinterest/ktlint/BaselineTests.kt
+++ ktlint/src/test/kotlin/com/pinterest/ktlint/BaselineTests.kt
@@ -32,16 +32,16 @@
         val ps = PrintStream(stream)
         System.setOut(ps)
 
         try {
-            main(arrayOf("src/test/resources/TestBaselineFile._kt"))
+            main(arrayOf("src/test/resources/TestBaselineFile.kt"))
         } catch (e: ExitException) {
             // handle System.exit
         }
 
         val output = String(stream.toByteArray())
-        assertTrue(output.contains(".*:3:24: Unnecessary block".toRegex()))
-        assertTrue(output.contains(".*:4:1: Unexpected blank line\\(s\\) before \"}\"".toRegex()))
+        assertTrue(output.contains(".*:1:24: Unnecessary block".toRegex()))
+        assertTrue(output.contains(".*:2:1: Unexpected blank line\\(s\\) before \"}\"".toRegex()))
     }
 
     @Test
     fun testBaselineReturnsNoErrors() {
@@ -49,16 +49,16 @@
         val ps = PrintStream(stream)
         System.setOut(ps)
 
         try {
-            main(arrayOf("src/test/resources/TestBaselineFile._kt", "--baseline=src/test/resources/test-baseline.xml"))
+            main(arrayOf("src/test/resources/TestBaselineFile.kt", "--baseline=src/test/resources/test-baseline.xml"))
         } catch (e: ExitException) {
             // handle System.exit
         }
 
         val output = String(stream.toByteArray())
-        assertFalse(output.contains(".*:3:24: Unnecessary block".toRegex()))
-        assertFalse(output.contains(".*:4:1: Unexpected blank line\\(s\\) before \"}\"".toRegex()))
+        assertFalse(output.contains(".*:1:24: Unnecessary block".toRegex()))
+        assertFalse(output.contains(".*:2:1: Unexpected blank line\\(s\\) before \"}\"".toRegex()))
     }
 
     @Test
     fun testExtraErrorNotInBaseline() {
@@ -66,16 +66,16 @@
         val ps = PrintStream(stream)
         System.setOut(ps)
 
         try {
-            main(arrayOf("src/test/resources/TestBaselineExtraErrorFile._kt", "--baseline=src/test/resources/test-baseline.xml"))
+            main(arrayOf("src/test/resources/TestBaselineExtraErrorFile.kt", "--baseline=src/test/resources/test-baseline.xml"))
         } catch (e: ExitException) {
             // handle System.exit
         }
 
         val output = String(stream.toByteArray())
-        assertFalse(output.contains(".*:3:24: Unnecessary block".toRegex()))
-        assertTrue(output.contains(".*:4:1: Unexpected blank line\\(s\\) before \"}\"".toRegex()))
+        assertFalse(output.contains(".*:1:24: Unnecessary block".toRegex()))
+        assertTrue(output.contains(".*:2:1: Unexpected blank line\\(s\\) before \"}\"".toRegex()))
     }
 
     private class ExitException(val status: Int) : SecurityException("Should not exit in tests")
 }

```

paul-dingemans:
@romtsn Can you give it another try? I have fixed the tests.

PR #1305, opened 2021-12-11
Fix indentation of delegated super type call entry

paul-dingemans:
## Description

Forces the BY keyword in a super type call entry to be indented when preceded by a new line

The code sample from issue #1210 is now formatted as:
```
object ApplicationComponentFactory : ApplicationComponent.Factory
    by DaggerApplicationComponent.factory()
```

As a side effect, this also formats
```
class Test2 : Foo
by Bar(
    a = 1,
    b = 2,
    c = 3
)
```
to
```
class Test2 : Foo
    by Bar(
        a = 1,
        b = 2,
        c = 3
    )
```

Closes #1210

## Checklist

- [X] tests are added
- [X] `CHANGELOG.md` is updated

Issue #1210, opened 2021-08-11
Internal Error when upgrading from 0.41.0 to 0.42.1


NikolayMetchev:
After upgrading 2 files cause ktlint to fail with an Internal Error (indent)
[SettlementsSpec.kt.zip](https://github.com/pinterest/ktlint/files/6966917/SettlementsSpec.kt.zip)
[generators.kt.zip](https://github.com/pinterest/ktlint/files/6966925/generators.kt.zip)

## Expected Behavior
the file should get formatted

## Observed Behavior
SettlementsSpec.kt:0:0: Internal Error (indent). Please create a ticket at https://github.com/pinterest/ktlint/issues (if possible, provide the source code that triggered an error) (cannot be auto-corrected)

## Steps to Reproduce
Format the attached file 

## Your Environment
* Version of ktlint used: 0.42.1
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task):
Gradle Task:
task ktlintFmt(type: JavaExec) {
    mainClass.set("com.pinterest.ktlint.Main")
    classpath = configurations.ktlint
    args "-F", "src/**/*.kt"
    onlyIf {
        System.env['CI'] == null
    }
}
* Version of Gradle used (if applicable): 7.1.1
* Operating System and version: MacOS Big Sur
* Link to your project (if it's a public repository):

shashachu:
@NikolayMetchev are you able to test with the latest SNAPSHOT? I believe this commit fixes your issues https://github.com/pinterest/ktlint/pull/1203

mhernand40:
FWIW, the following snippet also triggers an Internal Error in ktlint `0.42.1`
```Kotlin
package com.repro

import kotlin.properties.Delegates

class Repro {
    private var repro
        by Delegates.vetoable("") { _, old, new ->
            true
        }
}
```

The error message is:
```
Internal Error (indent). Please create a ticket at https://github.com/pinterest/ktlint/issues (if possible, provide the source code that triggered an error) (cannot be auto-corrected)
```

No issue if I re-format to
```Kotlin
package com.repro

import kotlin.properties.Delegates

class Repro {
    private var repro by Delegates.vetoable("") { _, old, new ->
        true
    }
}
```

paul-dingemans:
@mhernand40 Tnx for your example and observation how reformat made a difference. The stacktrace of the error is as follows:
```
[DEBUG] Internal Error (indent)
[...]

PR #1322, opened 2021-12-22
Fix indent of "by" keyword

paul-dingemans:
## Description

Original fix for #1210 (https://github.com/pinterest/ktlint/pull/1305) solved the runtime exception during formatting but also reformatted some delegate by constructs which are conflicting with the defaultIntelliJ formatting.

The runtime exception is still solved. I have added an explicit assert statement after the formatting stage which break units tests that would result in a negative indent level but otherwise still woud succeed. This should protect us against similar indent problem in the future.

The new formatting introduced in https://github.com/pinterest/ktlint/pull/1305 is reverted.

## Checklist

- [ ] tests are added


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/IndentationRule.kt
```diff
@@ -742,7 +748,13 @@ class IndentationRule : Rule("indent"), Rule.Modifier.RestrictToRootLast {
         val byKeywordLeaf = n
             .findChildByType(DELEGATED_SUPER_TYPE_ENTRY)
             ?.findChildByType(BY_KEYWORD)
-        if (n.prevLeaf()?.textContains('\n') == true && byKeywordLeaf?.prevLeaf().isWhiteSpaceWithNewline()) {
+        if (n.prevLeaf()?.textContains('\n') == true &&
+            byKeywordLeaf?.prevLeaf().isWhiteSpaceWithNewline()
+        ) {
+            Unit
```
romtsn, line 754: hm, can we rather put an early-return here and in the if-block below?

romtsn (approved):

pvegh:
pls 4 minor release kthxbye

romtsn:
> kthxbye
why?

PR #1327, opened 2021-12-28
Add kotlin-logging as logging framework

### LINKED ISSUE #589: Separate console output and reporters
Currently ktlint uses plain reporter for console output. Better to implement console output as using some logging solution. This would allow more easier add/change output formatting (for example - color, see #582).

paul-dingemans:
## Description

Refactor output written to stdout / stderr to log statements where applicable.

The kotlin-logging framework uses SLF4J as underlying logging framework. The latter framework does not allow to change the desired log level to be changed at runtime. As ktlint supports the command line flag "--debug" to print additional logging, it was needed to add logback as underlying logging framework.

The IndentationRule and the RuleExtension class both changed the logging behavior based on the environment variable KTLINT_DEBUG. This variable could be assigned multiple values. This variable is now split into two distinct variables KTLINT_UNIT_TEST_TRACE and KTLINT_UNIT_TEST_DUMP_AST which can either be set to "on" to enable TRACE resp. DUMP_AST functionality or anything else to disable it. End users of ktlint should use the new command line flag "--trace" resp. the existing "--print-ast".

Solved a bug in DumpASTRule when retrieval of the linenumber results in a IndexOutOfBoundsException cause by an AST mutation resulting in an invalid offset.

In RuleExtension improved the logic regarding dumping the AST. When invoked from the extension, the DumpASTRule print the output to STDOUT instead of STDERR.

Following needs explicit attention while reviewing:
* Check the way the Gradle dependencies are added. I have little knowledge of Gradle.
* I was not able to find published checksums for the added dependencies, so I have added the Gradle generated keys

Closes #589 

## Checklist

- [ ] tests are added


nbadal:
FYI, as a consequence of this change, if any of these modules are used in a classpath with an existing logger (e.g. in an IDE), it will throw a `java.lang.LinkageError` unless slf4j is excluded as a transitive dependency:
```groovy
api("com.pinterest.ktlint:ktlint-core:0.45.1") {
    exclude("org.slf4j")
}
```
Just a gotcha for any other ktlint library consumers 😁

Issue #188, opened 2018-04-16
Place class/fun parameters onto the same line when below max_line_length


gabrielittner:
When you got something like the following ktlint should unwrap the parameters of test2, test3 and test4 because they fit in one line.

```
    fun test1(a: A, b: B) {
        // ...
    }

    fun test2(
        a: A,
        b: B
    ) {
        // ...
    }

    fun test3(a: A,
        b: B
    ) {
        // ...
    }

    fun test4(a: A, b: B
    ) {
        // ...
    }
```

The current behavior in 0.21.0 is that test1 and test2 stay like they are. test3 and test4 are reformatted to look like test2.

shyiko:
(somewhat) related to #251

paul-dingemans:
Code samples below are already fixed in a version prior to 0.43.2 
```
    fun test1(a: A, b: B) {
        // ...
    }

    fun test2(
        a: A,
        b: B
    ) {
        // ...
    }

    fun test3(a: A,
        b: B
    ) {
        // ...
    }

    fun test5(
    ) {}
```

Code sample below will be fixed by + PR #1331 
```
    fun test4(a: A, b: B
    ) {
        // ...
    }
```

paul-dingemans:
Will be covered by #1341 which is in progress. Completion is to be expected in Ktlint 0.46.0.

PR #1331, opened 2021-12-30
Add new experimental rule 'function-return-type-wrapping'

### LINKED ISSUE #941: [Android] Return type allowed on separate line
Ktlint in Android mode allows the return type of a function signature to be on a separate line when the parameters are not on their own lines.

The Android Style Guide says ([link](https://developer.android.com/kotlin/style-guide#functions)):

> When a function signature does not fit on a single line, break each parameter declaration onto its own line. Parameters defined in this format should use a single indent (+4). The closing parenthesis ()) and return type are placed on their own line with no additional indent.

**It seems like ktlint does not consider the return type to be part of the function signature.**

## Example / Steps to Reproduce

```
ktlint --android --stdin <<< '
fun thisismytestmethod(aaaaaaaaaaaaaaaa: String, bbbbbbbbbbbbbb: String, ccccccccccccccc, String):
    String {
        return "test"
    }'
```

## Expected Behavior

Ktlint should detect that the function signature does not fit on one line and give error(s) that the parameters are not on their own line.

## Observed Behavior

No error.

## Your Environment

* Operating System and version: macos 10.15.7

paul-dingemans:
## Description

In case of a multiline function signature (e.g. the signature already contains a newline) or in case of a single line function signature which exceeds the max-line-length, force the return type to be on the same line as the closing parenthesis. This can either be wrapping the closing parenthesis plus return type to a new line, or joining the line containing the closing parenthesis with the return type. See unit tests for examples.

This rule partly overlaps with the parameter-list-wrapping but does not conflict with it.

Closes #941, #188

## Checklist

- [X] tests are added

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ by passing the `--experimental` flag to `ktlint`.
 - No spaces around unary operators (id: `experimental:unary-op-spacing`)
 - Declarations with annotations should be separated by a blank line (id: `experimental:spacing-between-declarations-with-annotations`)
 - Declarations with comments should be separated by a blank line (id: `experimental:spacing-between-declarations-with-comments`)
+- In a multiline function signature the return type should be placed on the same line as the closing parenthesis (id: `experimental:function-return-type-wrapping`)
 
 
 ## EditorConfig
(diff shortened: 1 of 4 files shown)


paul-dingemans:
This rule will be integrated into the new rule to rewrite the entire function signature. See #1341

PR #1332, opened 2021-12-30
Display explanatory note about path after installing a git hook

### LINKED ISSUE #634: Pre-commit doesn't find ktlint file
the file `ktlint` is located at `/home/nick/workspace/MyProject/ktlint`
in pre commit :
```
git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs ktlint --relative .
```
But if I make commit, the fail message is:
```
Commit failed with error
0 files committed, 1 file failed to commit: MyCommitMessage
xargs: ktlint: No such file or directory
```

But when I change pre commit file to full path to ktlint file:
```
git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs /home/nick/workspace/MyProject/ktlint --relative .
```
Then commit work correctly. Why the error occurs in first case?

paul-dingemans:
## Description

It is not worth it to ensure that the git hook installation is completely fool proof. I have just added a small note which is displayed on successful install of the git hook:
```
Be aware that this hook assumes to find ktlint on the PATH. Either
ensure that ktlint is actually added to the path or expand the ktlint command in the hook with the path.
```

It is reasonable to expect from a developer that (s)he knows how to fix the path.

Closes #634

## Checklist

- [ ] tests are added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Fix indentation of secondary constructor (`indent`) ([#1222](https://github.com/pinterest/ktlint/issues/1222))
 - Fix alignment of arrow when trailing comma is missing in when entry (`trailing-comma`) ([#1312](https://github.com/pinterest/ktlint/issues/1312))
 - Fix indent of delegated super type entry (`indent`) ([#1210](https://github.com/pinterest/ktlint/issues/1210))
-- Display explanatory note about path after installing a git hook ([#634](https://github.com/pinterest/ktlint/issues/634))
 
 ### Changed
 - Update Kotlin version to `1.6.0` release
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Fix indentation of secondary constructor (`indent`) ([#1222](https://github.com/pinterest/ktlint/issues/1222))
 - Fix alignment of arrow when trailing comma is missing in when entry (`trailing-comma`) ([#1312](https://github.com/pinterest/ktlint/issues/1312))
 - Fix indent of delegated super type entry (`indent`) ([#1210](https://github.com/pinterest/ktlint/issues/1210))
+- Display explanatory note about path after installing a git hook ([#634](https://github.com/pinterest/ktlint/issues/634))
```
romtsn, line 21:
```suggestion
```
Since it's a chore, we can omit it from the changelog I think

PR #1336, opened 2022-01-13
Fix git hooks: do not run ktlint on all files of the current directory

francescocervone:
## Description

Ktlint should run on modified files only.

Fixes https://github.com/pinterest/ktlint/issues/1339


romtsn:
Thanks, this makes a lot of sense, am not sure how it wasn't caught before 🤦

PR #1338, opened 2022-01-20
Fix warn in gradle config ex. replace main > mainClass

h2000:
## Description

Fix for the example gradle config, because JavaExec main = "xxx" is deprecated and should be replaced with mainClass Property.


romtsn:
Thanks!

Issue #1195, opened 2021-08-05
ktlint fails on Java 16+


itsandreramon:
```
Exception in thread "main" java.util.concurrent.ExecutionException: java.lang.NoClassDefFoundError: Could not initialize class org.jetbrains.kotlin.com.intellij.pom.java.LanguageLevel
        at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122
        at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
        at com.pinterest.ktlint.KtlintCommandLine.parallel(Main.kt:575)
        at com.pinterest.ktlint.KtlintCommandLine.parallel$default(Main.kt:547)
        at com.pinterest.ktlint.KtlintCommandLine.lintFiles(Main.kt:298)
        at com.pinterest.ktlint.KtlintCommandLine.run(Main.kt:262)
        at com.pinterest.ktlint.Main.main(Main.kt:70)
Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.jetbrains.kotlin.com.intellij.pom.java.LanguageLevel
        at org.jetbrains.kotlin.com.intellij.core.CoreLanguageLevelProjectExtension.<init>(CoreLanguageLevelProjectExtension.java:26)
        at org.jetbrains.kotlin.com.intellij.core.JavaCoreProjectEnvironment.<init>(JavaCoreProjectEnvironment.java:42)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreProjectEnvironment.<init>(KotlinCoreProjectEnvironment.kt:26)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$ProjectEnvironment.<init>(KotlinCoreEnvironment.kt:121)
        at org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment$Companion.createForProduction(KotlinCoreEnvironment.kt:425)
        at com.pinterest.ktlint.core.internal.KotlinPsiFileFactory.initializePsiFileFactory(KotlinPsiFileFactory.kt:74)
        at com.pinterest.ktlint.core.internal.KotlinPsiFileFactory.acquirePsiFileFactory$ktlint_core(KotlinPsiFileFactory.kt:39)
        at com.pinterest.ktlint.core.KtLint.format(KtLint.kt:378)
        at com.pinterest.ktlint.internal.FileUtilsKt.formatFile(FileUtils.kt:182)
[...]

PR #1359, opened 2022-02-02
Fix for Java 16+ CLI

shashachu:
## Description

Conditionally add `--add-opens` flag based on java version

Fixes https://github.com/pinterest/ktlint/issues/1195

## Checklist

- [ x ] `CHANGELOG.md` is updated

### DIFF
--- a/ktlint/build.gradle
+++ b/ktlint/build.gradle
@@ -58,6 +58,13 @@ def shadowJarExecutableTask = tasks.register("shadowJarExecutable", DefaultTask.
   doLast {
     File execFile = outputs.files.getFiles().first()
     execFile.withOutputStream {
+      // From this SO answer: https://stackoverflow.com/a/56243046
+      // First we get the major Java version as an integer, e.g. 8, 11, 16. It has special handling for the leading 1
+      // of older java versions, e.g. 1.8 = Java 8
+      // JV = $(java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
+      // Then if that java version is >= 16, we add the --add-opens command
+      // X = X=$( [ "$JV" -ge "16" ] && echo "--add-opens java.base/java.lang=ALL-UNNAMED" || echo "")
+      // exec java $X -Xmx512m -jar "$0" "$@"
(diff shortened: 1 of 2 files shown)


ktlint/build.gradle
```diff
@@ -58,7 +58,7 @@ def shadowJarExecutableTask = tasks.register("shadowJarExecutable", DefaultTask.
   doLast {
     File execFile = outputs.files.getFiles().first()
     execFile.withOutputStream {
-      it.write "#!/bin/sh\n\nexec java -Xmx512m -jar \"\$0\" \"\$@\"\n\n".bytes
+      it.write "#!/bin/sh\n\nJV=\$(java -version 2>&1 | head -1 | cut -d'\"' -f2 | sed '/^1\\./s///' | cut -d'.' -f1)\n\nX=\$( [ \"\$JV\" -ge \"16\" ] && echo \"--add-opens java.base/java.lang=ALL-UNNAMED\" || echo \"\")\n\nexec java \$X -Xmx512m -jar \"\$0\" \"\$@\"\n\n".bytes
```
paul-dingemans, line 61: Maybe it is an idea to add some comments (and the bash script without espace characters) here.

```
#!/bin/sh
JV=$(java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\\./s///' | cut -d'.' -f1)
X=$( [ "$JV" -ge "16" ] && echo "--add-opens java.base/java.lang=ALL-UNNAMED" || echo "")
exec java $X -Xmx512m -jar "$0" "$@"
```

paul-dingemans (commented):
I think this will work. Unfortunately, I am not able to reproduce any of the Java16 issues anymore on my local machine with the current version of ktlint with java 16 nor 17. So I can not test whether your fix solve those problems.

ktlint/build.gradle
shashachu: That's a good idea. I'm able to repro and confirm this fixes it, but @romtsn pointed out this likely won't work on Windows. I don't have a Windows machine to test on.

paul-dingemans: Actually I had same doubt about windows. But the old code already uses a bash shell. The other utilities like cut and head are rather basic and most likely will work. But we won't know for sure untill tried. I have an old windows machine that I used for development years ago. There I used a tool called git-bash which supported all basic Unix tools.

shashachu: I've updated it with comments. I think what I'll do is merge it and then we can ask folks on the bug thread to test on Windows with the SNAPSHOT.

PR #1378, opened 2022-02-22
Add a generic type to ReporterProvider

Goooler:
## Description

Return subtypes when invoking `get()`

## Checklist

- [ ] tests are added


Goooler:
Just a cleanup, make sure invocations of `get` with proper types instead of casts, this change may need not tests?

PR #1395, opened 2022-03-05
Fix lint message and refactor

paul-dingemans:
## Description

Change message from "Unnecessary space(s)" to "Unnecessary long whitespace". The length of the whitespace element is considered while it is not checked whether the whitespace actually contains spaces only. However, it still make sense to replace a non-indent whitespace which is too long, by a single space.

## Checklist

- [X] tests are added


paul-dingemans:
Replaced with #1400

Issue #1421, opened 2022-03-18
Logger failes to initialize when not providing the Classic Logback Logger in 0.45.0 


ZacSweers:
## Expected Behavior

It should work

## Observed Behavior

This failure occurs (when using via Spotless)

```
Step 'ktlint' found problem in 'src/main/java/slack/appprofile/di/AppProfileModule.kt':
--
  | Could not initialize class com.pinterest.ktlint.core.internal.EditorConfigLoaderKt
  | java.lang.NoClassDefFoundError: Could not initialize class com.pinterest.ktlint.core.internal.EditorConfigLoaderKt
  | at com.pinterest.ktlint.core.internal.EditorConfigLoader.loadPropertiesForFile(EditorConfigLoader.kt:111)
  | at com.pinterest.ktlint.core.KtLint.prepareCodeForLinting(KtLint.kt:234)
  | at com.pinterest.ktlint.core.KtLint.format(KtLint.kt:344)
  | at com.pinterest.ktlint.core.KtLint.format(KtLint.kt:319)
  | at com.pinterest.ktlint.core.KtLint.format(KtLint.kt:315)
  | at com.diffplug.spotless.glue.ktlint.KtlintFormatterFunc.applyWithFile(KtlintFormatterFunc.java:60)
  | at com.diffplug.spotless.FormatterFunc$NeedsFile.apply(FormatterFunc.java:154)
  | at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:78)
  | at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:88)
  | at com.diffplug.spotless.Formatter.compute(Formatter.java:230)
  | at com.diffplug.spotless.PaddedCell.calculateDirtyState(PaddedCell.java:203)
  | at com.diffplug.spotless.PaddedCell.calculateDirtyState(PaddedCell.java:190)
  | at com.diffplug.gradle.spotless.SpotlessTaskImpl.processInputFile(SpotlessTaskImpl.java:92)
  | at com.diffplug.gradle.spotless.SpotlessTaskImpl.performAction(SpotlessTaskImpl.java:78)
```

## Steps to Reproduce

Use editorconfig and ktlint 0.45.0 in a project

## Your Environment

* Version of ktlint used: 0.45.0
* Relevant parts of the `.editorconfig` settings
```toml
[*]
charset=utf-8
end_of_line=lf
insert_final_newline=true
trim_trailing_whitespace=true
[...]

PR #1424, opened 2022-03-19
Remove logback specific configuration from ktlint-core package

z3d1k:
## Description
Removed logback dependency from `ktlint-core` module to fix [#1421](https://github.com/pinterest/ktlint/issues/1421)
- Logging level for ktlint cli is configured through `loggingModifier` function, which allows to perform arbitrary logic with logger object and executes after logger initialization.
- `logback-test.xml` file in `ktlint-test-logging` module to use in unit tests. Logging level configured providing environment property `KTLINT_UNIT_TEST_TRACE` with value `ON` (case insensitive)

Usage of `ktlint-core` package by external tools was successfully checked by updating ktlint version in [maven plugin](https://github.com/z3d1k/ktlint-maven-plugin/pom.xml)

## Checklist

- [x] PR description added
- [N/A] tests are added

In case of adding a new rule:
- [N/A] `README.md` is updated
- [N/A] Rule has been applied on Ktlint itself and violations are fixed


ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLintKLoggerInitializer.kt
```diff
@@ -1,58 +1,11 @@
...
-                    """
-                    System environment variable $KTLINT_DEBUG is no longer used to change the logging behavior while running unit tests.
-                    Now set one or more of environment variables below:
-                        $KTLINT_UNIT_TEST_TRACE=[on|off]
-                        $KTLINT_UNIT_TEST_DUMP_AST=[on|off]
-                    """.trimIndent()
-                }
-            }
-    }
+    also { logger -> loggerModifier(logger) }
```
Goooler, line 11:
```suggestion
public fun KLogger.initKtLintKLogger(): KLogger = also(loggerModifier)
```

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
```diff
@@ -299,6 +297,16 @@ class KtlintCommandLine {
         }
     }
 
+    private fun configureLogging() {
+        loggerModifier = { kLogger ->
+            (kLogger.underlyingLogger as Logger).level = when {
+                trace -> Level.TRACE
```
Goooler, line 303: If there need a cast check like this?
```kotlin
(kLogger.underlyingLogger as? Logger)?.level
```

paul-dingemans (commented):
Super nice solution and thanks again for your contribution. I did some tweaks on it an committed them on your branch. Will merge it to master,

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLintKLoggerInitializer.kt
paul-dingemans: Has been rewritten to:
```
public fun KLogger.initKtLintKLogger(
    loggerModifier: (KLogger) -> Unit = { _ -> }
): KLogger = apply { loggerModifier(this) }
```

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLintKLoggerInitializer.kt
```diff
@@ -1,58 +1,11 @@
...
-            ?.let {
-                logger.error {
-                    """
-                    System environment variable $KTLINT_DEBUG is no longer used to change the logging behavior while running unit tests.
-                    Now set one or more of environment variables below:
-                        $KTLINT_UNIT_TEST_TRACE=[on|off]
-                        $KTLINT_UNIT_TEST_DUMP_AST=[on|off]
-                    """.trimIndent()
-                }
-            }
```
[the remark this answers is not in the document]
paul-dingemans: Functionality is now re-added in a different (and improved) form in the `ktlint-test`. Now, a message is printed in each unit test that trace information and the AST_DUMP can be enabled by setting environment variables.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLintKLoggerInitializer.kt
```diff
@@ -1,58 +1,11 @@
...
 
-// Via command line parameters "--trace" and "--print-ast" the end user of ktlint can change the logging behavior. As
-// unit tests are not invoked via the main ktlint runtime, those command line parameters can not be used to change the
-// logging behavior while running the unit tests. Instead, the environment variables below can be used by ktlint
-// developers to change the logging behavior. Note, when set the environment variables also affect the runtinme of
-// ktlint. As of that the name of the variables start with KTLINT_UNIT_TEST to clarify the intent.
-public const val KTLINT_UNIT_TEST_DUMP_AST = "KTLINT_UNIT_TEST_DUMP_AST"
-public const val KTLINT_UNIT_TEST_TRACE = "KTLINT_UNIT_TEST_TRACE"
-public const val KTLINT_UNIT_TEST_ON_PROPERTY = "ON"
+public var loggerModifier: (KLogger) -> Unit = { _ -> }
```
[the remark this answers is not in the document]
z3d1k: The solution where `loggerModifier` is passed as a parameter looks much cleaner, I agree.
However, the difference between these solutions is that if a modifier function is passed as a parameter - modification will be applied only to the current logger (one created by this call), but all others will still be created with the default logging level. If modifier represented by a variable - all loggers that will be created using `initKtLintKLogger` will also use the modifier method for configuration after creation.
In the general case, it is not a very good idea to make such a global configuration, but here it is justifiable in my opinion.

ktlint/src/main/kotlin/com/pinterest/ktlint/Main.kt
z3d1k: A more strict cast will allow "fail fast" behaviour and will be instantly visible if anyone would try to change underlying logger implementation for `ktlint` module. Otherwise, it can lead to an annoying problem where passing parameter just does not do anything since casting error is silently swallowed.

ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLintKLoggerInitializer.kt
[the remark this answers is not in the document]
paul-dingemans: Good catch. I forgot about this case.  I have now changed it so that:
* The default loggerModifier can only be set once.
* It is still possible to override the loggerModifier for a specific class by passing a loggerModifier to `initKtLintKLogger`.
* When `initKtLintKLogger` is called without parameter, the defaultLoggerModifier is applied when set.

```
/**
 * Default modifier for the KLogger of new instances of classes calling [initKtLintKLogger]. Classes for which
 * [initKtLintKLogger] has been called before setting this variable will not be changed. Also note, that this modifier
 * can only be set once.
 */
public lateinit var defaultLoggerModifier: (KLogger) -> Unit

/**
 * Initializes the logger with the [defaultLoggerModifier] when set.
 */
public fun KLogger.initKtLintKLogger(): KLogger {
    return if (::defaultLoggerModifier.isInitialized) {
        apply { defaultLoggerModifier(this) }
    } else {
        this
    }
}

/**
 * Initializes the logger with the [loggerModifier].
 */
public fun KLogger.initKtLintKLogger(
    loggerModifier: (KLogger) -> Unit
): KLogger = apply { loggerModifier(this) }
```

PR #1427, opened 2022-03-20
Allow to set a default logger modifier

paul-dingemans:
## Description

* The default loggerModifier can only be set once.
* It is still possible to override the loggerModifier for a specific class by passing a loggerModifier to initKtLintKLogger.
* When initKtLintKLogger is called without parameter, the defaultLoggerModifier is applied when set.

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:
- [ ] `README.md` is updated

### DIFF
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLintKLoggerInitializer.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLintKLoggerInitializer.kt
@@ -12,8 +12,26 @@ public const val KTLINT_UNIT_TEST_DUMP_AST = "KTLINT_UNIT_TEST_DUMP_AST"
 public const val KTLINT_UNIT_TEST_ON_PROPERTY = "ON"
 
 /**
- * Initializes the logger. Optionally the logger can be modified using the [loggerModifier].
+ * Default modifier for the KLogger of new instances of classes calling [initKtLintKLogger]. Classes for which
+ * [initKtLintKLogger] has been called before setting this variable will not be changed. Also note, that this modifier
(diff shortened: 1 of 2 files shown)


ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLintKLoggerInitializer.kt
```diff
@@ -12,8 +12,26 @@ public const val KTLINT_UNIT_TEST_DUMP_AST = "KTLINT_UNIT_TEST_DUMP_AST"
 public const val KTLINT_UNIT_TEST_ON_PROPERTY = "ON"
 
 /**
- * Initializes the logger. Optionally the logger can be modified using the [loggerModifier].
+ * Default modifier for the KLogger of new instances of classes calling [initKtLintKLogger]. Classes for which
+ * [initKtLintKLogger] has been called before setting this variable will not be changed. Also note, that this modifier
+ * can only be set once.
+ */
+public lateinit var defaultLoggerModifier: (KLogger) -> Unit
```
z3d1k, line 19: > Also note, that this modifier can only be set once.
Are there any checks for that? `lateinit` only checks whether the variable has been initialized before it is accessed, no additional checks on its modification are performed.

paul-dingemans: Tnx again. That was a rookie mistake. I have improved it in #1430. I can not prevent the variable from being set more than once as it will break some integration tests in ktlint. I have now succeeded to print a warning (to discourage doing it) whenever it happens.

PR #1442, opened 2022-04-02
Fix backward incompatibility issues released via KtLint 0.45.0 and 0.45.1

### LINKED ISSUE #1434: `0.45.0` backwards incompatibility: `KtLint.Params#userData` gets ignored, unclear migration path + more feedback
Initially reported on Kotlin slack: https://kotlinlang.slack.com/archives/CKS3XG0LS/p1647897839331439 but I was asked to create issue here. 

## Expected Behavior

`0.45.0` is backward compatible with `0.44.0`, or has clear migration path

## Observed Behavior

I'll copy-paste what I wrote on Slack:
1. Can someone share what’s the future of `userData` map, does it only change its purpose, or will it be removed completely at some point? The code I was upgrading had been passing `indent_size` and `disabled_rules` within `userData` map. The migration path for `indent_size` seemed to be hidden but I was able to identify `DefaultEditorConfigProperties.indentSizeProperty` which looked legitimate and worked in my case, but I couldn’t find a replacement for `disabled_rules`. The fact `ExperimentalParams` still takes `userData` suggests that’s still the _“recommended”_ option, am I wrong here? What’s the proper way to pass rules that I expect to be disabled?
2.  It seems like there is a behavior change introduced in `0.45.0`, where values passed as `userData` using the now deprecated `lint` method are ignored. Maybe I misread the release notes, but they had vibe you _strongly advise_ to migrate now, because later it will be more difficult, but in overall `0.45.0` is still backwards compatible. If breaking the compatibility was intended, then I’ll share my personal opinion: I think it would be better for api consumers if you removed the method already in this release so api consumers would get a compile-time error, instead of runtime one (or bug reports from end users 😛). If breaking the compatibility wasn’t intended, then FYI it did, at least the case described above 😛 
And just for context: I attempted to bump ktlint in [Gradle Plugin](https://github.com/jeremymailen/kotlinter-gradle/pull/243), so I might be missing something here, as the release notes weren’t 100% clear for me (for example: I’ve never seen `ASTNode.getEditorConfigValue(editorConfigProperty)` or `EDITOR_CONFIG_USER_DATA_KEY` mentioned there 😬 )
3. I noticed `isUnitTestContext = true` gets passed when calling any of `KtLint#lint` methods. It doesn’t break anything, but looks _suspicious_, so I thought it might be worth pointing out, especially when seeing todos like this 👇
AFAIU both [ktlint-gradle](https://github.com/JLLeitschuh/ktlint-gradle/blob/ba42cdc3ab8ab7c2454c6e1a44696f1718db43ea/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/worker/KtLintWorkAction.kt#L75) and [kotlinter-gradle](https://github.com/jeremymailen/kotlinter-gradle/blob/018d1c35a4b57a42c452915a3154e12c71444405/src/main/kotlin/org/jmailen/gradle/kotlinter/tasks/lint/LintWorkerAction.kt#L78) use lint and format in regular usage, not in tests soo… 👀 
image

Feel free to split them into 3 separate issues if it's more convenient for you :) 

## Steps to Reproduce

Run `KtLint.lint(KtLint.Params(userData = mapOf("indent_size" to "5"), ...))` on a file with respective indentation and observe the lint reporting failures.

## Your Environment

* Version of ktlint used: 0.45.1
* Relevant parts of the `.editorconfig` settings: none
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): I tried to upgrade kotlinter-gradle and the tests failed: https://github.com/jeremymailen/kotlinter-gradle/pull/243/files I had to switch to the `ExperimentalParams` class which introduced even more confusion - hence all the questions above
* Version of Gradle used (if applicable): it's not
* Operating System and version: `macos-latest`, `windows-latest`

paul-dingemans:
## Description

* Deprecate data class 'Params' as it only provides the userData parameter  which prohibits API Consumers to split the '.editorconfig' properties  from the other user data.
* Deprecate function 'toExperimentalParams' to force API Consumers to  start using the 'ExperimentalParams'.
* In ExperimentalParams clarify that 'userData' should not contain any  '.editorconfig' properties. Now also a runtime error is thrown  whenever the lint/format methods are called and 'userData' contains  '.editorconfig' properties.
* Remove confusing TODO's about moving methods 'lint' and 'format' to  the 'ktlint-test' module.
* Update deprecation messages to clarify intention.
* The 'lint' and 'format' methods in module 'ktlint-core' no longer  call the 'VisitorProvider' with parameter 'isUnitTestContext'  enabled. API Consumers call these methods with production code  for which this parameter should be disabled.
* The RuleExtension calls the 'lint' and 'format' methods but do  provide a VisitorProvider for which the parameter  'isUnitTestContext' is enabled.
* Deprecate the 'format' which accepted the 'ExperimentalParams'  and 'Iterable<RuleSet>' while the latter is already included  in the first.
* Add new 'format' method which accepts 'ExperimentalParams' and 'VisitorProvider' only.
* Move 'EditorConfigOverride' from 'ktlint-test' to 'ktlint-core'

Closes #1434

## Checklist

- [X] PR description added
- [X] tests are added

In case of adding a new rule:
- [ ] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Fixed
+- Resolve compatibility issues introduced in 0.45.0 and 045.1 ([#1434](https://github.com/pinterest/ktlint/issues/1434))
 
 ### Changed
 * Set Kotlin development version to `1.6.20-RC2` and Kotlin version to `1.6.10`.
--- a/ktlint-core/build.gradle
+++ b/ktlint-core/build.gradle
@@ -7,6 +7,7 @@ dependencies {
   api deps.kotlin.compiler
   api deps.ec4j
   api deps.logging
+  api deps.assertj
 
   // Standard ruleset is required for EditConfigLoaderTest only
   testImplementation projects.ktlintRulesetStandard
--- a/ktlint-core/build.gradle
+++ b/ktlint-core/build.gradle
@@ -7,7 +7,6 @@ dependencies {
   api deps.kotlin.compiler
   api deps.ec4j
   api deps.logging
-  api deps.assertj
 
   // Standard ruleset is required for EditConfigLoaderTest only
   testImplementation projects.ktlintRulesetStandard
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
@@ -1,12 +1,14 @@
 package com.pinterest.ktlint.core
 
+import com.pinterest.ktlint.core.api.DefaultEditorConfigProperties
+import com.pinterest.ktlint.core.api.EditorConfigOverride
+import com.pinterest.ktlint.core.api.EditorConfigOverride.Companion.emptyEditorConfigOverride
 import com.pinterest.ktlint.core.api.EditorConfigProperties
 import com.pinterest.ktlint.core.api.FeatureInAlphaState
 import com.pinterest.ktlint.core.api.UsesEditorConfigProperties
 import com.pinterest.ktlint.core.internal.EditorConfigGenerator
 import com.pinterest.ktlint.core.internal.EditorConfigLoader
 import com.pinterest.ktlint.core.internal.EditorConfigLoader.Companion.convertToRawValues
-import com.pinterest.ktlint.core.internal.EditorConfigOverridesMap
 import com.pinterest.ktlint.core.internal.KotlinPsiFileFactoryProvider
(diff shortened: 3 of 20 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Fixed
+- Resolve compatibility issues introduced in 0.45.0 and 045.1 ([#1434](https://github.com/pinterest/ktlint/issues/1434))
```
Goooler, line 10:
```suggestion
- Resolve compatibility issues introduced in 0.45.0 and 0.45.1 ([#1434](https://github.com/pinterest/ktlint/issues/1434))
```

PR #1460, opened 2022-05-05
Run rules during unit tests in correct order

paul-dingemans:
## Description

With KtLintAssertThat an AssertJ style `assertThat` is created for a specific rule. In addition to that rule, it is possible to run additional rules. In case the rule which is to be tested has defined a VisitorModifier which requires one or more additional rules to be loaded and to be enabled, it is even mandatory that those rules are added to the unit test.

The goal is to execute the rules during unit tests in the same order as when running the CLI version of KtLint. During each unit test a dynamic RuleSet is being created for a limited set of rules (e.g. the rule for which the `assertThat` is created plus the additional rules specified in the unit tests). The rules in this minimized ruleSet are executed in the order as defined by the VisitorModifier as defined in the rules.

* Split naming policy of rule id and rule set id. The naming policy of the latter is not changed. The naming policy of the ruleId is changed so that the ruleSetId prefix can be specified optionally. If the ruleId is not prefixed with a ruleSetId then it is assumed to be equal to "standard". For all experimental rules, the ruleSetId has been added.
* Remove obsolete parameter "isUnitTestContext" from VisitorProvider
* As the VisitorModifiers are now also used and checked during unit tests, it is required to run the additional rules during the lint phase as well. Some code examples in tests in which the IndentationRule is added as additional rule are changed to comply with that rule.

## Checklist

- [X] PR description added
- [X] tests are added

In case of adding a new rule:
- [ ] `README.md` is updated


paul-dingemans:
Closed as due to incorrect merge with master the squashed commit would contain too much changes. Replaced with PR #1470

PR #1105, opened 2021-03-11
1077 needless line dot qualified expression

paul-dingemans:
## Description

Lint/format needless blank lines in dot qualified expression

## Checklist

- [X] tests are added
- [X] `CHANGELOG.md` is updated

PR #1469, opened 2022-05-16
Allow a single whitespace line in chained expressions

### LINKED ISSUE #1248: Consecutive blank lines are reported in chained method calls
## Expected Behavior

Comments within chained method calls should be able to have a new line before them.

## Observed Behavior

The line with the comment is marked as problematic

     Needless blank line(s) (no-consecutive-blank-lines)

## Steps to Reproduce

Example:
```
 // when
 webTestClient
    .get()
    .uri(requestUri)
    .exchange()

    // then
    .expectStatus().isNotFound
```

## Your Environment

* Version of ktlint used: 0.42.1
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): Gradle with org.jlleitschuh.gradle.ktlint:10.2.0
* Version of Gradle used (if applicable): 7.2
* Operating System and version: Mac OS / Linux

seadowg:
## Description

Closes #1248

There's a lot of discussion in the issues as to the right approach, but here's my proposal for a fix. Given the rule is intended to detect "consecutive" blank lines, I've made sure a single blank line is allowed in a chained expression (but that multiple is still disallowed and formats correctly). This will allow for the "blank line followed by a comment" examples brought up in the issue and keeps the whitespace rules consistent between regular statements and chained expressions.

The reasoning behind arguing for blank lines to be allowed in chained expressions is that there's nothing about disallowing them in either the [Kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html) or the [Android Kotlin style guide](https://developer.android.com/kotlin/style-guide) (as far as I can see) and people have brought up use cases where it's useful in the issue.

## Checklist

- [x] PR description added
- [x] tests are added

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoBlankLinesInChainedMethodCallsRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoBlankLinesInChainedMethodCallsRule.kt
@@ -4,6 +4,7 @@ import com.pinterest.ktlint.core.Rule
 import com.pinterest.ktlint.core.ast.ElementType.DOT_QUALIFIED_EXPRESSION
 import org.jetbrains.kotlin.com.intellij.lang.ASTNode
 import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
+import org.jetbrains.kotlin.com.intellij.psi.impl.source.tree.LeafPsiElement
 
 public class NoBlankLinesInChainedMethodCallsRule : Rule("no-blank-lines-in-chained-method-calls") {
     override fun visit(
@@ -14,6 +15,7 @@ public class NoBlankLinesInChainedMethodCallsRule : Rule("no-blank-lines-in-chai
         val isBlankLine = node is PsiWhiteSpace && node.getText().contains("\n\n")
         if (isBlankLine && node.treeParent.elementType == DOT_QUALIFIED_EXPRESSION) {
             emit(node.startOffset + 1, "Needless blank line(s)", true)
+            (node as LeafPsiElement).rawReplaceWithText("\n" + node.getText().split("\n\n")[1])
         }
     }
 }
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoBlankLinesInChainedMethodCallsRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoBlankLinesInChainedMethodCallsRule.kt
@@ -1,0 +1,13 @@
+package com.pinterest.ktlint.ruleset.standard
+
+import com.pinterest.ktlint.core.Rule
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+
+public class NoBlankLinesInChainedMethodCallsRule : Rule("no-blank-lines-in-chained-method-calls") {
+    override fun visit(
+        node: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit
+    ) {
+    }
+}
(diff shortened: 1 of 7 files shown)


paul-dingemans:
@seadowg Tnx for making the effort for this PR.

Basically you have reverted issue #1105 without offering an alternative solution for that issue. I do not think that we should do so as that functionality is expected to be in place by other users who do not want allow single blank line within dot qualified expression. 

So I think that your PR has to be expanded with a new rule that just disallows (needless) blank lines in a dot qualified expression. This new rule should be a simplified version of the current rule. To allow blank lines in a dot qualified expression you will need to disable that additional rule in your projects.

seadowg:
> So I think that your PR has to be expanded with a new rule that just disallows (needless) blank lines in a dot qualified expression. This new rule should be a simplified version of the current rule. To allow blank lines in a dot qualified expression you will need to disable that additional rule in your projects.

Sounds good! For the name of the new rule, how about `no-blank-lines-in-chain`? That's what pops into my head after looking at the language used in `chain-wrapping` and `no-consecutive-blank-lines`.

paul-dingemans:
> Sounds good! For the name of the new rule, how about `no-blank-lines-in-chain`? That's what pops into my head after looking at the language used in `chain-wrapping` and `no-consecutive-blank-lines`.

I prefer the longer version  `no-blank-lines-in-chained-method-calls`.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoBlankLinesInChainedMethodCallsRule.kt
```diff
@@ -15,7 +15,10 @@ public class NoBlankLinesInChainedMethodCallsRule : Rule("no-blank-lines-in-chai
         val isBlankLine = node is PsiWhiteSpace && node.getText().contains("\n\n")
         if (isBlankLine && node.treeParent.elementType == DOT_QUALIFIED_EXPRESSION) {
             emit(node.startOffset + 1, "Needless blank line(s)", true)
-            (node as LeafPsiElement).rawReplaceWithText("\n" + node.getText().split("\n\n")[1])
+
+            if (autoCorrect) {
```
seadowg, line 19: I assume we need this check, but I couldn't find a good way to write a test that drives it out. Help would be appreciated!

seadowg:
@paul-dingemans this is probably ready for another look.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/NoBlankLinesInChainedMethodCallsRule.kt
paul-dingemans: If you mean `if (autoCorrect) {` then you have already done so in test case `single blank line in dot qualified expression returns lint error`. 

When calling `hasLintViolations`, the rule engine is executed with `autocorrect` set to `false` which results in emitting the error without fixing it. When calling 'isFormattedAs`, the rule engine is invoked with `autocorrect` set to `true` resulting in formattting the code.

```
        noBlankLinesInChainedMethodCallsRuleAssertThat(code)
            .hasLintViolations(LintViolation(3, 1, "Needless blank line(s)"))
            .isFormattedAs(
                """
                fun foo(inputText: String) {
                    inputText
                        .toLowerCase()
                }
                """.trimIndent()
            )
            ```

seadowg: Ah right. But, if I remove the `if` statement (and just always remove the blank line), I don't see a failure. I'm assuming we don't want to be messing around with the code when `autocorrect` is `false`. Maybe we need to add a check to `hasLintViolations` to check that the code is unchanged?

seadowg: @paul-dingemans thanks for merging, but still a little concerned there's no coverage for the `if (autocorrect)` check. I'd be happy to look at that in the future if you agree it's an issue?

seadowg: Sorry, being more specific, I mean that it doesn't have "branch" coverage - it does indeed have "line" coverage. If I change the `if` statement to just be `if (true)` for example (or just remove the `if` and keep the `rawReplaceWithText`), the tests will still all pass. There's no test that ensures that we only run reformat when `autoCorrect` is `true`.

paul-dingemans: You don't have to worry about that. There is not a single rule for which this is being tested.

PR #1472, opened 2022-05-20
Respect git hooksPath setting when installing a git hook

### LINKED ISSUE #1465: Feature Request: installGitPreCommitHook into core.hooksPath
## Expected Behavior

As a feature request, instead of hardcoding into this directory, also check if the ` core.hooksPath` exists and create the hook file into this path instead, if it is set. See https://git-scm.com/docs/githooks

## Current Behavior

Currently, it is hardcoded to `.git/hooks` so we have to manually copy-paste the generated file into the `core.hooksPath`

paul-dingemans:
## Description

Closes #1465

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:
- [ ] `README.md` is updated


ktlint/src/main/kotlin/com/pinterest/ktlint/internal/GitHookInstaller.kt
```diff
@@ -56,12 +70,18 @@ object GitHookInstaller {
...
 
-        return hooksDir
+    private fun getHooksDirName() = try {
+        Runtime.getRuntime().exec("git config --get core.hooksPath")
+            .inputStream
+            .bufferedReader()
+            .readText()
+            .trim()
+            .ifEmpty { DEFAULT_GIT_HOOKS_DIR }
+    } catch (ex: IOException) {
```
Goooler, line 83:
```suggestion
    } catch (_: IOException) {
```

Goooler: 5fb47621ef5d8abce4edcbea7213437e1b3ab70a

PR #1480, opened 2022-05-29
Pre-commit don't runt ktlint if nothing changed

spacecowboy:
If no kotlin files has changed, then it shouldn't run ktlint. As in the case where you just edited a README for example

## Description

This changes the pre-commit hook so that it only runs ktlint if any kotlin files have been changed.

Existing behavior will run ktlint on all files in the event that no kotlin files were present in the changed files. For example, if the commit only changes a README file.

## Checklist

- [x] PR description added

### DIFF
--- a/ktlint/src/main/resources/ktlint-git-pre-commit-hook.sh
+++ b/ktlint/src/main/resources/ktlint-git-pre-commit-hook.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 # https://github.com/pinterest/ktlint pre-commit hook
-git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs ktlint --relative
+git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs --no-run-if-empty ktlint --relative
 if [ $? -ne 0 ]; then exit 1; fi


ktlint/src/main/resources/ktlint-git-pre-commit-hook.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 # https://github.com/pinterest/ktlint pre-commit hook
-git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs ktlint --relative
+git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs --no-run-if-empty ktlint --relative
```
paul-dingemans, line 3: Cool change. I was not aware of that xargs option. Same needs to be done for the other hooks to keep it all consistent.

paul-dingemans: Unfortunately, this won't work for example on MacOs. The "flag --no-run-if-empty" is not recognized (see https://stackoverflow.com/questions/8296710/how-to-ignore-xargs-commands-if-stdin-input-is-empty) as balnk inut is ignored by default. The change therefore can not be merged as it is right now.

ntr-ross: In order to keep things as 'sh'-oriented as possible (i.e. no bash embellishments) may I suggest something like:
```suggestion
[ `uname -s` != "Darwin" ] && no_run=--no-run-if-empty
git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs $no_run ktlint --relative
```

Or another option which would cater for other older/bsd-style `xargs` would be:
```suggestion
`xargs -r </dev/null >/dev/null 2>&1` && no_run=--no-run-if-empty
git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs $no_run ktlint --relative
```
but this seems really verbose and possibly even over-kill.

Just a thought.

paul-dingemans: I am not sure whether it will work on all OS but at least it does work on MacOs ;-)

I would prefer the first alternative as it is more readable but I would like a clarifying comment above the uname command to explain that this is MacOS specific.

The same trick should be applied to all other hooks as well. Please search on "xargs" to find them all.

spacecowboy: First suggestion looks best IMO too.

Just tweaked it to adhere to ShellCheck

spacecowboy:
> Yeah, the reason for using back-ticks instead of `$(...)` was to make it as backward-compatible as possible (since the file interpreter is `#!/bin/sh`).
> 
> Also, in order to be as defensive as possible, I guess it should probably be something like:
> 
> ```
> [ "x`...`" != "xDarwin" ] ... 
> ```
> 
> just in case a system type ever has a leading hyphen. astonished (Not likely I know.)

I seriously doubt that any shells that old are in wide use.But if you feel strongly about it I can change it.

And guarding against leading hyphens just seem like Voodoo programming. Let the person with that person come with their own PR.

ntr-ross:
> I seriously doubt that any shells that old are in wide use.But if you feel strongly about it I can change it.

Oh, no strong opinion here.  I was just keying off the defined shebang interpreter.  I didn't know how far back compatibility was supposed to go.  If you are fine with '$-notation' then so am I. 💯 

> And guarding against leading hyphens just seem like Voodoo programming. Let the person with that person come with their own PR.

I hear yah.  It's certainly justified for unstructured data but OS names really should not be an issue.

👍

paul-dingemans:
@spacecowboy and @ntr-ross Tnx for your contributions! The change is merged.

PR #1484, opened 2022-06-02
Code cleanups

Goooler:
## Description

Follow up #1482 & #1479

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:
- [ ] `README.md` is updated


ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
```diff
@@ -163,7 +162,6 @@ internal fun File.location(
 /**
  * Run lint over common kotlin file or kotlin script file.
  */
-@OptIn(FeatureInAlphaState::class)
```
paul-dingemans, line 166: Cool that you removed those. I still had that on my list. I found some more, which I have removed and added to this PR.

PR #1489, opened 2022-06-08
Function signature body

paul-dingemans:
## Description

By default the function signature rule kept the first line of a body expression on the same line as the function signature if the maximum line length is not exceeded (per Kotlin styleguide). This is very readable for multine function signature in which the return type has a short name or is omitted:
```
//                                                                             assume max line lenght is here ->
fun aVeryLoooooooooooooooooooooooongFunctionName(
     // some parameters causing the function signature to long for single line
) = listOf(foo())
     .filter { ... }
```

But in other case this can also result in more difficult to read code:
```
fun aVeryLoooooooooooooooooooooooongFunctionName() = listOf(foo())
     .filter { ... }
```

The `.editorconfig` parameter `ktlint_function_signature_body_expression_wrapping` can be used to determine how the expression body should be wrapped:
* default: Keep the first line of the body expression on the same line as the function signature if max line length is not exceeded.
* multiline: Force the body expression to start on a separate line in case it is a multiline expression. A single line body expression is wrapped only when it does not fit on the same line as the function signature.
* always: Always force the body expression to start on a separate line.

## Checklist

- [X] PR description added
- [X] tests are added

In case of adding a new rule:
- [ ] `README.md` is updated

### DIFF
--- a/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/FunctionSignatureRule.kt
+++ b/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/FunctionSignatureRule.kt
@@ -22,6 +22,7 @@ import com.pinterest.ktlint.core.ast.ElementType.WHITE_SPACE
 import com.pinterest.ktlint.core.ast.children
 import com.pinterest.ktlint.core.ast.isRoot
 import com.pinterest.ktlint.core.ast.isWhiteSpace
+import com.pinterest.ktlint.core.ast.isWhiteSpaceWithNewline
 import com.pinterest.ktlint.core.ast.lineIndent
 import com.pinterest.ktlint.core.ast.nextCodeLeaf
 import com.pinterest.ktlint.core.ast.nextCodeSibling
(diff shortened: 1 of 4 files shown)


paul-dingemans:
@Goooler Can you please help with the failing windows build? It keeps failing due to insufficient memory.

```
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 32744 bytes for ChunkPool::allocate
org.gradle.internal.remote.internal.MessageIOException: Could not write '/127.0.0.1:55310'.
# An error report file with more information is saved as:
	at org.gradle.internal.remote.internal.inet.SocketConnection.flush(SocketConnection.java:140)
# D:\a\ktlint\ktlint\ktlint-ruleset-experimental\hs_err_pid4748.log
	at org.gradle.internal.remote.internal.hub.MessageHub$ConnectionDispatch.run(MessageHub.java:333)
#
```

paul-dingemans:
Closed in favour of #1498. I have reapplied all commits one by one on a new branch but skipped the intermediate merge of the master branch (https://github.com/pinterest/ktlint/pull/1489/commits/9efaf60dcf931f85b31d0cadce943dbb6614e520). This apparently fixed the failing build on Windows.

PR #1501, opened 2022-06-18
Setup publication of documentation with mkdocs-material to github pages

paul-dingemans:
## Description

Replace old github pages documentation with mkdocs-material. This PR only configuration the build pipeline and replaces the old documentation with a temporary placeholder. Once it is verified that the pipeline works, we can start writing the new documentation.

Most likely, some additional steps are needed:
* [Configure the branch](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) for which admin settings are required
* Disabling publication of old site

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:
- [ ] `README.md` is updated

### DIFF
--- a/.github/workflows/gradle-docs-publish.yml
+++ b/.github/workflows/gradle-docs-publish.yml
@@ -1,0 +1,15 @@
+name: Publish documentation
+on:
+  push:
+    branches:
+      - master
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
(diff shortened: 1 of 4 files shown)


shashachu:
I've configured the repo to publish from the /docs folder.

paul-dingemans:
@shashachu It looks like that the configuration is not correct. I think you should publish from the `ktlint/doc` instead of `ktlint/doc/docs` folder as it can not find file `mkdocs.yml` in https://github.com/pinterest/ktlint/runs/6955531486?check_suite_focus=true

PR #1503, opened 2022-06-19
temporarily revert root build.gradle back to groovy

shashachu:
The publish task is not working and I'm not familiar enough with the Kotlin DSL to immediately debug.


Goooler:
Could you please paste the log here? @shashachu

PR #1506, opened 2022-06-19
Use stable Kotlin compiler frontend

3flex:
## Description
Using K2 means library consumers must also use the new compiler (and may not be ready to), otherwise they will have errors like "\<classname> is compiled by the new Kotlin compiler frontend and cannot be loaded by the old compiler"

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:
- [ ] `README.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+* Remove experimental flag `-Xuse-k2` as it forces API Consumers to compile their projects with this same flag ([#1506](https://github.com/pinterest/ktlint/issues/1506)).
+
 ### Changed
 
 ### Removed
--- a/buildSrc/src/main/kotlin/ktlint-kotlin-common.gradle.kts
+++ b/buildSrc/src/main/kotlin/ktlint-kotlin-common.gradle.kts
@@ -19,11 +19,4 @@ kotlin {
     }
 }
 
-tasks.withType<KotlinCompile>().configureEach {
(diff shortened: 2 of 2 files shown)


paul-dingemans:
Sorry, that was not intended. We will revert that change this week. @Goooler

3flex:
If by "revert" you mean downgrading from Kotlin 1.7, that shouldn't be necessary. It just needs to stop using the K2 compiler mode.

Goooler (approved):

Goooler:
Should we note this change in the changelog?

paul-dingemans:
> Should we note this change in the changelog?

yes

> Should we note this change in the changelog?

Yes and done so.

shashachu:
thanks!

akuleshov7:
@3flex thank you for this PR - yesterday we spent the whole day for trying to stabilise the build on K2 😂 in [diktat](https://github.com/saveourtool/diktat/pull/1379). We were afraid of @paul-dingemans and  there was such a fear that we decided not even to report it to ktlint 😂😂😂

paul-dingemans:
That sounds a bit rough to say. You should never fear the conversation with me. We all make mistakes. As open source developers we are all dedicated to make the world better even if we sometimes have different opinions.

akuleshov7:
> That sounds a bit rough to say. You should never fear the conversation with me. We all make mistakes. As open source developers we are all dedicated to make the world better even if we sometimes have different opinions.

I’m just kidding 🙏 just an inside joke related to our last arguing about RuleSetProvider 😁

PR #1544, opened 2022-07-13
use sed entirely with regex to get the java major version

### LINKED ISSUE #1543: ktlint prints an error when running with JAVA_TOOL_OPTIONS
## Expected Behavior

running ktlint with JAVA_TOOL_OPTIONS set should not produce an error

## Observed Behavior

running ktlint with JAVA_TOOL_OPTIONS produces the following error
```
/usr/local/Cellar/ktlint/0.46.1/libexec/ktlint: line 5: [: Picked up JAVA_TOOL_OPTIONS: integer expression expected
```

## Steps to Reproduce

download ktlint
export some JAVA_TOOL_OPTIONS
run ktlint

## Your Environment

* Version of ktlint used: 0.46.1
* Relevant parts of the `.editorconfig` settings: n/a
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): n/a
* Version of Gradle used (if applicable): n/a
* Operating System and version: macos 12.3
* i have JAVA_TOOL_OPTIONS set

## Investigation
This seems to be happening from these two lines in the ktlint start
```
# incorrectly retrieves version
JV=$(java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
echo $JV
# echoes 'Picked up JAVA_TOOL_OPTIONS:'
X=$( [ "$JV" -ge "16" ] && echo "--add-opens java.base/java.lang=ALL-UNNAMED" || echo "")
# this command fails because the version is not correctly parsed
```

Ktlint still runs okay, it just spits some errors in the beginning. This could be solved with a different way to get the java version. Perhaps something with sed or awk to search for the 'version' string instead of head.

rawilder:
## Description

Fix java version being incorrectly set when JAVA_TOOL_OPTIONS has a value. It being set causes unexpected output for the current java version parsing method.

#### Investigation
This seems to be happening from these two lines in the ktlint start
```
# incorrectly retrieves version
JV=$(java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
echo $JV
# echoes 'Picked up JAVA_TOOL_OPTIONS:'
X=$( [ "$JV" -ge "16" ] && echo "--add-opens java.base/java.lang=ALL-UNNAMED" || echo "")
# this command fails because the version is not correctly parsed
```

Ktlint still runs okay, it just spits some errors in the beginning. This could be solved with a different way to get the java version. Perhaps something with sed or awk to search for the 'version' string instead of head. 

#### Local test results
A couple of tests failed locally but the same tests failed on master for me locally.
```
CLI basic checks > Given CLI argument --help then return the help output() FAILED
    org.opentest4j.AssertionFailedError at SimpleCLITest.kt:14

CLI basic checks > Given CLI argument --version then return the version information output() FAILED
    org.opentest4j.AssertionFailedError at SimpleCLITest.kt:29

CLI basic checks > Given some code without errors then return from lint with normal exit code and no error output() FAILED
    org.opentest4j.AssertionFailedError at SimpleCLITest.kt:45

```

Closes #1543 

## Checklist

- [x] PR description added
- [ ] tests are added // is this something I can test considering it is in the build execution itself?


ktlint/build.gradle.kts
```diff
@@ -61,7 +61,7 @@ val shadowJarExecutable by tasks.registering(DefaultTask::class) {
         // Then if that java version is >= 16, we add the --add-opens command
         // X = X=$( [ "$JV" -ge "16" ] && echo "--add-opens java.base/java.lang=ALL-UNNAMED" || echo "")
         // exec java $X -Xmx512m -jar "$0" "$@"
-        execFile.appendText("#!/bin/sh\n\nJV=\$(java -version 2>&1 | head -1 | cut -d'\"' -f2 | sed '/^1\\./s///' | cut -d'.' -f1)\n\nX=\$( [ \"\$JV\" -ge \"16\" ] && echo \"--add-opens java.base/java.lang=ALL-UNNAMED\" || echo \"\")\n\nexec java \$X -Xmx512m -jar \"\$0\" \"\$@\"\n\n")
+        execFile.appendText("#!/bin/sh\n\nJV=\$(java -version 2>&1 | sed -E -n 's/.* version \"([^.-]*).*\".*/\\1/p')\n\nX=\$( [ \"\$JV\" -ge \"16\" ] && echo \"--add-opens java.base/java.lang=ALL-UNNAMED\" || echo \"\")\n\nexec java \$X -Xmx512m -jar \"\$0\" \"\$@\"\n\n")
```
paul-dingemans, line 64: Can you explain why the old command failed on your system? Is it for example running on MS Windows? If so, please add some OS specific tests which will ensure that the same problem does not re-occur in the future.

paul-dingemans (changes requested):

paul-dingemans:
I am just reading the issue #1544 after making my review remarks. That issue clarifies already a lot.

rawilder:
Would you like me to update the PR description with text from the issue to make it clearer what this is doing?

paul-dingemans:
> Would you like me to update the PR description with text from the issue to make it clearer what this is doing?

That would have helped with the initial review. It would be great if you can still update it though for future reference.

rawilder:
Updated description with more info.

paul-dingemans (approved):

PR #1583, opened 2022-08-14
Enable trailing comma

paul-dingemans:
## Description

Enable trailing comma on declaration and call site. 

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:


shashachu:
Just to confirm: this not only allows but enforces a trailing comma?

paul-dingemans:
> Just to confirm: this not only allows but enforces a trailing comma?

Yes indeed. The codebase just feel way more consistent when the trailing comma is used always or never. This prevents discussions about usage of a trailing comma at specific location. But it is possible to set trailing comma's for call site and declaration site independently (also see separate commits).

The build is probably broken because other changes have been merged before this one and the trailing comma is not yet applied consistently in those changes.

PR #1585, opened 2022-08-14
Set API visibility explicitly

### LINKED ISSUE #1373: Clean up API (define visibility explicitly)
Specify the visibity of classes, functions, properties, etc. excplicitly. Currently it is not clear to our API users what is supposed to be internal functionality. Implicitly exposing internals of ktlint makes it harder to refactor the code base. 

Of course this would be a breaking change. So it might be a good a idea that we define a goal and go working toward a 1.0 release in which this can be settled.

paul-dingemans:
## Description

All rules are public so that they can be reused in alternative RulesetProviders.

Generic declarations from `package.kt` and `ElementType` are most likely to be used by external rule providers and are considered to be part of the public API.

Declarations which are most likely not to be used by API consumers have been made public but are also deprecated with note that they are marked for removal from public API in KtLint 0.48.

Closes #1373

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:


shashachu:
thanks for doing this bit of housekeeping.

PR #1606, opened 2022-08-24
Option `--patterns-from-stdin` and Git hook scripts update

mfederczuk:
## Description

This PR adds the option~~s~~ `--patterns-from-stdin[=<delimiter>]` ~~and `-z`~~ that'll read patterns separated by a delimiter (default is newline, giving an empty string ~~or using the `-z` option~~ is `NUL` byte) and add them to the patterns given on the command line.

Since the `--stdin` option would already read from `stdin`, it'll throw an exception if both `--stdin` and `--patterns-from-stdin`~~/`-z`~~ are given.

These options were primarily added to update the Git hook scripts.  
The main problems with the old ones were that files with special characters (e.g.: newline) in the name would not be passed along correctly (since Git quotes those filenames without the `-z` option) and that the `--no-run-if-empty` hack wouldn't work on non-GNU systems. (since that option is a GNU extension)

## Checklist

- [x] PR description added
- [x] tests are added
- [x] CHANGELOG.md` is updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
+* patterns can be read in from `stdin` with the `--patterns-from-stdin` command line options/flags ([#1606](https://github.com/pinterest/ktlint/pull/1606))
+
 ### Fixed
 
 * Let a rule process all nodes even in case the rule is suppressed for a node so that the rule can update the internal state ([#1644](https://github.com/pinterest/ktlint/issue/1644))
--- a/docs/install/cli.md
+++ b/docs/install/cli.md
@@ -189,6 +189,9 @@ ktlint installGitPrePushHook
 
 `--relative`: Print files relative to the working directory (e.g. dir/file.kt instead of /home/user/project/dir/file.kt)
 
+`--patterns-from-stdin[=<delimiter>]`: Reads additional patterns from `stdin`, where the patterns are separated by `<delimiter>`. If `=<delimiter>` is omitted, newline is used as fallback delimiter. If an empty string is given, the `NUL` byte is used as delimiter instead.
+Options `--stdin` and `--patterns-from-stdin` are mutually exclusive, only one of them can be given at a time.
+
 `-v`, `--verbose` or `--debug`: Turn on debug output. Also option `--trace` is available, but this is meant for ktlint library developers.
 
 `-V` or `--version`: Prints version information and exit.
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/KtlintCommandLine.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/KtlintCommandLine.kt
@@ -43,8 +43,10 @@ import kotlin.system.exitProcess
 import mu.KLogger
 import mu.KotlinLogging
 import org.jetbrains.kotlin.utils.addToStdlib.applyIf
+import picocli.CommandLine
 import picocli.CommandLine.Command
 import picocli.CommandLine.Option
+import picocli.CommandLine.ParameterException
 import picocli.CommandLine.Parameters
 
 private lateinit var logger: KLogger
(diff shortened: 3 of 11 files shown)


paul-dingemans:
Thnx for your contribution. I have to park this for a while until problems in latest release have been mitigated. I will come back to you in a couple of weeks.

docs/install/cli.md
```diff
@@ -189,6 +189,10 @@ ktlint installGitPrePushHook
 
 `--relative`: Print files relative to the working directory (e.g. dir/file.kt instead of /home/user/project/dir/file.kt)
 
+`--patterns-from-stdin[=<delimiter>]`: Reads additional patterns from `stdin`, where the patterns are separated by `<delimiter>`. If `=<delimiter>` is omitted, newline is used as fallback delimiter. If an empty string is given, the `NUL` byte is used as delimiter instead.  
+`-z` is a shorthand for `--patterns-from-stdin=''`
```
paul-dingemans, line 193: In the PR description it is documented that the option is primarily added for usage in the git hooks. In what other use cases would this flag be useful?

As I do not expect this flag to be used in a lot of other situations, I would like that option `-z` is removed. In the git hooks I would prefer to used the longer more descriptive `--patterns-from-stdin=''` anyways. The additional complexity of the mutual exclusivity will not be needed then.

ktlint/src/main/kotlin/com/pinterest/ktlint/internal/KtlintCommandLine.kt
```diff
@@ -217,6 +240,11 @@ internal class KtlintCommandLine {
         }
         logger = configureLogger()
 
+        assertStdinAndPatternsFromStdinOptionsMutuallyExclusive()
+
+        val stdinPatterns: Set<String> = readPatternsFromStdin()
```
paul-dingemans, line 245: Please add a CLI Test for the new flag. See SimpleCLITest for example of such a test.

ktlint/src/main/resources/ktlint-git-pre-commit-hook-android.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/sh
-# https://github.com/pinterest/ktlint pre-commit hook
-# On Linux xargs must be told to do nothing on no input. On MacOS (linux distribution "Darwin") this is default behavior and the xargs flag "--no-run-if-empty" flag does not exists
-[ "$(uname -s)" != "Darwin" ] && no_run_if_empty=--no-run-if-empty
-git diff --name-only --cached --relative | grep '\.kt[s"]\?$' | xargs $no_run_if_empty ktlint --android --relative
-if [ $? -ne 0 ]; then exit 1; fi
+
+# <https://github.com/pinterest/ktlint> pre-commit hook
+
+git diff --name-only -z --cached --relative -- '*.kt' '*.kts' | ktlint --android --relative -z
```
paul-dingemans, line 5: Cool that we can rid of the awkward workaround! Please replace flag `-z` with the longer `--patterns-from-stdin=''` in all hooks.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
+* patterns can be read in from `stdin` with the `--patterns-from-stdin` and `-z` command line options/flags
```
paul-dingemans, line 11: Add reference to the PR to end of line as follows `([#1606](https://github.com/pinterest/ktlint/pull/1606))`

paul-dingemans (changes requested):
Tnx for your patience. Please address the review remarks.

docs/install/cli.md
mfederczuk: > The additional complexity of the mutual exclusivity will not be needed then.

Sadly it is still needed, the mutual exclusivity has nothing to do with the `-z` shorthand, it exists because both the `--stdin` and `--patterns-from-stdin` options would try to read the entirety of the standard input, so only one of them can be given at the same time.

paul-dingemans: Ah yes, clear.

paul-dingemans (approved):

PR #1612, opened 2022-08-26
Update ImportOrdering to explain pointer

blundell:
## Description

The import ordering rule always points to the first line of the import block.

This always seems to confuse people into thinking it's pointing at the line which is at fault. (it's not)

Expand the error message to explain that the error is somewhere in the whole imports block.

image

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt
@@ -199,8 +199,10 @@ public class ImportOrderingRule :
         private val IDEA_PATTERN = parseImportsLayout("*,java.**,javax.**,kotlin.**,^")
 
         private const val IDEA_ERROR_MESSAGE = "Imports must be ordered in lexicographic order without any empty lines in-between " +
-            "with \"java\", \"javax\", \"kotlin\" and aliases in the end"
-        private const val ASCII_ERROR_MESSAGE = "Imports must be ordered in lexicographic order without any empty lines in-between"
+            "with \"java\", \"javax\", \"kotlin\" and aliases in the end. " +
+            "Pointer is at the start of the imports block. Error is somewhere in the whole block."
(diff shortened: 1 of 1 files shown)


paul-dingemans:
Thx for trying to improve ktlint. I do not think that the error message should be changed as proposed. Main reason is that it should not be that hard to determine the correct (or least a better offset). In method below the expected and actual imports are compared:
```
    private fun importsAreEqual(actual: List<ASTNode>, expected: List<ASTNode>): Boolean {
        if (actual.size != expected.size) return false

        val combined = actual.zip(expected)
        return combined.all { (first, second) ->
            if (first is PsiWhiteSpace && second is PsiWhiteSpace) {
                return@all (first as PsiWhiteSpace).text == (second as PsiWhiteSpace).text
            }
            return@all first == second
        }
    }
```
The first import for which the actual import is not equal to the expected import should be the line that needs to be reported.

Would you mind to try to try above?

paul-dingemans (changes requested):
Thx for trying to improve ktlint. I do not think that the error message should be changed as proposed. Main reason is that it should not be that hard to determine the correct (or least a better offset). In method below the expected and actual imports are compared:

    private fun importsAreEqual(actual: List<ASTNode>, expected: List<ASTNode>): Boolean {
        if (actual.size != expected.size) return false

        val combined = actual.zip(expected)
        return combined.all { (first, second) ->
            if (first is PsiWhiteSpace && second is PsiWhiteSpace) {
                return@all (first as PsiWhiteSpace).text == (second as PsiWhiteSpace).text
            }
            return@all first == second
        }
    }
The first import for which the actual import is not equal to the expected import should be the line that needs to be reported.

Would you mind to try to try above?

blundell:
@paul-dingemans what if there are multiple ordering errors, pointing at one of them could also lead to confusion as its not pointing to all the errors. I assume this may be some of the reason why it just points to the whole block.

paul-dingemans:
Yes, in case of multiple errors it can still be confusing but at least you are directed to the first wrong line. After fixing it, the line number at least will be increased.

paul-dingemans:
@blundell Please create a new PR if you want to improve the error message as I suggested before.

PR #1628, opened 2022-09-02
Migrate buildSrc to composite build

Goooler:
## Description
Ref https://github.com/detekt/detekt/pull/4090

https://docs.gradle.org/7.5.1/userguide/composite_builds.html

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:


gradle.properties
```diff
@@ -21,7 +21,7 @@ org.gradle.caching=true
 org.gradle.vfs.watch=true
 
 # Uncomment following line to temporary bypass dependencies verification
-#org.gradle.dependency.verification=lenient
+org.gradle.dependency.verification=lenient
```
Goooler, line 24: https://github.com/gradle/gradle/issues/20234

Goooler:
@3flex Can you take a look at this?

paul-dingemans:
@Goooler How will users of KtLint benefit from this? Which users will benefit? What do users need to do to benefit from the changes? Please document in the [user guide](https://pinterest.github.io/ktlint)

Goooler:
@paul-dingemans Nope, just for faster builds, cause `buildSrc` is slowm, a change in [buildSrc causes the whole project to become out-of-date](https://docs.gradle.org/7.2/userguide/organizing_gradle_projects.html#sec:build_sources).

paul-dingemans:
> @paul-dingemans Nope, just for faster builds, a change in [buildSrc causes the whole project to become out-of-date](https://docs.gradle.org/7.2/userguide/organizing_gradle_projects.html#sec:build_sources).

So only developers of the ktlint library will benefit?

Goooler:
Yes.

build-logic/build.gradle.kts
```diff
@@ -0,0 +1,20 @@
...
+    logger.warn("Enabling kotlin dev version!")
+    libs.versions.kotlinDev.get()
+} else {
+    libs.versions.kotlin.get()
+}
+
+dependencies {
+    implementation(kotlin("gradle-plugin", kotlinVersion))
+    implementation(libs.dokka)
+}
```
TWiStErRob, line 20: This goes against the laziness version catalogs are built on, it's probably better (and simpler using libs):
```suggestion
dependencies {
    if (project.hasProperty("kotlinDev")) {
        // Pass '-PkotlinDev' to command line to enable kotlin-in-development version
        logger.warn("Enabling kotlin dev version!")
        implementation(libs.kotlin.plugin.dev)
    } else {
        implementation(libs.kotlin.plugin)
    }
    implementation(libs.dokka)
}
```

where
```
[libraries]
kotlin-plugin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
kotlin-plugin-dev = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlinDev" }
```

gradle/libs.versions.toml
```diff
@@ -11,6 +12,7 @@ shadow = "com.github.johnrengelman.shadow:7.1.2"
 
 [libraries]
 kotlin-compiler = { module = "org.jetbrains.kotlin:kotlin-compiler-embeddable", version.ref = "kotlin" }
+dokka = "org.jetbrains.dokka:dokka-gradle-plugin:1.7.10"
```
TWiStErRob, line 15: Any reason not to do this?
```suggestion
dokka = { module = "org.jetbrains.dokka:dokka-gradle-plugin", version.ref = "kotlin" }
```

Goooler: Cause Dokka was out later than Kotlin, we'd like to bump Kotlin first.

build-logic/build.gradle.kts
Goooler: Could also be tweaked to below?
```kotlin
dependencies {
    val kotlinPlugin = if (project.hasProperty("kotlinDev")) {
        // Pass '-PkotlinDev' to command line to enable kotlin-in-development version
        logger.warn("Enabling kotlin dev version!")
        libs.kotlin.plugin.dev
    } else {
        libs.kotlin.plugin
    }
    implementation(kotlinPlugin)
    implementation(libs.dokka)
}
```

gradle/libs.versions.toml
TWiStErRob: Ah, good to know, thanks!

build-logic/build.gradle.kts
Goooler: 8670227077af7d7ec25f0335c6359f52fa4a0b73

Goooler:
@TWiStErRob Can you take a look again?

ktlint/build.gradle.kts
```diff
@@ -76,11 +76,10 @@ tasks.register<Checksum>("shadowJarExecutableChecksum") {
     description = "Generates MD5 checksum for ktlint executable"
     group = "Distribution"
 
-    files = shadowJarExecutable.get().outputs.files
+    inputFiles.setFrom(shadowJarExecutable.get().outputs.files)
```
TWiStErRob, line 79: While you're fixing deprecations, consider going fully lazy:
`inputFiles.setFrom(shadowJarExecutable.map { it.outputs.files })`
(Same for few line down.)

---

Below are some feedback on these tasks, not in scope in this PR just wanted to mention them.

---

Note that finalizedBy will execute checksum even on failure.

---

The recommendation from Gradle is to always create a full custom task, that way you don't need mapping like here, but you can just say `from(otherTask.outputDir)` where `outputDir` is a `DirectoryProvider`. Also caching and path sensitivity is not defined in register.

---

After writing all this, I realized that there's an overlapping output directory between tasks here. This is dangerous as noted by Gradle. The only example is see though is `cleanShadowJarExecutableChecksum` will delete the output of shadowJar too.

ktlint-ruleset-template/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
-    `ktlint-kotlin-common` // replace it with 'org.jetbrains.kotlin.jvm'
+    id("ktlint-kotlin-common")
```
TWiStErRob, line 2: Was this `` ` -> id("") `` because Gradle doesn't generate the magic any more? For consistency would it make sense to replace the other backtick plugins with strings too?

TWiStErRob (approved):
Lgtm, some minor suggestions which maybe be scope creep. I think it's ready for review. Note: I'm an outsider to ktlint.

ktlint-ruleset-template/build.gradle.kts
Goooler: Yeah, will generate typesafe id with buildSrc.

ktlint/build.gradle.kts
Goooler: a0737b4cb38eabc566d1fbf8162860dddfece391

gradle.properties
TWiStErRob: _(Whops this got a bit long, hope it helps readers who take the time to read to learn something new 😅)_

You can undo this if you add this to `build-logic/build.gradle.kts`:
```
buildscript {
    repositories {
        mavenCentral()
        //gradlePluginPortal()
    }
}
```

Or to be more specific:
```
buildscript {
    repositories {
        // Note: this is actually redundant, because exclusiveContent wins anyway, but it makes things
        // a bit more explicit and future-proof, in case the exclusiveContent gets removed.
        gradlePluginPortal {
            content {
                // See https://github.com/gradle/gradle/issues/20234
                excludeVersion("de.undercouch", "gradle-download-task", "4.1.1")
            }
        }
        // Note: with exclusiveContent, order doesn't matter.
        exclusiveContent {
            forRepository { mavenCentral() }
            filter {
                // See https://github.com/gradle/gradle/issues/20234
                includeVersion("de.undercouch", "gradle-download-task", "4.1.1")
            }
        }
    }
}
```

---

Here's a bit of an explanation:
As shown in https://github.com/gradle/gradle/issues/20234#issuecomment-1254096198, the poms are different in the different repositories. It seems that mavenCentral() PGP-validates for this artifact's POM, so that's the main source.

The validation error you're seeing here:
```
> Configure project :build-logic
Dependency verification failed for configuration ':build-logic:classpath'
One artifact failed verification: gradle-download-task-4.1.1.pom (de.undercouch:gradle-download-task:4.1.1) from repository Gradle Central Plugin Repository
This can indicate that a dependency has been compromised. Please carefully verify the signatures and checksums.
```
can be interpreted as follows:
 * `Configure project :build-logic` -> so the problem is before executing settings.gradle.kts from the main project
 * `configuration  ':build-logic:classpath'` -> so the problem is with the `buildscript`'s classpath inside the `build-logic` module. Remember the old way we added plugins (`buildscript { dependencies { classpath ...`)
 * `gradle-download-task-4.1.1.pom from repository Gradle Central Plugin Repository` -> the problem is with this artifact in `gradlePluginPortal()`
 * via `gradlew -p build-logic :buildEnvironment` I can see:
    ```
    > Task :buildEnvironment

    ------------------------------------------------------------
    Root project 'build-logic'
    ------------------------------------------------------------

    classpath
    +--- org.gradle.kotlin.kotlin-dsl:org.gradle.kotlin.kotlin-dsl.gradle.plugin:2.3.3
    |    \--- org.gradle.kotlin:gradle-kotlin-dsl-plugins:2.3.3
    |         +--- org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.21
    |         |    +--- de.undercouch:gradle-download-task:4.1.1
    ```
    So this is the path to the dependency transitively.
 * `org.gradle.kotlin.kotlin-dsl.gradle.plugin` is on `classpath` because
   * ``plugins { `kotlin-dsl` }``
   * if there's no `buildscript {}` in a build.gradle.kts
     * it either comes from `pluginManagement` in settings.gradle (which doesn't exist in build-logic)
     * or it is implied to be `buildscript { repositories { gradlePluginPortal() } }`
   * and of course `plugins {}` adds dependencies to `classpath` configuration

So to prevent a verification error we need to make sure the dependency comes from the right repository (mavenCentral()). We have multiple ways to achieve this:
 1) By explicitly declaring `mavenCentral() + gradlePluginPortal()` we're giving priority to `mavenCentral()` (which has the [right POM](https://github.com/gradle/gradle/issues/20234#issuecomment-1254096198), and for anything not present in central, we fall back to Gradle's. 
 2) By explicitly declaring `mavenCentral()` without `gradlePluginPortal()` we're restoring the original artifact resolution and therefore getting the right POM.
 3) By keeping `gradlePluginPortal()` as the main repository, but taking this specific artifact from `mavenCentral()` via `exclusiveContent`.

Relying on the order (1) feels a bit hacky, that's why (2) is probably preferable, but if the complexity is acceptable (pending pinterest's approval) maybe (3) is also ok because it's closer to the default Gradle behavior only fixing the one small thing that's broken.

---

As an additional note, we should be also mindful that `build-logic` gets built based on this `classpath`, but then its `runtimeClasspath` becomes part the including build's `pluginManagement` (`buildscript`>`classpath`). Which means that `pluginManagement.repositories` must apply the same fix (it already has both repos in the right order). At this point I tried using only `gradlePluginPortal()` in `/settings.gradle.kts` and it resulted in these validation errors:
```
> Configure project :
Dependency verification failed for configuration ':classpath'
3 artifacts failed verification:
  - gradle-download-task-4.1.1.pom (de.undercouch:gradle-download-task:4.1.1) from repository Gradle Central Plugin Repository
  - kotlin-gradle-plugin-1.7.10.module (org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.10) from repository Gradle Central Plugin Repository
  - org.jetbrains.kotlin.jvm.gradle.plugin-1.7.10.pom (org.jetbrains.kotlin.jvm:org.jetbrains.kotlin.jvm.gradle.plugin:1.7.10) from repository Gradle Central Plugin Repository

> Configure project :ktlint
Dependency verification failed for configuration ':ktlint:classpath'
4 artifacts failed verification:
  - gradle-download-task-4.1.1.pom (de.undercouch:gradle-download-task:4.1.1) from repository Gradle Central Plugin Repository
  - kotlin-gradle-plugin-1.7.10.module (org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.10) from repository Gradle Central Plugin Repository
  - org.jetbrains.kotlin.jvm.gradle.plugin-1.7.10.pom (org.jetbrains.kotlin.jvm:org.jetbrains.kotlin.jvm.gradle.plugin:1.7.10) from repository Gradle Central Plugin Repository
  - dokka-gradle-plugin-1.7.10.pom (org.jetbrains.dokka:dokka-gradle-plugin:1.7.10) from repository Gradle Central Plugin Repository
```

Goooler: But seems I fix the sum by 1dc4a305ff30bad7f5c133ae7fb42841f802ecd4.

Goooler: This is because a change in [buildSrc causes the whole project to become out-of-date](https://docs.gradle.org/7.2/userguide/organizing_gradle_projects.html#sec:build_sources).

Goooler: > https://github.com/pinterest/ktlint/commit/1dc4a305ff30bad7f5c133ae7fb42841f802ecd4 means you accept that the POMs are different (because you validated there's no funky stuff going on in there).

Would prefer to follow what Gradle did, and `buildscript { mavenCentral() }` might be a little noisy, all repos should be configured in settings.gradle might be neat.

TWiStErRob: I meant that you changed the repository declarations too, not just the file structure. I agree with settings.gradle, but there's two of them and only one declares repos. The point here is replacing the repos with a trusted source.

I'm aware of the buildSrc problem, I did multiple migrations like this, and I'm doing a similar migration at work, but I couldn't prove where this solution is better. Do you know a surefire way to test this benefit?

Goooler: https://github.com/gradle/gradle/issues/21271#issuecomment-1191143416

paul-dingemans:
@Goooler @TWiStErRob Tnx for your contribution and collaboration on this topic. I have tried to follow along, but due to my limited Gradle knowledge, I fail to understand what has been done. If you both think that this change is ready for merge, then I will just have good faith in you both and do the merge ;-)
Please confirm that it is ready for merge.

Goooler:
I'm ready.

TWiStErRob:
Looks good to me.

QA (Tip: links are build scans, select Timeline page for visual details):
 1. [`gradlew clean test`](https://gradle.com/s/6mywlajgeyd6q) on this branch to get a baseline (nth try, not relevant scan)
 2. [`gradlew test --console=verbose`](https://gradle.com/s/iwzlpsmn3fz3g) and observe everything is up to date after (1)
 3. Make some significant-looking changes to build-logic (all files changed, even an additional Gradle callback is added):

    <details><summary>patch (click to expand)</summary>

    ```patch
    Index: build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
    IDEA additional info:
    Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
    <+>UTF-8
    ===================================================================
    diff --git a/build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts b/build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
    --- a/build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts	(revision 1dc4a305ff30bad7f5c133ae7fb42841f802ecd4)
    +++ b/build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts	(date 1663875693894)
    @@ -11,4 +11,6 @@
         explicitApi()
     }
     
    +afterEvaluate { println(kotlin.explicitApi) }
    +
     addAdditionalJdkVersionTests()
    Index: build-logic/src/main/kotlin/ToolchainForTests.kt
    IDEA additional info:
    Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
    <+>UTF-8
    ===================================================================
    diff --git a/build-logic/src/main/kotlin/ToolchainForTests.kt b/build-logic/src/main/kotlin/ToolchainForTests.kt
    --- a/build-logic/src/main/kotlin/ToolchainForTests.kt	(revision 1dc4a305ff30bad7f5c133ae7fb42841f802ecd4)
    +++ b/build-logic/src/main/kotlin/ToolchainForTests.kt	(date 1663875652574)
    @@ -33,6 +33,7 @@
             }
             setJvmArgs(jvmArgs)
         }
    +    println("$this: $jdkVersion -> $jdkVersionTests")
         tasks.named("check") {
             dependsOn(jdkVersionTests)
         }
    Index: build-logic/src/main/kotlin/ktlint-publication.gradle.kts
    IDEA additional info:
    Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
    <+>UTF-8
    ===================================================================
    diff --git a/build-logic/src/main/kotlin/ktlint-publication.gradle.kts b/build-logic/src/main/kotlin/ktlint-publication.gradle.kts
    --- a/build-logic/src/main/kotlin/ktlint-publication.gradle.kts	(revision 1dc4a305ff30bad7f5c133ae7fb42841f802ecd4)
    +++ b/build-logic/src/main/kotlin/ktlint-publication.gradle.kts	(date 1663875669242)
    @@ -18,14 +18,12 @@
     
     tasks.named<Jar>("javadocJar") {
         dependsOn(tasks.dokkaJavadoc)
    -    archiveClassifier.set("javadoc")
         from(tasks.dokkaJavadoc)
    +    archiveClassifier.set("javadoc")
     }
     
    -project.version = project.property("VERSION_NAME")
    -    ?: throw GradleException("Project version property is missing")
    -project.group = project.property("GROUP")
    -    ?: throw GradleException("Project group property is missing")
    +project.version = project.property("VERSION_NAME")!!
    +project.group = project.property("GROUP")!!
     
     publishing {
         publications {
    ```
    </details>

 4. [`gradlew test --console=verbose`](https://gradle.com/s/se6lczvet462m)

Build (4) is a 16 second build, where most of the things were up to date, except `shadowJarExecutable`, which I couldn't explain reasonably (other types of changes did not trigger this).
But this is good enough for verification, especially contrasting with:

 1. [`gradlew clean test`](https://gradle.com/s/ml2o5l5ekyrm6) on `master` branch to get a base build
 2. [`gradlew test --console=verbose`](https://gradle.com/s/c6sxquy2nm5ze) and observe everything is up to date after (1)
 3. Make the exact same changes as above to `buildSrc`.
 4. [`gradlew test --console=verbose`](https://gradle.com/s/d2neu4opvf2eo)

where build (4) took 41 seconds, re-compiling almost everything, running some tests.

This should prove that using an `includeBuild` makes the build way faster when changing build logic.

settings.gradle.kts
```diff
@@ -3,6 +3,8 @@ pluginManagement {
         mavenCentral()
         gradlePluginPortal()
     }
+
+    includeBuild("build-logic")
```
TWiStErRob, line 7: One small nit that I'll throw in the air for @paul-dingemans to decide: we can put this build-logic into any folder. Not sure if the preference is to have all the "modules" visible, or only the main project source code. I usually prefer to keep things separate: i.e. GitHub stuff in `.github`, Gradle stuff in `gradle/*`, Documentation in `docs` etc. With `buildSrc` our hands were tied, there was no choice, Gradle hardcoded that path, but with freedom given I personally prefer to migrate `/buildSrc` to `/gradle/plugins` folder.

Goooler: We put publication & test logic into `build-logic` for almost modules to reuse.

TWiStErRob: Yep, I'm just saying to follow this step:
https://mobile.twitter.com/jeoj/status/1561994066530803712

(`@jeoj` is a master working at Gradle)

Goooler: TIL, thanks for your sharing. I think we can keep the current state to let users easily understand, will migrate to the idiomatic way one day.

Issue #1644, opened 2022-09-11
ktlint indentation rule has problems with ignoring braces


zsqw123:
## Expected Behavior
The following code should not be formatted
![image](https://user-images.githubusercontent.com/49520432/189515148-a9ad5783-d0ba-4160-8faa-a8a04aec7cb0.png)

## Observed Behavior
But when I run `ktlint -F`, I get the following.
![image](https://user-images.githubusercontent.com/49520432/189515078-03e4077a-42f8-4674-a391-9a85255ac004.png)

## Steps to Reproduce
```kotlin
fun main() {
    fun innerFun0() {
        // some code
    } // ktlint-disable

    fun innerFunc1() {
        // some code
    }
}

```

## Your Environment

* Version of ktlint used: 0.47
* No `.editorconfig`
* Version of Gradle used (if applicable): No
* Operating System and version: MacOS 12.5

## Possible reasons
ktlint's indentation rule records the braces with a stack, and when I use `// ktlint-disable` on the line where the braces are, the indentation rule does not check this `ASTNode` and the stack does not change, hence the problem.

zsqw123:
I think the solution to this is simply to make ktlint always perform a visit for each node. But for nodes that do not require format, we do not modify them

Issue #1645, opened 2022-09-11
ktlint indentation rule has problems with disable check on a single line


zsqw123:
## Behavior
Like the code below, I want to disable the format of on the first `.append` method, but that doesn't seem to work
```kotlin
fun main() {
    StringBuilder()
    .append("1") // ktlint-disable
    .append("1")

    // expect:
    StringBuilder()
    .append("1") // ktlint-disable
        .append("1")

    // but:
    StringBuilder()
        .append("1") // ktlint-disable
        .append("1")
}
```

## Steps to Reproduce
Using the code mentioned in **[Behavior](#Behavior)**, and run `ktlint -F`

## Your Environment
* Version of ktlint used: 0.47
* No .editorconfig
* Version of Gradle used (if applicable): No
* Operating System and version: MacOS 12.5

## Possible reasons
ktlint's rules use `startOffset` for determining whether to disable, but `PsiWhiteSpace`'s `startOffset` is on the previous line when it contains `\n`, which seems to lead to this error

paul-dingemans:
Reference to PR #1646 is invalid. This issue is not fixed by that PR.

paul-dingemans:
I am not in favor of changing this behavior as I am considering to remove the ktlint-disable/ktlint-enable directives as the `@Suppress("ktlint:all")` is more native.

PR #1646, opened 2022-09-11
fix indentation rule has problems with ignoring braces

zsqw123:
## Description

fix ktlint indentation rule has problems with ignoring braces #1644 

## Checklist

- [x] PR description added
- [x] tests are added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-* Let a rule process all nodes even in case the rule is suppressed for a node so that the rule can update the internal state ([#1645](https://github.com/pinterest/ktlint/issue/1645))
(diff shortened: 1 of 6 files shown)


paul-dingemans:
Thx for your contribution. I like the idea. The way it is implemented is a bit hard to understand. So I will refactor your solution before merging.

zsqw123:
In fact, my idea is simple. I just change the logic of suppress from never `visit` to never `emit`. And I tested it locally, and it passed the unit test.

paul-dingemans:
> In fact, my idea is simple. I just change the logic of suppress from never `visit` to never `emit`. And I tested it locally, and it passed the unit test.

Yes the idea is solid and your solution worked fine! I wanted a different implementation so that the logic for determining the new autoCorrect and emit values in a single place (now in the SuppressHandler).

zsqw123:
#1644 is different from #1645. This pr fixes #1644 only

PR #1658, opened 2022-09-23
Add CLI option `--log-level=<level>'

### LINKED ISSUE #1632: Set log level as (CLI) parameter
Add new CLI flag `--log-level=<level>` and remove flags `--debug`, `--trace` and `--verbose`.  The new flag can be sued to set the log level to any of the used log levels. In addition to current situation, this makes it also possible to suppress warning and/or errors on. As of that users can more easily control the amount of logging produced by ktlint.

Possible downside is that `--log-level=debug` is more cumbersome to specify than `--debug` or `-v`.

_Originally posted by @paul-dingemans in https://github.com/pinterest/ktlint/issues/1624#issuecomment-1235646808_

### LINKED ISSUE #1652: Logging output is sent to stdout instead of stderr when using --stdin
When using ktlint with --stdin, it's expected that the output of stdout is only the formatted code. Instead, an info logging line is prepended to the stdout with 0.47.1. The pre-commit hook we use (based on https://github.com/hallettj/git-format-staged) relies on the --stdin functionality, and using the latest version the committed files end up with an extra line in the file

## Expected Behavior
When called with -F --stdin, only the formatted code should be sent to stdout. Additonal output may be in stderr

## Observed Behavior
An INFO level logging line is always output as the first line of stdout

## Steps to Reproduce
```
> echo "class Foo {
fun foo() {}
}" | ./ktlint-0.47.1 -F --stdin 2>/dev/null
```

Outputs
```
10:02:35.047 [main] INFO com.pinterest.ktlint.internal.KtlintCommandLine - Enable default patterns [**/*.kt, **/*.kts]
class Foo {
    fun foo() {}
}
```

paul-dingemans:
## Description

Replace CLI options `--debug`, `--trace`, `--verbose` and `-v` are with `--log-level=<level>` or the short version `-l=<level>.

Closes #1632

* Allow disabling logging in CLI by setting `--log-level=none` or `-l=none`.

Closes #1652

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:


build.gradle
```diff
@@ -34,11 +34,12 @@ task ktlint(type: JavaExec, group: LifecycleBasePlugin.VERIFICATION_GROUP) {
...
-    // we are not pleased ourselves with the results on the ktlint code base.
-    '--experimental'
-    // Do not run with option "--verbose" or "-v" as the lint violations are difficult to spot between the amount of
-    // debug output lines.
+    // Experimental rules run by default run on the ktlint code base itself. Experimental rules should not be released
+    // if we are not pleased ourselves with the results on the ktlint code base.
+    '--experimental',
+    // Run with log-level error so that all lint violations can be spotted easily and are not drowned in a lot of log
+    // lines.
+    '--log-level=error'
```
Goooler, line 42: Seems this change needs to resolve conflicts with 28ff448abf9c795900ced4191da06ec6c1ce0764, should add a trailing comma after `'--log-level=error'`.

paul-dingemans: Yes, I noted that as well when I was reviewing the code. Tnx for notifying it as well.

PR #1669, opened 2022-10-07
Enforce trailing commas at declaration site by default

3flex:
## Description
[Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas) says:

>  The Kotlin style guide encourages the use of trailing commas at the declaration site

But ktlint's default config for this rule enforces no comma which directly conflicts with the coding conventions which _encourages_ trailing commas.

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Calling this API with a file path results in the `.editorconfig` files that will
 
 ### Changed
 * Update Kotlin development version to `1.7.20` and Kotlin version to `1.7.20`.
+* The default value for trailing comma's on declaration site is changed to `true` unless the `android codestyle` is enabled. Note that KtLint from a consistency viewpoint *enforces* the trailing comma on declaration site while default IntelliJ IDEA formatting only *allows* the trailing comma but leaves it up to the developer's discretion. ([#1669](https://github.com/pinterest/ktlint/pull/1669))
 
 ## [0.47.1] - 2022-09-07
 
--- a/docs/rules/standard.md
+++ b/docs/rules/standard.md
@@ -176,6 +176,18 @@ Rule id: `trailing-comma-on-call-site`
 
 Consistent removal (default) or adding of trailing comma's on declaration site.
 
+!!! important
+KtLint uses the IntelliJ IDEA `.editorconfig` property `ij_kotlin_allow_trailing_comma` to configure the rule. When this property is enabled, KtLint *enforces* the usage of the trailing comma at declaration site while IntelliJ IDEA default formatter only *allows* to use the trailing comma but leaves it to the developer's discretion to actually use it (or not). KtLint values *consistent* formatting more than a per-situation decision.
+
+!!! note
+In KtLint 0.48.x the default value for using the trailing comma on declaration site has been changed to `true` except when codestyle `android` is used.
+
+    The Kotlin coding conventions](https://kotlinlang.org/docs/reference/coding-conventions.html#names-for-test-methods) encourages the usage of trailing comma's on the declaration site, but leaves it to the developer's discretion to use trailing comma's on the call site. But next to this, it also states that usage of trailing commas has several benefits:
+    
+     * It makes version-control diffs cleaner – as all the focus is on the changed value.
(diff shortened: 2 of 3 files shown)


paul-dingemans:
Tnx for your patch. For this setting I tend to agree with you that `true` as default setting is more aligned with Kotlin Coding Conventions. For Android, I can not find such a recommendation and as of that it should not change (set `androidDefaultValue` to `false`). Please also update the documentation and changelog.

3flex:
I don't follow the logic for Android - I thought the Android style guide was essentially considered an extension of the Kotlin Coding Conventions, with some extra rules, and a few things that override the standard conventions, though perhaps that's not how ktlint deals with it.

In any case, I don't think it's correct to set the Android value to `false` here: that means commas will be forbidden at declaration sites on Android which doesn't make sense - if commas aren't enforced because it's not covered by the Android style guide, then they shouldn't be forbidden, for the same reason.

If you don't agree I might make this a draft and open a new issue to discuss further and perhaps get external input.

paul-dingemans:
> I don't follow the logic for Android - I thought the Android style guide was essentially considered an extension of the Kotlin Coding Conventions, with some extra rules, and a few things that override the standard conventions, though perhaps that's not how ktlint deals with it.

As far as I know, the Kotlin Coding Convention and Androids Kotlin Style Guide are indenpendent of each other, but luckily they do overlap on a lot of things. For some editorconfig settings a different default value is used however.

> In any case, I don't think it's correct to set the Android value to false here: that means commas will be forbidden at declaration sites on Android which doesn't make sense - if commas aren't enforced because it's not covered by the Android style guide, then they shouldn't be forbidden, for the same reason.

KtLint indeed has a more strict interpretation of the the flags than IntelliJ / Android. KtLint considers a consistent usage of the trailing more important than IntelliJ / Android. As the rules are already released with default value `false`, this won't change the behavior for Android style guide by setting the `androidDefaultValue` to `false` while `defaultValue` (for default Kotlin style) is set to `true`.

> If you don't agree I might make this a draft and open a new issue to discuss further and perhaps get external input.

I leave this up to you. However, I do not expect much input of the community when you raise this as a separate issue or mention it in the Slack channel. I would love to have more debate with the community about such changes but I have not found a way to get sufficient and timely feedback of the community.

3flex:
> the Kotlin Coding Convention and Androids Kotlin Style Guide are indenpendent of each other

Android Kotlin code is still Kotlin code, so wouldn't the Kotlin Coding Conventions apply? As long as the Android style guide doesn't have a more specific rule or override the Kotlin Coding Conventions, where you'd use the more specific guidance.

If we have `androidDefaultValue = false` while `defaultValue = true` then Android projects that have Kotlin/JVM modules will require trailing commas in the JVM modules and forbid them in the Android modules which I think will be quite confusing.

I'll still proceed with the PR even if we won't make them consistent, but hopefully this is convincing enough that we can have the same default for both Android and JVM...

paul-dingemans:
>> the Kotlin Coding Convention and Androids Kotlin Style Guide are indenpendent of each other

> Android Kotlin code is still Kotlin code, so wouldn't the Kotlin Coding Conventions apply? As long as the Android style guide doesn't have a more specific rule or override the Kotlin Coding Conventions, where you'd use the more specific guidance.

From the Android Kotlint Style Guide:
> This document serves as the **complete** definition of Google’s Android coding standards for source code in the Kotlin Programming Language. A Kotlin source file is described as being in Google Android Style if and only if it adheres to the rules herein.

To me, the word **complete** suggests that it is independent of any other coding standard.

> If we have androidDefaultValue = false while defaultValue = true then Android projects that have Kotlin/JVM modules will require trailing commas in the JVM modules and forbid them in the Android modules which I think will be quite confusing.

This is a valid concern. But the same is already happening for the max line length as well. For Android the default value is set to `100` while for default Kotlin the max line length is not set.

BraisGabin:
Following the arguments I would vote to disable the rule for android by default. If the code convention says nothing it doesn't imply that they should be forbidden. 

But I would rather prefer to follow the Kotlin convention when the android convention mention nothing about it.

PR #1670, opened 2022-10-07
Enforce trailing commas at call site by default

3flex:
## Description

The [Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas) says re: trailing commas:

> The Kotlin style guide... leaves it at your discretion for the call site.

However it also says:

> Using trailing commas has several benefits:
>
>* It makes version-control diffs cleaner – as all the focus is on the changed value.
>* It makes it easy to add and reorder elements – there is no need to add or delete the comma if you manipulate elements.
>* It simplifies code generation, for example, for object initializers. The last element can also have a comma.

This suggests only benefits but no downsides (other than personal taste) to enforcing the use of trailing commas by default.

ktlint currently enforces no commas at call sites by default, which is opinionated in a way that conflicts with this guidance and the rationale for using trailing commas in all cases.

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:


paul-dingemans:
Tnx for your patch. For this setting I have some more diffculty with setting `true` as default setting. As of that it should be explained in documentation and changelog that for consistency reasons the default is changed. For Android, I can not find such a recommendation and as of that it should not change (set `androidDefaultValue` to `false`). Please also update the documentation and changelog.

PR #1671, opened 2022-10-07
Register missing ktlint_disabled_rules in editorConfigProperties

tcrawford-figure:
## Description

`editorConfigProperties` seemed to be missing the newly created `ktlintDisabledRulesProperty`. For any downstream plugins that rely on this value to know what properties are valid for ktlint, they may filter out anything but these properties. 

Ultimately, this could leads to errors like the following if that's the only editor config override provided:
>Can not create an EditorConfigOverride without properties. Use 'emptyEditorConfigOverride' instead.

Or if it's not, I can safely assume that the disabled rules will be ignored if `disabled_rules` aren't provided as well.

Currently, `disabled_rules` still works but it spams a lot of messages stating that it is disabled. I find this to be an unpleasant developer experience and also leads to confusion for a developer who is simply switching to `ktlint_disabled_rules` from `disabled_rules` only to find that their disabled rules are no longer honored.

I did not find an issue reported regarding this issue, but please feel free to link any issues if I missed something.

## Checklist

- [x] PR description added
- [x] tests are added

In case of adding a new rule:


paul-dingemans:
Tnx for fixing this bug. The property indeed needs to be defined.

shubham96301:
Can someone please create a release for this fix?

shubham96301:
still waiting for the release

baseilos:
@paul-dingemans is there a timeline when this will be released?

paul-dingemans:
A couple of days, if all goes well.

PR #1678, opened 2022-10-13
Use Sequence to improve iterate performance

Goooler:
## Description

https://kotlinlang.org/docs/sequences.html

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
+++ b/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt
@@ -99,7 +99,7 @@ public object KtLint {
         val isInvokedFromCli: Boolean = false,
     ) {
         internal val ruleRunners: Set<RuleRunner> =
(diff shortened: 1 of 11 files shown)


paul-dingemans:
Hi @Goooler ,

Tnx for this PR. Am I correct to assume that you have applied the sequence to every collection that was missing it? Or did you actual profile the application?

According to https://kotlinlang.org/docs/sequences.html

> However, the lazy nature of sequences adds some overhead which may be significant when processing smaller collections or doing simpler computations. Hence, you should consider both Sequence and Iterable and decide which one is better for your case.

Most of the changes in this PR will typically run on small collections, although I have spotted a few which potential could contain a reasonable amount (tens to hundreds, but not thousands) of items.

Goooler:
I just added `.asSequence()` for collections that iterate more than once, it would be a slight improvement, but anyway, I'm closing this PR now.

PR #1693, opened 2022-11-02
Lint and format context receiver

### LINKED ISSUE #1672: Styling for context receiver
Context receiver was discussed in https://github.com/pinterest/ktlint/issues/1436 and introduced in Kotlin 1.6.20 https://kotlinlang.org/docs/whatsnew1620.html#prototype-of-context-receivers-for-kotlin-jvm

## Expected Rule behavior
Ensure that we have context receiver separated from other declarations based on the code examples in https://kotlinlang.org/docs/whatsnew1620.html#prototype-of-context-receivers-for-kotlin-jvm
Of course when context receiver is added to the coding convention, there is a chance that Kotlin coding convention might suggest against the above.

## Additional information
* Current version of ktlint:
* Styleguide section: There are a few references but no official style guide yet. This means that this rule could wait until context receiver is promoted to stable feature.
 - https://kotlinlang.org/docs/whatsnew1620.html#prototype-of-context-receivers-for-kotlin-jvm has some examples
 - https://github.com/Kotlin/KEEP/blob/master/proposals/context-receivers.md

paul-dingemans:
## Description

Change `indent` rule to accodomodate for the context receiver list. Add new rule `content-receiver-wrapping` to wrap a context receiver to a separate line regardless of the max line length setting, and, wrap context receiver and type projections to separate lines when max line length is exceeded.

Some example of code after auto correct:
```
    // ALways wrap regardless of whether max line length is set
    context(Foo)
    fun fooBar()

    // Wrap each context receiver to a separate line when the
    // entire context receiver list does not fit on a single line
    context(
        Fooooooooooooooooooo1,
        Foooooooooooooooooooooooooooooo2
    )
    fun fooBar()

    // Wrap each context receiver to a separate line when the
    // entire context receiver list does not fit on a single line.
    // Also, wrap each of it projection types in case a context
    // receiver does not fit on a single line after it has been
    // wrapped.
    context(
        Foooooooooooooooo<
            Foo,
            Bar
            >
    )
    fun fooBar()
```

Closes #1672

## Checklist

- [X] PR description added
- [X] tests are added

In case of adding a new rule:
- [X] Rule is added to [rules documentation](https://pinterest.github.io/ktlint/rules/standard/)


chao2zhang:
Thank you! I will test in my repositories and report back if there are any edge cases

PR #1716, opened 2022-11-25
Bump Gradle to 7.6

Goooler:
## Description

https://docs.gradle.org/7.6/release-notes.html

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:


build-logic/src/main/kotlin/ToolchainForTests.kt
```diff
@@ -9,7 +9,7 @@ val Project.javaToolchains: JavaToolchainService
 
 fun Project.addAdditionalJdkVersionTests() {
     // Tests should be run on all supported LTS versions of the JDK. For example, see https://endoflife.date/java
-    listOf(8, 11, 17, 18).forEach(::addJdkVersionTests)
+    listOf(11, 17, 18).forEach(::addJdkVersionTests)
```
paul-dingemans: Why removing Java8? It is not end-of-life.

Goooler: I'd just tested builds on my M1 mac, already reverted line, right?

paul-dingemans: Indeed, you did revert in commit with description [Merge branch 'trunk' into gradle_760](https://github.com/pinterest/ktlint/pull/1716/commits/cad9abe901e5f05b23ba874adb17a5a397f5faf3). Not really what I expected.

PR #1737, opened 2022-12-18
Add `ktlint_official` codestyle

### LINKED ISSUE #1699: Add new codestyle `KtLint Official`
Currently KtLint supports the [Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html) and [Android's Kotlin styleguide](https://developer.android.com/kotlin/style-guide). Both codestyles try to prevent conflicts with the IntelliJ IDEA Default formatter. As a result, KtLint sometimes is forced to fix formatting to reproduce erroneous styling of IntelliJ IDEA Default formatting. Some of those issues have been reported to Jetbrains issuetracker several years ago and are still not picked up.

The new codestyle `KtLint Official` will have following characteristics:
* Tries to match with IntelliJ IDEA Default formatting but may produce code which will be formatted differently when applying the IntelliJ IDEA default formatting.
* Primarily adheres to the [Kotlin Coding Conventions](https://kotlinlang.org/docs/coding-conventions.html). If [Android's Kotlin styleguide](https://developer.android.com/kotlin/style-guide) contains additional rules which are not conflicting with the Kotlin Coding conventions, those will be used. The `Ktlint Official` codestyle will be opinionated and might even contain rules like indentation of raw string literals (#925) which by default will be disabled in the other codestyles.
* The initial release of the `KtLint Official` codestyle will use an opt-in model, meaning that this codestyle has to be set explicitly in the `.editorconfig` (or via a CLI parameter). The intent however is that this becomes the default codestyle for Ktlint from which user have to opt-out out by explicitly configuring another codestyle.
 
Currently, the main differences between the Kotlin and Android codestyles in KtLint are limited to different default values for the `.editorconfig` properties. With introduction of the 'KtLint Official' codestyle, some parts of the rules in KtLint might actually get a different implementation.

Please checkout the issues which potentially will be solved in https://github.com/pinterest/ktlint/labels/ktlint-official-codestyle. Please drop a comment with your thoughts. If you have another candidate for the `Ktlint Official` codestyle then please create a separate issue and it will be tagged as https://github.com/pinterest/ktlint/labels/ktlint-official-codestyle.

paul-dingemans:
## Description

Add code style `ktlint_official` (functionally equivalent with old `official` code style), `intellij_idea` (replacement for old `official` code style) and `android_studio` (replacement for old `android` code style).

Closes #1699

## Checklist

- [X] PR description added
- [X] tests are added

In case of adding a new rule:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Unreleased
 
+### API changes
+
+#### EditorConfigProperty
+
+Field `defaultAndroidValue` has been renamed to `intellijIdeaCodeStyleDefaultValue`. Note that also new fields `ktlintOfficialCodeStyleDefaultValue` and `intellijIdeaCodeStyleDefaultValue`. Read more about this in the section "Ktlint Official code style".
+
+
+### 'Ktlint Official` code style and renaming of existing code styles
+
+A new code style with name `ktlint_offical` has been added. This code style makes it possible to innovate ktlint beyond current expectations of users that are familiar with the existing code styles. 
+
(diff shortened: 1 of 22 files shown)


paul-dingemans:
Replace with #1744

PR #1748, opened 2022-12-26
Interface experimental rule

paul-dingemans:
## Description

Add interface "Rule.Experimental" to mark a rule as an experimental rule. Rules no longer have to be part of the "experimental" rule set to be treated as an experimental rule. This allows experimental rules to be defined in any rule set. Whenever a rule is no longer experimental it does not need to be moved to a non-experimental rule set. Benefit is that users of that rule, do not need to change ktlint disable directives.

Simplify code by moving logic from QualifiedRuleId to Rule. All logic regarding qualified rule (set) id is now kept together in once class.

The 'experimental` rule set of Ktlint has been merged with the `standard` rule set.

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:


paul-dingemans:
Replaced with https://github.com/pinterest/ktlint/pull/1767

Issue #1688, opened 2022-10-31
Annotation with entries should not force entries on separate lines


paul-dingemans:
Code below is acceptated unchanged by IntelliJ IDEA default formatter:
```
@[JvmStatic  Provides]
fun foo() = 42
```

Ktlint 0.48 snapshot rewrites it to:
```
@[
JvmStatic
Provides
]
fun foo() = 42
```

PR #1762, opened 2023-01-04
Fix another issue with annotations incorrectly being forced to be multi-line

shashachu:
Similar to #1688, an array of annotations would be forced to be multi-line if any of the entries had a parameter.

- [x] PR description added
- [x] tests are added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+* Issue with the `annotation` rule which would sometimes incorrectly force an array of annotations to be multi-line
+
 ### Changed
 
 ## [0.48.1] - 2023-01-03
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/AnnotationRule.kt
@@ -61,26 +61,27 @@ public class AnnotationRule : Rule("annotation") {
     ) {
         require(node.elementType == ANNOTATION_ENTRY)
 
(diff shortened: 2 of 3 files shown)


paul-dingemans:
This PR seem to violate the [Kotlin Coding Conventions regarding annotations](https://kotlinlang.org/docs/coding-conventions.html#annotations):

The main rule is:
> Place annotations on separate lines before the declaration to which they are attached, and with the same indentation:
>
> ```
> @Target(AnnotationTarget.PROPERTY)
> annotation class JsonExclude
> ```

An described exception is:
> Annotations without arguments may be placed on the same line:
> 
> ```
> @JsonExclude @JvmField
> var x: String
> ```

Although not mentioned explicitly, I read this as that by default all annotations go on a separate lines but it is allowed to put annotations without argument on the same line.

According to [Androids Kotlin Style Guide](https://developer.android.com/kotlin/style-guide) it is not allowed to use annotations with parameters inside `@[...]`:
> @[...] syntax may only be used with an explicit use-site target, and only for combining 2 or more annotations without arguments on a single line.
>
> ```
> @field:[JvmStatic Volatile]
> var disposable: Disposable? = null
> ```

Do you you have other references that would show that this should be allowed and is widely accepted in the community?

shashachu:
Ah, yes that seems to be what the docs imply. It's widely used in our own code but I don't have any idea whether it's industry standard. Let me do a bit of digging.

shashachu:
IntelliJ/Android Studio both allow the construct, but the docs do imply this is what you should do:

```
@[JvmStatic Provides]
@Foo("bar")
fun foo() = 42
```

Interestingly,
```
@[JvmStatic Provides] @Foo("bar")
fun foo() = 42
```

gets reformatted to the correct multi-line format in IJ, but not with ktlint. so I'll go ahead and file a separate issue.

shashachu:
https://github.com/pinterest/ktlint/issues/1765

Issue #1771, opened 2023-01-11
Setting ktlint_disabled_rules enables running of experimental rules


johnk-scalable:
## Expected Behavior

- Set ktlint_disabled_rules property to some value to disable some rules
- Don't explicitly disable experimental rules
- Run ktlint
- Expected that only standard rulesets and custom rulesets are run (minus the rules that have been explicitly disabled)

## Observed Behavior
standard rules, custom rules **and experimental rules** are run (minus the rules that have been explicitly disabled)

## Your Environment
* Version of ktlint used: 0.48.1

paul-dingemans:
> Please provide more information how to reproduce. Enclose ".editorconfig", command executed and log files produced with flag --log-level=debug.

Nvm, just see your PR.

PR #1772, opened 2023-01-11
Ensure that enabling the ktlint_disabled_rules property doesn't accidentally enable all experimental rules

johnk-scalable:
## Description

Solving issue: https://github.com/pinterest/ktlint/issues/1771#issue-1528688024

Not sure about how to organise the tests file. It was covered by existing tests but I also added a new test.


paul-dingemans:
I will have a look next week.

paul-dingemans:
Tnx for your contribution. The test that you added was actually redundant, so it is removed.

Issue #1773, opened 2023-01-12
FunctionSignatureRule with long function name suggest expression body on the same line, when no space available


jtietema:
Given a long function name (as found in test cases for example) with an expression body and no parameters:
```kotlin
fun `a very long function name as found in a test case`() =
                "some-result"
```
The function name here exceeds the `maxLineLength` of the `FunctionSignatureRule` but not the overal `MaxLineLength`.

## Expected Behavior

To accept this formatting as breaking down into multiline function signature will not fix the problem.

## Observed Behavior

Ktlint will assume there are parameters and we will go into multiline mode. Suggesting to put the expression on the same line as the function signature.

## Steps to Reproduce

Unit testcase is included in the following PR: https://github.com/pinterest/ktlint/pull/1774

## Your Environment

* Version of ktlint used: 0.47.1
* `.editorconfig` not used
* Name and version (or code for custom task) of integration used (Gradle plugin, Maven plugin, command line, custom Gradle task): detekt formatting plugin 1.22
* Version of Gradle used (if applicable): 7.6
* Operating System and version: Fedora 37 - Linux

paul-dingemans:
The function signature rule does not have a separate maxLineLength property but uses the same property as the max line length rule.

Test names between backticks may exceed the max line length when property `ktlint_ignore_back_ticked_identifier` is enabled.

You have specified that you do not use a '.editorconfig' file. What is the exact command that you executed that reproduces problem above?

jtietema:
I am using the detekt-formatting plugin, so it is executed as part of `./gradlew detekt`.
Detekt does have a feature to override the `maxLineLength` for the rule https://detekt.dev/docs/rules/formatting#functionsignature

I have added a test case in the PR #1774 which demonstrates the issue. If you have a function with:
 -  a very long name that runs past the maxLineLength
 - has no arguments
 - has a short expression body
The ktlint rule will incorrectly calculate the remaining space as it will assume the `) =` will be on its own line. Resulting in a remaing space of  `maxLineLength - ") =".length` when in reality it is a single line (as we don't have parameters to wrap) so the remaining space should be `0` (at least, that is what I think it should be :sweat_smile:).

paul-dingemans:
This is a regression bug introduced in `0.48.x`.

PR #1774, opened 2023-01-12
Add check for presence of parameters before assuming multilinemode

jtietema:
## Description

The FunctionSignature rule would assume multiline mode when exceeding the maxLineLength, but this isn't always the case.

Given a long function name (as found in test cases for example) with an expression body and no parameters:
```kotlin
fun `a very long function name as found in a test case`() =
                "some-result"
```

The function name here exceeds the maxLineLength of the FunctionSignatureRule but not the overal MaxLineLength.

The ktlint error would suggest to put the expression body on the same line:
`First line of body expression fits on same line as function signature [FunctionSignature]`

Which obviously won't fit. The code would calculate the remaining space based on multiline mode (assuming the line length is just `) =`). This fix checks if there are parameters on the function and if not used the full line for space calculation, not the last 3 nodes.

Fixes: #1773

## Checklist

- [x] PR description added
- [x] tests are added

In case of adding a new rule:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
+* A function signature not having any parameters which exceeds the `max-line-length` should be ignored by rule `function-signature` ([#1773](https://github.com/pinterest/ktlint/issues/1773))
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


paul-dingemans:
Hi Jeroen,
Tnx for your contribution. As you can see, I added some explanatory comments and more tests.

Issue #1793, opened 2023-01-27
Git hooks: default patterns are used if no Kotlin files have changed


mfederczuk:
This is a follow up to my PR #1606, concerning an oversight on my part.

In the new Git hooks, if `git diff` doesn't report any Kotlin files changed (which happens either because no Kotlin files are staged (`pre-commit`), or because no Kotlin files have been changed between `HEAD` and the remote branch (`pre-push`)) then `ktlint` will fall back to the default patterns, which match *every* file in the project — tracked in VC or not.

For some projects, this is undesirable, either because the codebase contains a lot of legacy code that isn't properly formatted or because the project contains generated code that doesn't conform to style conventions.

I have a couple ideas on how to solve this issue:

1. Add yet another new option `--no-default-patterns`. Then the hooks could be updated to:

   ```sh
   git diff --name-only -z --cached --relative -- '*.kt' '*.kts' | ktlint --relative --patterns-from-stdin='' --no-default-patterns
   ```

2. Disable the default patterns if the option `--patterns-from-stdin` is used, even if no patterns are given

3. Remove the default patterns outright

4. Reintroduce the `xargs --no-run-if-empty` hack:

   ```sh
   if [ "$(uname -s)" != "Darwin" ] then
       no_run_if_empty=--no-run-if-empty
   fi

   git diff --name-only --cached --relative -- '*.kt' '*.kts' | xargs $no_run_if_empty ktlint --relative
   ```

   Though this would also reintroduce the issues with that hack:

   * Pathnames with whitespace and/or special characters in it won't be found by `ktlint` (due to either `xargs` splitting them or `git diff` quoting them)
   * `--no-run-if-empty` is an extension of GNU `xargs`

---

For anyone else having this problem, a quick and dirty workaround for now would be to simply always give at least one file on the command line. For example:

```sh
git diff --name-only -z --cached --relative -- '*.kt' '*.kts' | ktlint --relative --patterns-from-stdin='' 'settings.gradle.kts'
```

paul-dingemans:
At first glance, option 2 seems most attractive to me. Adding default patterns doesn't make sense in case you want to get them from stdin (regardless whether stdin is empty or not) anyways.

paul-dingemans:
Solved by #1801

PR #1801, opened 2023-01-30
Disabling default patterns if the option `--patterns-from-stdin` is given

mfederczuk:
## Description

Simple change to not add the default patterns if the CLI option `--patterns-from-stdin` is given.

This PR fixes issue #1793.

## Checklist

- [x] PR description added
- [x] tests are added


ktlint/src/main/kotlin/com/pinterest/ktlint/internal/KtlintCommandLine.kt
```diff
@@ -381,7 +374,7 @@ internal class KtlintCommandLine {
         if (stdin && stdinDelimiter != null) {
             throw ParameterException(
                 commandSpec.commandLine(),
-                "Options --stdin and --patterns-from-stdin mutually exclusive",
+                "Options --stdin and --patterns-from-stdin are mutually exclusive",
```
paul-dingemans, line 377: Nice catch.

ktlint/src/main/kotlin/com/pinterest/ktlint/internal/KtlintCommandLine.kt
```diff
@@ -291,13 +291,6 @@ internal class KtlintCommandLine {
         val stdinPatterns: Set<String> = readPatternsFromStdin()
         patterns.addAll(stdinPatterns)
 
-        // Set default value to patterns only after the logger has been configured to avoid a warning about initializing
-        // the logger multiple times
-        if (patterns.isEmpty()) {
-            logger.info { "Enable default patterns $DEFAULT_PATTERNS" }
-            patterns = ArrayList(DEFAULT_PATTERNS)
-        }
```
paul-dingemans, line 299: What about something like:
```
        // Set default value to patterns only after the logger has been configured to avoid a warning about initializing
        // the logger multiple times
       patterns = when {
           stdinPatterns.isNotEmpty() ->
                // TBD: log warning in case patterns were also specified at command line?
                stdinPatterns
           patterns.isEmpty() -> {
                logger.info { "Enable default patterns $DEFAULT_PATTERNS" }
                ArrayList(DEFAULT_PATTERNS)
           }
           else -> {
               // Keep patterns specified at command line
              patterns
           }
       }
```

ktlint/src/test/kotlin/com/pinterest/ktlint/SimpleCLITest.kt
```diff
@@ -153,6 +172,21 @@ class SimpleCLITest {
...
+        @TempDir
+        tempDir: Path,
+    ) {
+        CommandLineTestRunner(tempDir)
+            .run(
+                "too-many-empty-lines",
+                listOf("--patterns-from-stdin"),
+                stdin = ByteArrayInputStream(ByteArray(0)),
+            ) {
+                assertNormalExitCode()
```
paul-dingemans, line 186: I would expect that 0 files are processed. This can be validated using the debug output.

ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
```diff
@@ -40,6 +40,7 @@ internal val DEFAULT_PATTERNS = DEFAULT_KOTLIN_FILE_EXTENSIONS.map { "**/*.$it"
 internal fun FileSystem.fileSequence(
     patterns: List<String>,
     rootDir: Path = Paths.get(".").toAbsolutePath().normalize(),
+    enableDefaultPatterns: Boolean = true,
```
paul-dingemans, line 43: I dislike boolean parameters. Also, I doubt whether an additional parameter is needed. The `patterns` can be set to `DEFAULT_PATTERNS` in 'KtlintCommandLine` except when `--patterns-from-stdin` is set.

paul-dingemans (changes requested):
Tnx for your contribution. Please address the review remarks.

ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
mfederczuk: Adding the default patterns in `KtlintCommandLine` does sound better, but the problem is that `FileSystem.fileSequence()` adds the default patterns itself when no patterns are passed in.
If we don't wanna change the signature of `FileSystem.fileSequence()`, then we'd need to change it so that it doesn't add the default patterns at all:

```diff
diff --git a/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt b/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
index 61b3a92c..51614f60 100644
--- a/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
+++ b/ktlint/src/main/kotlin/com/pinterest/ktlint/internal/FileUtils.kt
@@ -60,15 +60,9 @@ internal fun FileSystem.fileSequence(
 
     val globs = expand(patternsExclusiveExistingFiles, rootDir)
 
-    val pathMatchers = if (globs.isEmpty()) {
-        DEFAULT_PATTERNS
-            .map { getPathMatcher("glob:$it") }
-            .toSet()
-    } else {
-        globs
-            .filterNot { it.startsWith(NEGATION_PREFIX) }
-            .map { getPathMatcher(it) }
-    }
+    val pathMatchers = globs
+        .filterNot { it.startsWith(NEGATION_PREFIX) }
+        .map { getPathMatcher(it) }
 
     val negatedPathMatchers = if (globs.isEmpty()) {
         emptySet()

```

But this change breaks [a test in `FileUtilsTest.kt`](https://github.com/pinterest/ktlint/blob/d51de4cb14473ed5cbb732039585d5247c7af2de/ktlint/src/test/kotlin/com/pinterest/ktlint/internal/FileUtilsTest.kt#L68-L85).
Do we just remove that test?

paul-dingemans: The signature of the fileSequence function can be changed as it is an internal function. So it does not break any API.

paul-dingemans:
This didn't work out as I expected. There is no need for the fileSequence to receive two different lists of patterns. I have modified your solution. Can you have a look at my changes?

mfederczuk:
> There is no need for the fileSequence to receive two different lists of patterns.

Yes, to be honest I also didn't like that solution, I just didn't know how to properly change `fileSequence()` without breaking tests and other behavior.
Now it looks much better.

I'd say like this it's ready to be merged.

paul-dingemans:
Merged. Tnx for the collaboration.

PR #1804, opened 2023-01-30
Tweak CI configs a bit

Goooler:
## Description

A few tweaks.

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:


.github/workflows/gradle-pr-build.yml
```diff
@@ -3,6 +3,11 @@ name: PR Build
 on:
   pull_request:
 
+env:
+  ORG_GRADLE_PROJECT_signingKey : ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEY }}
+  ORG_GRADLE_PROJECT_signingKeyId: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEYID }}
+  ORG_GRADLE_PROJECT_signingKeyPassword : ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
```
Goooler, line 9: @shashachu I believe these secrets are unnecessary for PR builds for now, right?

.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -1,5 +1,15 @@
...
+  push:
+    paths:
+      - 'gradlew'
+      - 'gradlew.bat'
+      - 'gradle/wrapper/'
+  pull_request:
+    paths:
+      - 'gradlew'
+      - 'gradlew.bat'
+      - 'gradle/wrapper/'
```
Goooler, line 12: If we can remove the check for `Validation`?

.github/workflows/gradle-pr-build.yml
Goooler: If so, I'll push a commit the remove them.

shashachu: no, unfortunately, at least without some other tweaking. These being missing were the reason why the PR builds started failing once @paul-dingemans committed the version to be a non-SNAPSHOT version.

.github/workflows/gradle-wrapper-validation.yml
shashachu: This is probably correct, but the validation is also very quick so I don't think it's a huge savings either.

Goooler: Cause there might be some flaky failures from this action.

.github/workflows/gradle-pr-build.yml
Goooler: Gotcha.

shashachu (approved):

shashachu:
Ah, actually we might be stuck a little bit because I'm not sure if we can conditionally require status checks. So this can't be merged (except by me) because it's waiting for the Gradle Wrapper Validation to pass.

shashachu:
@Goooler I wonder if instead we can just return success from the Validation step if the wrapper files haven't changed?

Goooler:
Yep, that's reasonable, but I'm not sure how to implement it.

paul-dingemans:
I am not sure whether this PR should be merged, but I can not do so at this moment at there still a Validation step which is not completed.

PR #1821, opened 2023-02-14
Allow disabling signing

### LINKED ISSUE #1817: Allow skipping Signing on Release Builds for Packaging
## Expected Behavior
I'm the maintainer of the Archlinux AUR package `ktlint` which currently downloads the release binary and checksum verifies them. I now want to switch it over to building from source to possibly get it into the community repo at some point.

I expected to easily be able to build the binary as part of the package build. For this I'd expect to skip signing the binary.

## Current Behavior
When building release version it attempts to sign the binary, which is not really useful when packaging, since the package will be signed afterwards. It seems to sign whenever the version is not a snapshot version with no way to skip it.

## Additional information
* Current version of ktlint: 0.48.2

pschichtel:
## Description

This PR makes it possible to disable code signing for release builds. This is primarily useful for package maintainers building the project from source. In this scenario, ktlint's own signing is not necessary, since the resulting package will be signed anyway. The property `ktlint.disable-signing` (name open for discussion) is introduced, which when set during release builds, skips the signing.

Fixes #1817

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/build-logic/src/main/kotlin/ktlint-publication.gradle.kts
+++ b/build-logic/src/main/kotlin/ktlint-publication.gradle.kts
@@ -97,6 +97,8 @@ signing {
     val signingPassword = System.getenv("ORG_GRADLE_PROJECT_signingKeyPassword")
     useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
 
+    // This property allows OS package maintainers to disable signing
+    val enableSigning = project.findProperty("ktlint.publication.signing.enable") != "false"
     sign(publishing.publications["maven"])
-    isRequired = !version.toString().endsWith("SNAPSHOT")
+    isRequired = enableSigning && !version.toString().endsWith("SNAPSHOT")
 }


ktlint/build.gradle.kts
```diff
@@ -48,9 +48,13 @@ val shadowJarExecutable by tasks.registering(DefaultTask::class) {
     description = "Creates self-executable file, that runs generated shadow jar"
     group = "Distribution"
 
+    val disableSigning = (project.findProperty("ktlint.disable-signing") as? String).toBoolean()
```
Goooler, line 51:
```suggestion
    val disableSigning = project.findProperty("ktlint.disable-signing") == "true"
```

Goooler: I believe you just need to amend this flag to 
https://github.com/pinterest/ktlint/blob/ec953e68c788d9fd68a79176c864e66e7bd5303a/build-logic/src/main/kotlin/ktlint-publication.gradle.kts#L101

Goooler: Test this change with:
1. Add `ktlint.disable-signing=true` in your gradle.properties.
2. Update `VERSION_NAME` to `0.49.0`.
3. Run `./gradlew publishToMavenLocal`.

pschichtel: I tested for both a snapshot and a non-snapshot version:

* `./gradlew :ktlint:shadowJarExecutable` (default behavior)
* `./gradlew :ktlint:shadowJarExecutable -Pktlint.disable-signing=false`  (explicitly enabled)
* `./gradlew :ktlint:shadowJarExecutable -Pktlint.disable-signing=true`  (explicitly disabled)

Goooler: I meant you just need to update the `isRequired`.

pschichtel: ah that's what you meant. makes sense.

Goooler (approved):

build-logic/src/main/kotlin/ktlint-publication.gradle.kts
```diff
@@ -97,6 +97,7 @@ signing {
     val signingPassword = System.getenv("ORG_GRADLE_PROJECT_signingKeyPassword")
     useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
 
+    val disableSigning = project.findProperty("ktlint.disable-signing") == "true"
```
paul-dingemans, line 100: Add comment and explain why and for who this property is added. This line is very vulbnerable for removal in the future as it is not used internally in ktlint. Also I would like change the name and behavior of the property to `ktlint.publication.signing.enable`. When the property is absent or set to another value then "false" it should be considered as enabled.

paul-dingemans (changes requested):
Nice an simple solution. Please address remark.

build-logic/src/main/kotlin/ktlint-publication.gradle.kts
pschichtel: done and tested

pschichtel:
The test failures seem unrelated.

PR #1871, opened 2023-03-19
Add cast extensions to replace std libs functions

Goooler:
## Description

> Usage of this function is unsafe because it does not have native compiler support
This means that compiler won't report UNCHECKED_CAST, CAST_NEVER_SUCCEED or similar
diagnostics in case of error cast (which can happen immediately or after some
refactoring of class hierarchy)
Consider using regular `as` and `as?` in Kotlin 1.8.20

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:


ktlint-logger/src/main/kotlin/com/pinterest/ktlint/logger/util/Any.kt
```diff
@@ -0,0 +1,5 @@
+package com.pinterest.ktlint.logger.util
+
+public inline fun <reified T : Any> Any?.safeAs(): T? = this as? T
+
+public inline fun <reified T : Any> Any?.cast(): T = this as T
```
paul-dingemans, line 5: The logger module should not be abused to expose those functions. Better expose them in module `ktlint-rule-engine-core` (also see https://pinterest.github.io/ktlint/api/overview/) in package `com/pinterest/ktlint/rule/engine/core/api`. And, please provide a KDoc for each method.

Goooler: Done. I put them here cause I didn't notice `ktlint-ruleset-standard` depend on `ktlint-rule-engine-core`.

paul-dingemans (approved):

PR #1884, opened 2023-03-24
Polish Gradle integration docs

Goooler:
## Description

- Defer task configurations.
- Rename `ktlint` task to `ktlintCheck`.
- More idiomatic Gradle dsl.

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/docs/install/integrations.md
+++ b/docs/install/integrations.md
@@ -129,7 +129,7 @@ dependencies {
     // ktlint will pick them up
 }
 
-tasks.register("ktlint", JavaExec) {
+tasks.register("ktlintCheck", JavaExec) {
     group = "verification"
     description = "Check Kotlin code style."
     classpath = configurations.ktlint
@@ -139,7 +139,7 @@ tasks.register("ktlint", JavaExec) {
 }
 
 tasks.named("check") {
-    dependsOn tasks.named("ktlint")
+    dependsOn tasks.named("ktlintCheck")
 }
 
 tasks.register("ktlintFormat", JavaExec) {
@@ -176,7 +176,7 @@ dependencies {
     // ktlint(project(":custom-ktlint-ruleset")) // in case of custom ruleset
 }
 
-tasks.register<JavaExec>("ktlint") {
+val ktlintCheck by tasks.registering(JavaExec::class) {
     group = LifecycleBasePlugin.VERIFICATION_GROUP
     description = "Check Kotlin code style"
     classpath = ktlint
@@ -186,6 +186,10 @@ tasks.register<JavaExec>("ktlint") {
         "**.kts",
         "!**/build/**",
     )
+}
+
+tasks.check {
+    dependsOn(ktlintCheck)
 }
 
 tasks.register<JavaExec>("ktlintFormat") {
--- a/docs/install/integrations.md
+++ b/docs/install/integrations.md
@@ -134,8 +134,8 @@ tasks.register("ktlintCheck", JavaExec) {
     description = "Check Kotlin code style."
     classpath = configurations.ktlint
     mainClass = "com.pinterest.ktlint.Main"
-    args "src/**/*.kt", "**.kts", "!**/build/**"
     // see https://pinterest.github.io/ktlint/install/cli/#command-line-usage for more information
+    args "src/**/*.kt", "**.kts", "!**/build/**"
 }
 
 tasks.named("check") {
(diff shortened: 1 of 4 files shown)


docs/install/integrations.md
```diff
@@ -127,20 +129,26 @@ dependencies {
...
+}
+
+tasks.register("ktlintFormat", JavaExec) {
+    group = "formatting"
     description = "Fix Kotlin code style deviations."
     classpath = configurations.ktlint
-    mainClass.set("com.pinterest.ktlint.Main")
-    args "-F", "src/**/*.kt"
+    mainClass = "com.pinterest.ktlint.Main"
+    jvmArgs "--add-opens=java.base/java.lang=ALL-UNNAMED"
```
Goooler, line 150: I'm not sure why there are 2 args in Maven integration but 1 in Gradle's,

https://github.com/pinterest/ktlint/blob/8757eb6b99c66d83d2d160545eb0a1a45f62fc4c/docs/install/integrations.md?plain=1#L25-L26

Goooler: I just see `--add-opens java.base/java.lang=ALL-UNNAMED` in [gantsign/ktlint-maven-plugin](https://github.com/gantsign/ktlint-maven-plugin#using-java-17-and-later), and this should be used >= Java 16?

paul-dingemans: After some digging around, I found only [this reference](https://github.com/pinterest/ktlint/issues/1195#issuecomment-895034955) in the issue tracker but that didn't lead to any code change. The `--add-opens` for `java.base/java.util` was added in this [commit](https://github.com/pinterest/ktlint/commit/8dd55c4718edb55cb1f34e4d422a612ec2424c6c#diff-28fa7445d504e87ed9fcf4e359d8d6858a9ca1ea6a12cc5b917b9a8e644ca858R130). Most likely this was based on the example in [Gradle documentation](https://docs.gradle.org/7.5/userguide/upgrading_version_7.html#removes_implicit_add_opens_for_test_workers). I have reused the change in the `CommandLineTestRunner` and added both `--add-opens` to the documentation you mentioned above. I will create a PR to validate whether all keeps working without the `add-opens` for `java.base/java.util`.

Goooler: Yeah, I believe 2 args are only needed in tests, but 1 arg is for formatting tasks.

docs/install/integrations.md
```diff
@@ -127,20 +129,26 @@ dependencies {
...
 
-task ktlint(type: JavaExec, group: "verification") {
+tasks.register("ktlintCheck", JavaExec) {
+    group = "verification"
     description = "Check Kotlin code style."
     classpath = configurations.ktlint
-    mainClass.set("com.pinterest.ktlint.Main")
-    args "src/**/*.kt"
+    mainClass = "com.pinterest.ktlint.Main"
+    args "src/**/*.kt", "!**/build/**"
```
paul-dingemans, line 137: Also add "**.kts" like on line 68?

docs/install/integrations.md
```diff
@@ -127,20 +129,26 @@ dependencies {
...
+
+tasks.register("ktlintFormat", JavaExec) {
+    group = "formatting"
     description = "Fix Kotlin code style deviations."
     classpath = configurations.ktlint
-    mainClass.set("com.pinterest.ktlint.Main")
-    args "-F", "src/**/*.kt"
+    mainClass = "com.pinterest.ktlint.Main"
+    jvmArgs "--add-opens=java.base/java.lang=ALL-UNNAMED"
+    args "-F", "src/**/*.kt", "!**/build/**"
```
paul-dingemans, line 151: Also add "**.kts" like on line 68?

docs/install/integrations.md
```diff
@@ -169,29 +177,36 @@ dependencies {
...
+    group = LifecycleBasePlugin.VERIFICATION_GROUP
+    description = "Check Kotlin code style"
     classpath = ktlint
     mainClass.set("com.pinterest.ktlint.Main")
     // see https://pinterest.github.io/ktlint/install/cli/#command-line-usage for more information
-    args = listOf("src/**/*.kt")
+    args(
+        "**/src/**/*.kt",
+        "**.kts",
+        "!**/build/**",
```
paul-dingemans, line 189: This set of patterns is used on multiple places now. It might be better to extract this to a shared constant.

Goooler: Yeah, but they are examples in docs, might be better not to reuse.

docs/install/integrations.md
Goooler: We can.

Goooler: 78de319793b7d9fab35a6bae409d65981d12ffa7

docs/install/integrations.md
Goooler: 78de319793b7d9fab35a6bae409d65981d12ffa7

docs/install/integrations.md
paul-dingemans: Indeed, I did not realize that.

docs/install/integrations.md
paul-dingemans: Yes, lets verify and fix in #1886

PR #1892, opened 2023-03-25
Reuse common configurations

Goooler:
## Checklist

- [ ] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/.github/workflows/gradle-pr-build.yml
+++ b/.github/workflows/gradle-pr-build.yml
@@ -35,10 +35,3 @@ jobs:
         run: ./gradlew build ktlintCheck
       - name: Build with dev Kotlin version
         run: ./gradlew -PkotlinDev build ktlintCheck
-      - name: Upload test results
-        uses: actions/upload-artifact@v3
-        if: failure()
-        with:
-          name: Test artifacts
-          path: '**/build/reports/tests/**'
-          if-no-files-found: warn
--- a/build-logic/src/main/kotlin/ktlint-publication.gradle.kts
+++ b/build-logic/src/main/kotlin/ktlint-publication.gradle.kts
@@ -5,6 +5,11 @@ plugins {
     signing
 }
 
+if (project.hasProperty("isKotlinDev")) {
+    val definedVersion = ext["VERSION_NAME"].toString().removeSuffix("-SNAPSHOT")
+    ext["VERSION_NAME"] = "$definedVersion-kotlin-dev-SNAPSHOT"
+}
+
 project.version = project.property("VERSION_NAME")
     ?: throw GradleException("Project version property is missing")
 project.group = project.property("GROUP")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -11,20 +11,6 @@ allprojects {
         val definedVersion = ext["VERSION_NAME"].toString().removeSuffix("-SNAPSHOT")
         ext["VERSION_NAME"] = "$definedVersion-kotlin-dev-SNAPSHOT"
     }
-
-    val skipTests: String = System.getProperty("skipTests", "false")
-    tasks.withType<Test>().configureEach {
-        if (skipTests == "false") {
-            useJUnitPlatform()
-        } else {
-            logger.warn("Skipping tests for task '$name' as system property 'skipTests=$skipTests'")
-        }
-
-        if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_16)) {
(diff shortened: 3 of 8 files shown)


Goooler:
Depends on #1891.

.github/workflows/gradle-pr-build.yml
```diff
@@ -35,10 +35,3 @@ jobs:
         run: ./gradlew build ktlintCheck
       - name: Build with dev Kotlin version
         run: ./gradlew -PkotlinDev build ktlintCheck
-      - name: Upload test results
-        uses: actions/upload-artifact@v3
-        if: failure()
-        with:
-          name: Test artifacts
-          path: '**/build/reports/tests/**'
-          if-no-files-found: warn
```
Goooler, line 44: I'm removing this action, you can view test results via Gradle build scan.

Goooler, line 44: I'm removing this action, you can view test results via Gradle build scan.

Goooler, line 44: I'm removing this action, you can view test results via Gradle build scan for now.

ktlint-api-consumer/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
 dependencies {
     // Any SLF4J compatible logging framework can be used. The "slf4j-simple" logging provider is configured in file
     // ktlint-api-consumer/src/main/resources/simplelogger.properties
-    runtimeOnly("org.slf4j:slf4j-simple:2.0.7")
+    runtimeOnly(libs.slf4j)
```
paul-dingemans, line 8: Please revert this change. I intended this module to be an easy example of how an API consumer can look like. So the slf4j dependency is not meant to be used anywhere else in the project.

Maybe something similar should also be done for line 10 and 11 as well as for other modules. We might need to distinct between modules meant for publication as artifact (e.g. being a part of the Ktlint API or CLI) and other modules that are only meant for clarification. The latter group of modules should not reuse any of the common build logic but be a standalone build script. What do you think?

Goooler: cb9b94a0c80157fa858c89ece8e8357ecac50bd4

.github/workflows/gradle-pr-build.yml
paul-dingemans: I am not liking this. I don't like to fill in my email adress each time that I want to get access to a build scan. I am developing using a laptop from my employer. I don't want to use my work email adress to get access to the scan. Neither do I want to access my personal email on my work laptop.

What would be the benefit of removing it?

Goooler: It's easier to view and share build scans, I noticed I have no need to fill email address to view build scans that had been published on CI, you can have a try like https://gradle.com/s/tcalcngp6a3tk.

paul-dingemans: Yes, I just noticed that too in another failed build. And in no way, I meant that the Gradle build scan had to be removed ;-)
So let try it out then.

build.gradle.kts
```diff
@@ -6,27 +6,6 @@ plugins {
     alias(libs.plugins.shadow) apply false
 }
 
-allprojects {
-    if (project.hasProperty("isKotlinDev")) {
-        val definedVersion = ext["VERSION_NAME"].toString().removeSuffix("-SNAPSHOT")
-        ext["VERSION_NAME"] = "$definedVersion-kotlin-dev-SNAPSHOT"
-    }
```
Goooler, line 13: Killing `allprojects`.

build-logic/src/main/kotlin/ktlint-publication.gradle.kts
```diff
@@ -5,6 +5,11 @@ plugins {
     signing
 }
 
+if (project.hasProperty("isKotlinDev")) {
+    val definedVersion = ext["VERSION_NAME"].toString().removeSuffix("-SNAPSHOT")
+    ext["VERSION_NAME"] = "$definedVersion-kotlin-dev-SNAPSHOT"
+}
```
Goooler, line 11: All modules need to be published depending on `ktlint-publication.gradle.kts`.

ktlint-api-consumer/build.gradle.kts
Goooler: I still we should put this dep into version catalog, we should not reuse publication logic but deps.

PR #1900, opened 2023-03-29
Execute tests in parallel

Goooler:
## Description

https://docs.gradle.org/8.0/userguide/performance.html#execute_tests_in_parallel

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
+++ b/build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
@@ -31,6 +31,13 @@ tasks.withType<Test>().configureEach {
         logger.warn("Skipping tests for task '$name' as system property 'skipTests=$skipTests'")
     }
 
+    maxParallelForks = if (System.getenv("CI") != null) {
+        Runtime.getRuntime().availableProcessors()
+    } else {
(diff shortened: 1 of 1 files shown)


Goooler:
Less CLI tests time in 

https://scans.gradle.com/s/ygma7eqwlhroa/timeline?sort=longest#jileod76e2zko

vs 

https://scans.gradle.com/s/ntsgtpeogqgq4/timeline?sort=longest#jileod76e2zko

paul-dingemans:
Interesting. Lets see what happens when we enable it.

There is another big improvement that we can make although it might result in flaky test if it is narrowed down too much. The CommandLineTestRunner terminates a process of a CLI tests after a maximum duration. This is introduced because sometimes the forked process is not terminated correctly (especially for Windows) although ktlint was already completed. When comparing the individual CLI tests between Ubuntu and Windows (https://scans.gradle.com/s/io7srx42twl5g/tests/overview?help=total-time) you can see that on Ubuntu it takes around 4 seconds while on Windows it is around 10 seconds. The 10 seconds matches with https://github.com/pinterest/ktlint/blob/master/ktlint-cli/src/test/kotlin/com/pinterest/ktlint/cli/CommandLineTestRunner.kt#L269.

PR #1968, opened 2023-04-22
Disable the gradle configuration cache

paul-dingemans:
## Description

The configuration cache is not yet supported by all tasks used in the build pipelines. For all github workflows the configuration-cache was already disabled. But also on the local machine, this now causes a simple task like `gradlew test` to fail as the `shadowJar` task is invoked which fails with error:
```
- Task `:ktlint-cli:shadowJarExecutable` of type `org.gradle.api.DefaultTask`: cannot serialize Gradle script object references as these are not supported with the configuration cache.
```

## Checklist

- [X] PR description added
- [ ] tests are added

In case of adding a new rule:


paul-dingemans:
@Goooler FYI

PR #1985, opened 2023-04-28
Move Baseline into ktlint-baseline module

shashachu:
## Description
Currently the `Baseline` class is in `ktlint-cli` which doesn't publish an artifact to maven. Moved into a separate published artifact for API consumers.

## Checklist

- [x] PR description added


ktlint-baseline/gradle.properties
```diff
@@ -0,0 +1,2 @@
+POM_NAME=ktlint-cli-reporter
+POM_ARTIFACT_ID=ktlint-cli-reporter
```
wakingrufus, line 2: Are these the right values?

shashachu: oooops. obviously copy/paste. thanks.

paul-dingemans:
Note to self: change architecture diagram in docs

PR #1989, opened 2023-05-05
Delete redundant "org.gradle.configuration-cache=false"

tatsuyafujisaki:
## Description
This pull request deletes the redundant `org.gradle.configuration-cache=false`, which is default.

> By default, Gradle does not use the configuration cache.

https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:usage:enable

## Checklist

- [x] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,4 +20,3 @@ POM_DEVELOPER_NAME=Pinterest, Inc.
 org.gradle.jvmargs=-Xmx4g
 org.gradle.parallel=true
 org.gradle.caching=true
-org.gradle.configuration-cache=false


paul-dingemans:
Is there any specific reason, other than its value is equal to the default value, why you want to delete the setting?

It was left on purpose (but should have been commented) as we learned that enabling it blocked the release process of the `0.49` release.

tatsuyafujisaki:
No. I just thought there was no reason to specify the default.
Now I see your point.
Thanks for the clarification.

Issue #1938, opened 2023-04-12
Statement may not start or end on same line as start/end of block


paul-dingemans:
Given code sample below:
```
fun foo1() { if (true) {
        // do something
    }
}

fun foo2() {
    if (true) {
        // do something
    } }
```
then it should be formatted as:
```
fun foo1() {
    if (true) {
        // do something
    }
}

fun foo2() {
    if (true) {
        // do something
    }
}
```

This issue should be covered by same (new) rule as for #1078

atulgpt:
Hi, @paul-dingemans I have added my draft PR for early review. Given this is my first PR something might be wrong 😅

atulgpt:
Hi, @paul-dingemans for adding documentation and updating CHANGELOG.md is there any documentation?

paul-dingemans:
> Hi, @paul-dingemans I have added my draft PR for early review. Given this is my first PR something might be wrong 😅
> 
> Also let me know if it needs to handle all the block codes, like blocks of `class`, `object`, `when`, and `if`. I am not sure if some existing rules already handle those

Handle this generically for all blocks. I would expect that if you just implement this for element type `BLOCK` that it covers all of that. But please do add sufficient unit tests to validate.

> Hi, @paul-dingemans for adding documentation and updating CHANGELOG.md is there any documentation?

There are no docs about how to document.

The `CHANELOG.DOC` contains a section `Unreleased` at the top where you have to add the rule under sub section `Added`. See older release notes for examples.

In sub directory `docs` you will find the documentation which is published on https://pinterest.github.io/ktlint/. As a new rul should be marked as `Rule.Experimental` it should also be documented in file `experimental.md`. Please, follow the same approach as for other rule.

PR #2001, opened 2023-05-07
Add rule for braces and code alignment

atulgpt:
## Description

Add rule such that code in function block should have new line in between braces and code

Closes https://github.com/pinterest/ktlint/issues/1938

## Checklist

- [x] PR description added
- [x] tests are added

In case of adding a new rule:
- [x] Rule is added to [rules documentation](https://pinterest.github.io/ktlint/rules/standard/)


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.pinterest.ktlint.rule.engine.core.api.noNewLineInClosedRange
+import com.pinterest.ktlint.rule.engine.core.api.prevCodeLeaf
+import com.pinterest.ktlint.rule.engine.core.api.upsertWhitespaceAfterMe
+import com.pinterest.ktlint.rule.engine.core.api.upsertWhitespaceBeforeMe
+import com.pinterest.ktlint.ruleset.standard.StandardRule
+import org.jetbrains.kotlin.com.intellij.lang.ASTNode
+
+public class NoSameLineStatementAsOfBlockStartOrEnd
+    : StandardRule(
+    "no-same-line-statement-as-of-block-start-or-end",
```
paul-dingemans, line 19: How about the shorter name `statement-wrapping`? Please, also rename class.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
```diff
@@ -0,0 +1,90 @@
...
+                tabWidth = editorConfig[INDENT_SIZE_PROPERTY],
+            )
+    }
+
+    override fun beforeVisitChildNodes(
+        node: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit,
+    ) {
+        if (node.elementType == ElementType.FUN) {
```
paul-dingemans, line 40: Do not restrict it to function bodies. I think, it is safe to apply this on any block. But it might be needed to add an exception to this, which would allow a block starting and ending on the same line. For example in definition of a function literal:
```
val foo = { bar() }
```

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
```diff
@@ -0,0 +1,90 @@
...
+            node.findChildByType(ElementType.BLOCK)?.let { blockAstNode ->
```
paul-dingemans, line 41: I prefer to wrap chained methods. It is okay to have one dot qualified call on an line. In case there are more, I put each dot on a newline. So this would become:
```
node
    .findChildByType(ElementType.BLOCK)
    ?.let { blockAstNode ->
    }
```
```

File: ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt [hunk 4]
```diff
@@ -0,0 +1,90 @@
...
+                val lBrace = blockAstNode.findChildByType(ElementType.LBRACE)!!
+                val rBrace = blockAstNode.findChildByType(ElementType.RBRACE)!!
+                val lBraceNextCodeElement = lBrace.nextCodeLeaf()!!
+                val rBracePrevCodeElement = rBrace.prevCodeLeaf()!!
+                val text = blockAstNode.text
+                val indexOfLBrace = text.indexOf("{")
+
+                assert(indexOfLBrace >= 0)
```
paul-dingemans, line 49: This is unneeded. II have never seen an LRBACE element having other content than `{`.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
```diff
@@ -0,0 +1,90 @@
...
+
+                val indexOfRBrace = text.lastIndexOf("}")
+                assert(indexOfRBrace >= 0)
+
+                if (lBraceNextCodeElement.elementType == ElementType.RBRACE) {
+                    return
+                }
```
paul-dingemans, line 56: I think this is to cover an empty block `{}`.  You could incorporate it in the outer check:
```
node
    .findChildByType(ElementType.BLOCK)
    .takeUnless { it.isEmptyBlock() } // function isEmptyBlock needs to be declared of course
    ?.let { blockAstNode ->
    }
```

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
```diff
@@ -0,0 +1,90 @@
...
+
+                if (noNewLineInClosedRange(lBrace, lBraceNextCodeElement)) {
+                    emit(
+                        lBraceNextCodeElement.startOffset,
+                        "Expected new line after `{` of function body", true
+                    )
```
paul-dingemans, line 62: Keep the emit on a single line when possible.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
```diff
@@ -0,0 +1,90 @@
...
+                assert(indexOfRBrace >= 0)
+
+                if (lBraceNextCodeElement.elementType == ElementType.RBRACE) {
+                    return
+                }
+
+                if (noNewLineInClosedRange(lBrace, lBraceNextCodeElement)) {
+                    emit(
+                        lBraceNextCodeElement.startOffset,
+                        "Expected new line after `{` of function body", true
```
paul-dingemans, line 61: Use single (or double) quotes instead of backticks inside the message.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
```diff
@@ -0,0 +1,90 @@
...
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit,
+    ) {
+        if (node.elementType == ElementType.FUN) {
+            node.findChildByType(ElementType.BLOCK)?.let { blockAstNode ->
+                val lBrace = blockAstNode.findChildByType(ElementType.LBRACE)!!
+                val rBrace = blockAstNode.findChildByType(ElementType.RBRACE)!!
+                val lBraceNextCodeElement = lBrace.nextCodeLeaf()!!
+                val rBracePrevCodeElement = rBrace.prevCodeLeaf()!!
+                val text = blockAstNode.text
```
paul-dingemans, line 46: I am not a fan of declaring a lot of variables upfront. Also the `!!` I try to avoid. Most often, I just constructs like
```
blockAstNode
    .findChildByType(ElementType.LBRACE)
   ?.let { lbrace ->
   }
```

paul-dingemans (changes requested):
Tnx for your contribution. I can see that you really picked up a lot of conventions that are used in the project. Well done and thanks for that. Most remarks are about little things that you missed. From a consistency perspective, I would like them to be refactored.

Please make sure to run gradle tasks `test` and `ktlintFormat` before pushing commits. This will reveal that your rule is not yet added to the `StandardRuleSetProvider` and fixes formatting errors.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
atulgpt: Do we honour the max line?

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
atulgpt: That will create two level of nesting as I am using `rBrace` and `rBracePrevCodeElement` both and hence two `let` calls will be required

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
atulgpt: What about
```
fun foo() { println() }
```
should we allow it? Currently, I have disallowed it. But if we are allowing above we should also allow single line functions with block

paul-dingemans: It should be disallowed. The example `val foo = { bar() }` is a function literal and will not have element type 'BLOCK'. The IntelliJ Plugin `PsiViewer` is a very helpful too to check how the AST of a code sample is constructed.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
paul-dingemans: Yes, that why I most often use that construct. I think for the lbrace the let-construction should be used. Starting from that point you know that you are inside a BLOCK that started with a LBRACE. This also means that RBRACE should exist (provided that the code could be compiled). So for RBRACE the `!!` or a `requireNotNull` can be used.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
paul-dingemans: Yes mostly. Sometimes in test I suppress max line length for the lint violations.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
atulgpt: Got it. Basically, we will allow single-line function literal but will NOT allow single-line function body. Thanks for the clarification.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
atulgpt: In this case, keeping emit in a single line was crossing the max len limit. Do we ignore max len for emit function to avoid clutter in PR?

paul-dingemans: No, but then I would have expected the ktlintFormat task to put each parameter on a distinct line while in your code the second and third parameter are on the same line.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
atulgpt: Update

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
atulgpt: Updated

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/NoSameLineStatementAsOfBlockStartOrEnd.kt
atulgpt: Used `requireNotNull`

atulgpt:
Hi, @paul-dingemans I have updated the PR with documentation, TCs with other cases. Please re-review

paul-dingemans:
I have refactored the rule and test quite a bit. The base however was solid and made it more easy for me to do this. So your effort is still appreciated.

atulgpt:
Thanks @paul-dingemans. Let me know if there is more Up for grabs issue

paul-dingemans:
> Thanks @paul-dingemans. Let me know if there is more Up for grabs issue

Would be cool if you want to help out more. It would be best if you connect to me on the [kotlinlang](https://kotlinlang.slack.com/archives/CKS3XG0LS) slack, or have a look at https://github.com/pinterest/ktlint/issues?q=is%3Aopen+is%3Aissue+label%3Aup-for-grabs.

paul-dingemans (approved):

paul-dingemans (approved):

PR #2003, opened 2023-05-09
Fix interchanged sample code

avinash3699:
## Description

It appears that the sample codes provided for the rule ID "no-empty-first-line-in-method-block" in [standard.md](https://github.com/pinterest/ktlint/blob/master/docs/rules/standard.md#no-leading-empty-lines-in-method-blocks) have been mistakenly interchanged, which could lead to confusion for viewers.


paul-dingemans:
Tnx for catching and fixing this mistake.

Issue #1981, opened 2023-04-26
Remove usages of ExtensionsArea.registerExtensionPoint(java.lang.String, java.lang.String, com.intellij.openapi.extensions.ExtensionPoint.Kind)


goodwinnk:
The method is deprecated and going to be deleted in the IntelliJ platform. 

https://github.com/JetBrains/intellij-community/blob/3d5848904794db7cbff9473e2fea7e9689e251e9/platform/extensions/src/com/intellij/openapi/extensions/ExtensionsArea.java#L19

Once it is deleted from the platform and after Kotlin updates the platform, the klint will fail on runtime.

paul-dingemans:
Tnx for the heads up. Can you point me in a direction how to mitigate this? I am not familiar with the piece of code in Ktlint where this is invoked.  From the comments I do understand that we need it for the `format` (autocorrect) mode which is quite an essential piece of functionality.

goodwinnk:
> Can you point me in a direction how to mitigate this?

We did the following [replacement](https://github.com/JetBrains/kotlin/commit/3caa1d13492fef61a47132c17475dcde02a47624) in the Kotlin.

There was a recent bootstrap *1.9.0-dev-6976* with the upgrade of the platform in the https://maven.pkg.jetbrains.space/kotlin/p/kotlin/bootstrap repository, where this fix can be tested. I think such test might reveal more problems, unfortunately.

Currently, there is no guarantee of binary compatibility in the Kotlin compiler, especially for third-party classes which are not present in the Kotlin repository and packed into *kotlin-compiler-embeddable*. With the evolution of the Kotlin compiler, I anticipate that there will be many more compatibility issues.

As a temporary measure, I suggest being very explicit about all undocumented usages (even proposed `CoreApplicationEnvironment.registerExtensionPoint`) that are in the core of the project, and raise issues in [YouTrack](https://youtrack.jetbrains.com/issues/KT) asking Kotlin team to document and support them.

paul-dingemans:
Tnx for your reply which was very helpful. I have analyzed the impact on Ktlint and reported below as https://youtrack.jetbrains.com/issue/KT-58704/Support-and-document-extension-point-org.jetbrains.kotlin.com.intellij.treeCopyHandler

===

[...]

PR #2044, opened 2023-05-19
Disable extension point org.jetbrains.kotlin.com.intellij.treeCopyHandler

paul-dingemans:
## Description

In Kotlin 1.9 the extension points of the embedded kotlin compiler will change. Ktlint only uses the `org.jetbrains.kotlin.com.intellij.treeCopyHandler` extension point. This extension is not yet supported in 1.9, neither is it documented ([#KT-58704](https://youtrack.jetbrains.com/issue/KT-58704/Support-and-document-extension-point-org.jetbrains.kotlin.com.intellij.treeCopyHandler)). Without this extension point it might happen that your custom rules will throw exceptions during runtime. See [#1981](https://github.com/pinterest/ktlint/issues/1981).

In Ktlint, 7 out of 77 rules needed small and sometimes bigger changes to become independent of the extension point `org.jetbrains.kotlin.com.intellij.treeCopyHandler`. The impact on your custom rules may vary dependent on the way the autocorrect has been implemented. When manipulating `ASTNode`s there seems to be no impact. When, manipulating `PsiElement`s, some functions consistently result in a runtime exception.

Based on the refactoring of the rules as provided by `ktlint-ruleset-standard` in Ktlint `0.49.x` the suggested refactoring is as follows:

* Replace `LeafElement.replaceWithText(String)` with `LeafElement.rawReplaceWithText(String)`.
* Replace `PsiElement.addAfter(PsiElement, PsiElement)` with `AstNode.addChild(AstNode, AstNode)`. Note that this method inserts the new node (first) argument *before* the second argument node and as of that is not a simple replacement of the `PsiElement.addAfter(PsiElement, PsiElement)`.
* Replace `PsiElement.replace(PsiElement)` with a sequence of `AstNode.addChild(AstNode, AstNode)` and `AstNode.removeChild(AstNode)`.

Be aware that your custom rules might use other functions which also throw exceptions when the extension point `org.jetbrains.kotlin.com.intellij.treeCopyHandler` is no longer supported.

In order to help you to analyse and fix the problems with your custom rules, ktlint temporarily supports to disable the extension point `org.jetbrains.kotlin.com.intellij.treeCopyHandler` using a flag. This flag is available in the Ktlint CLI and in the `KtlintRuleEngine`. By default, the extension point is enabled like it was in previous versions of ktlint.

At least you should analyse the problems by running your test suits by running ktlint and disabling the extension point. Next you can start with fixing and releasing the updated rules. All rules in this version of ktlint have already been refactored and are not dependent on the extension point anymore. In Ktlint CLI the flag is to be activated with parameter `--disable-kotlin-extension-point`. API Consumers that use the `KtlintRuleEngine` directly, have to set property `enableKotlinCompilerExtensionPoint` to `false`.

At this point in time, it is not yet decided what the next steps will be. Ktlint might drop the support of the extension points entirely. Or, if the extension point `org.jetbrains.kotlin.com.intellij.treeCopyHandler` is fully supported at the time that ktlint will be based on kotlin 1.9 it might be kept. In either case, the flag will be dropped in a next version of ktlint.
## Checklist

- [X] PR description added
- [X] tests are added

In case of adding a new rule:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,29 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Unreleased
 
+### Custom Rule Providers need to prepare for Kotlin 1.9
+
+In Kotlin 1.9 the extension points of the embedded kotlin compiler will change. Ktlint only uses the `org.jetbrains.kotlin.com.intellij.treeCopyHandler` extension point. This extension is not yet supported in 1.9, neither is it documented ([#KT-58704](https://youtrack.jetbrains.com/issue/KT-58704/Support-and-document-extension-point-org.jetbrains.kotlin.com.intellij.treeCopyHandler)). Without this extension point it might happen that your custom rules will throw exceptions during runtime. See [#1981](https://github.com/pinterest/ktlint/issues/1981).
+
+In Ktlint, 7 out of 77 rules needed small and sometimes bigger changes to become independent of the extension point `org.jetbrains.kotlin.com.intellij.treeCopyHandler`. The impact on your custom rules may vary dependent on the way the autocorrect has been implemented. When manipulating `ASTNode`s there seems to be no impact. When, manipulating `PsiElement`s, some functions consistently result in a runtime exception.
+
+Based on the refactoring of the rules as provided by `ktlint-ruleset-standard` in Ktlint `0.49.x` the suggested refactoring is as follows:
+
+* Replace `LeafElement.replaceWithText(String)` with `LeafElement.rawReplaceWithText(String)`.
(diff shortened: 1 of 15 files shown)


shashachu:
I'm wondering if it would be possible to log instead of require the user to opt in to use a flag. I'm thinking of Gradle warnings in the form of: `You are using treeCopyHandler. This will be unsupported in Kotlin 1.9.0 ....'

paul-dingemans:
It is not about opt-in by end users. If kotlin 1.9 does not support the extension point then the rule provider is responsible for making the rule compatible with 1.9 without falling back on this extension point. The idea of this change is that rule providers can check the vulnerability for failures in their rule whenever we upgrade to kotlin 1.9.


For normal users of ktlint this functionality has to be invisible. That is why the Ktlint cli parameter is hidden from the help command.

PR #2056, opened 2023-05-24
Add example of suppressing a custom rule (documents that "ktlint:" pr…

plcj:
…efix is required)

## Description

Add example of suppressing a custom rule to "How to suppres..." section of FAQ. It is done to document this feature and show that the "ktlint:" prefix is required.

## Checklist

- [ ] PR description added
- [ ] tests are added

In case of adding a new rule:

### DIFF
--- a/documentation/snapshot/docs/faq.md
+++ b/documentation/snapshot/docs/faq.md
@@ -68,6 +68,10 @@ An error can be suppressed using:
     // Suppress all rules for the annotated construct
     @Suppress("ktlint")
     import foo.*
+
+    // Suppress a single custom rule for the annotated construct
+    @Suppress("ktlint:custom-ruleset-id:custom-rule")
+    import foo.*
     ```
 === "[:material-heart:](#) EOL comments"
 


paul-dingemans (commented):
Adding this example does not add much value as there are already some comparable examples just a few line above:
```
// Suppress a single rule for the annotated construct
@Suppress("ktlint:standard:no-wildcard-imports")
import foo.*

// Suppress multiple rules for the annotated construct
@Suppress("ktlint:standard:no-wildcard-imports", "ktlint:standard:other-rule-id")
import foo.*
```
But the reference to the custom rule set can be mixed into this example as follows:
```
// Suppress a single rule (with id 'rule-id', defined in rule set with id 'rule-set-id') from the annotated construct
@Suppress("ktlint:rule-set-id:rule-id")
class Foo {}

// Suppress multiple rules for the annotated construct
@Suppress("ktlint:standard:no-wildcard-imports", "ktlint:custom-rule-set-id:custom-rule-id")
import foo.*
```

Would that be clear enough?

plcj:
Yeah that would do it. I just needed an example with a custom rule set to understand that all suppressions must be prefixed with "ktlint:" (or a statement to that effect). I pushed that change.

PR #2062, opened 2023-05-28
Mark `BlockCommentInitialStarAlignmentRule` and `StringTemplateIndentRule` as `RunAsLateAsPossible`

chao2zhang:
## Description

`BlockCommentInitialStarAlignmentRule` and `StringTemplateIndentRule` have to run after `IndentationRule`. And `IndentationRule` is marked as `RunAsLateAsPossible`.

`BlockCommentInitialStarAlignmentRule` and `StringTemplateIndentRule` should better be marked as `RunAsLateAsPossible`. This is to make the algorithm implementation to infer the run order to be easy to reason about

## Checklist

- [ ] PR description added
- [ ] tests are added

In case of adding a new rule:


paul-dingemans:
Please provide unit tests that show that the rule order needs to be changed and provide better output.

Those unit tests also serve as guard against future changes in rule sorting.

chao2zhang:
What would the role of `RunAsLateAsPossible` be if it is not used for rule sorting?

> Is there a specific reason that detekt wants to provide its own custom sorting order algorithm of the rules?
It's just a larger refactoring, and potentially be backwards incompatible. I will explore `KtLintRuleEngine`

paul-dingemans:
The rule sorter first runs all rules without any `VisitorModifier`.
The next group of rules are the rules marked with `RunAsLateAsPossible`.
The last group are the rules marked with `RunAfter`.

The Rule sorter is indeed an internal class as API consumers should only affect rule sorting by applying the `VisitorModifier`.

PR #2064, opened 2023-05-30
Updating `cli.md` to make slightly more clear on how to install this for windows

WimpyLigteBottel:
Adding extra info to help guide the user to installing ktlint on Windows.

This was confirmed on Windows 10 on my workstation.

### DIFF
--- a/documentation/snapshot/docs/install/cli.md
+++ b/documentation/snapshot/docs/install/cli.md
@@ -219,9 +219,6 @@ Options `--stdin` and `--patterns-from-stdin` are mutually exclusive, only one o
 ### Microsoft Windows users
 
 !!! tip "Microsoft Windows"
-    On Microsoft Windows you'll have to use `java -jar ktlint ...`.
-
-!!! tip "Microsoft Windows"
 
 1. You can download the file via wget on Windows or manually via [releases](https://github.com/pinterest/ktlint/releases)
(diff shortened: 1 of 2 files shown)


documentation/snapshot/docs/install/cli.md
```diff
@@ -220,3 +220,12 @@ Options `--stdin` and `--patterns-from-stdin` are mutually exclusive, only one o
 
 !!! tip "Microsoft Windows"
     On Microsoft Windows you'll have to use `java -jar ktlint ...`.
+
+!!! tip "Microsoft Windows"
```
WimpyLigteBottel, line 224: I see this is double, will need to fix this. as well as removing the extra jar file

paul-dingemans: Very recently, it is not yet released, we have also added a "ktlint.bat" file for windows to support Java 16+ usage  (#2059). Please update above to using the batch file. But also keep the "java -jar" for hard core CLI users that do not want to use a provided batch file.

paul-dingemans:
@WimpyLigteBottel Sorry, I rewrote your proposal. Please, let this not stop you from contributing again. Release docs should not have changed as `ktlint.bat` has not yet been released before. Also, I wanted to emphasis the use of the `ktlint.bat` by naming it as the first option. Als using `Git bash` on Windows is to be preferred above the `java -jar ktlint` command.

Issue #1078, opened 2021-02-11
Prevent multiple statements on the same line


hiqua:
## Duplicates
I could not find a matching issue, but I find it hard to believe that it doesn't exist yet. Apologies if this is indeed a duplicate.

## Expected Rule behavior
Do not allow multiple statements on the same line. ktlint does not complain with the following:
```kotlin
class MissingRules {
    fun f() {
        val a = 3; val b = 2
    }
}
```
Here, `a` and `b` should be declared separately.

## Additional information
* Current version of ktlint: 0.40.0
* Styleguide section: https://developer.android.com/kotlin/style-guide#one_statement_per_line
```
One statement per line
Each statement is followed by a line break. Semicolons are not used.
```

paul-dingemans:
I am not aware of any language constructs in Kotlin that use `;` as separator between elements.

atulgpt:
Hi, @paul-dingemans Should we disallow the semicolon itself? Even code like the below is possible in Kotlin. Which are not reported by `ktlint`
```
public fun f1() {
    val a = 3; val b = 2
}; public fun f2() {
    val a = 3; val b = 2
}
```
```
f1(); f2()
```
```
public class A {
    
}; public class B {
    
}
```
```
public class A; public class B
```

Issue #1953, opened 2023-04-18
Wrap chained calls


paul-dingemans:
Lines that contain too many chained calls should be wrapped. 

For example:
```
// 0
nonBlankLines.joinToString(separator = "") { it.substringAfter(":").lowercase(Locale.getAvailableLocales().firstOrNull() ?: Locale.getDefault()) }.toCharArray().distinct().count()

// 1
nonBlankLines
    .joinToString(separator = "") { it.substringAfter(":").lowercase(Locale.getAvailableLocales().firstOrNull() ?: Locale.getDefault()) }
    .toCharArray().distinct().count()

// 2
nonBlankLines
    .joinToString(separator = "") {
        it.substringAfter(":").lowercase(Locale.getAvailableLocales().firstOrNull() ?: Locale.getDefault())
    }
    .toCharArray().distinct().count()

// 3
nonBlankLines
    .joinToString(separator = "") {
        it.substringAfter(":").lowercase(Locale.getAvailableLocales().firstOrNull() ?: Locale.getDefault())
    }.toCharArray()
    .distinct()
    .count()

// 4
nonBlankLines
    .joinToString(separator = "") {
        it
            .substringAfter(":")
            .lowercase(
                Locale
                    .getAvailableLocales()
                    .firstOrNull()
                    ?: Locale.getDefault()
            )
    }.toCharArray()
    .distinct()
    .count()
```

I expect that only the last format can be considered deterministic although it migh be felt as way too stretched for others.

At least an exception should be made for a chained method call of only 1 level deep. Wrapping chained calls like "bar.foo()" does not make sense at all. But also "bar.foo().fooBar()" could be seen as acceptable in some contexts. So it should be configurable to what maximum amount of chained calls may be placed on a single line.

atulgpt:
Hi, @paul-dingemans I want to work on this as I believe this should be must have rule in the ktlint. At the same to me it seems to be very large rule(from impl point of view). Should we split it to smaller chunks? By when are we targeting this rule?

paul-dingemans:
If it can be split in smaller chunks, that is okay to me. It will be released in 0.50 if possible. Otherwise it will go with 1.0.

Can you draft the smaller chunks that you see?

atulgpt:
[...]

PR #2065, opened 2023-05-31
Fixes/1078/statement wrap after semi

atulgpt:
## Description

Add the `SEMICOLON` branch in `WrappingRule` and insert a newline after the semicolon

Fixes: https://github.com/pinterest/ktlint/issues/1078

## Checklist

- [x] PR description added
- [x] tests are added
- [x] [documentation](https://pinterest.github.io/ktlint/) is updated: NA as no new rule is added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ At this point in time, it is not yet decided what the next steps will be. Ktlint
 * Extract rule `no-single-line-block-comment` from `comment-wrapping` rule. The `no-single-line-block-comment` rule is added as experimental rule to the `ktlint_official` code style, but it can be enabled explicitly for the other code styles as well. ([#1980](https://github.com/pinterest/ktlint/issues/1980))
 * Clean-up unwanted logging dependencies ([#1998](https://github.com/pinterest/ktlint/issues/1998))
 * Fix directory traversal for patterns referring to paths outside of current working directory or any of it child directories ([#2002](https://github.com/pinterest/ktlint/issues/2002))
-* Prevent multiple expression in same line separated by semicolon ([#1078](https://github.com/pinterest/ktlint/issues/1078))
+* Prevent multiple expressions on same line separated by semicolon ([#1078](https://github.com/pinterest/ktlint/issues/1078))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ At this point in time, it is not yet decided what the next steps will be. Ktlint
 * Extract rule `no-single-line-block-comment` from `comment-wrapping` rule. The `no-single-line-block-comment` rule is added as experimental rule to the `ktlint_official` code style, but it can be enabled explicitly for the other code styles as well. ([#1980](https://github.com/pinterest/ktlint/issues/1980))
 * Clean-up unwanted logging dependencies ([#1998](https://github.com/pinterest/ktlint/issues/1998))
 * Fix directory traversal for patterns referring to paths outside of current working directory or any of it child directories ([#2002](https://github.com/pinterest/ktlint/issues/2002))
+* Prevent multiple expression in same line separated by semicolon ([#1078](https://github.com/pinterest/ktlint/issues/1078))
 
 ### Changed
 
--- a/ktlint-cli/src/main/kotlin/com/pinterest/ktlint/cli/internal/ReporterAggregator.kt
+++ b/ktlint-cli/src/main/kotlin/com/pinterest/ktlint/cli/internal/ReporterAggregator.kt
@@ -140,7 +140,8 @@ internal class ReporterAggregator(
         val stream =
             when {
                 reporterConfiguration.output != null -> {
-                    File(reporterConfiguration.output).parentFile?.mkdirsOrFail(); PrintStream(reporterConfiguration.output, "UTF-8")
+                    File(reporterConfiguration.output).parentFile?.mkdirsOrFail()
+                    PrintStream(reporterConfiguration.output, "UTF-8")
                 }
 
                 stdin -> {
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
@@ -11,6 +11,7 @@ import com.pinterest.ktlint.rule.engine.core.api.ElementType.COMMA
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.CONDITION
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.DESTRUCTURING_DECLARATION
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.DOT
+import com.pinterest.ktlint.rule.engine.core.api.ElementType.ENUM_ENTRY
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.EOL_COMMENT
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.FUN
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.FUNCTION_LITERAL
@@ -539,6 +540,11 @@ public class WrappingRule :
     ) {
         val previousCodeLeaf = node.prevCodeLeaf()?.lastChildLeafOrSelf() ?: return
         val nextCodeLeaf = node.nextCodeLeaf()?.firstChildLeafOrSelf() ?: return
+        if (previousCodeLeaf.treeParent.elementType == ENUM_ENTRY && nextCodeLeaf.elementType == RBRACE) {
+            // Allow
+            // enum class INDEX2 { ONE, TWO, THREE; }
+            return
+        }
         if (noNewLineInClosedRange(previousCodeLeaf, nextCodeLeaf)) {
             requireNewlineAfterLeaf(node, autoCorrect, emit, indent = previousCodeLeaf.indent())
         }
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
@@ -26,6 +26,7 @@ import com.pinterest.ktlint.rule.engine.core.api.ElementType.OBJECT_LITERAL
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.RBRACE
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.RBRACKET
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.RPAR
+import com.pinterest.ktlint.rule.engine.core.api.ElementType.SEMICOLON
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.STRING_TEMPLATE
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.SUPER_TYPE_CALL_ENTRY
 import com.pinterest.ktlint.rule.engine.core.api.ElementType.SUPER_TYPE_ENTRY
@@ -64,6 +65,7 @@ import com.pinterest.ktlint.rule.engine.core.api.nextCodeLeaf
 import com.pinterest.ktlint.rule.engine.core.api.nextCodeSibling
 import com.pinterest.ktlint.rule.engine.core.api.nextLeaf
 import com.pinterest.ktlint.rule.engine.core.api.nextSibling
+import com.pinterest.ktlint.rule.engine.core.api.noNewLineInClosedRange
 import com.pinterest.ktlint.rule.engine.core.api.prevCodeLeaf
 import com.pinterest.ktlint.rule.engine.core.api.prevLeaf
 import com.pinterest.ktlint.rule.engine.core.api.prevSibling
@@ -137,6 +139,7 @@ public class WrappingRule :
             ARROW -> rearrangeArrow(node, autoCorrect, emit)
             WHITE_SPACE -> line += node.text.count { it == '\n' }
             CLOSING_QUOTE -> rearrangeClosingQuote(node, autoCorrect, emit)
+            SEMICOLON -> insertNewLineBeforeSemi(node, autoCorrect, emit)
         }
     }
 
@@ -526,6 +529,18 @@ public class WrappingRule :
         val r = node.nextSibling { it.elementType == RBRACE } ?: return
         if (!r.prevLeaf().isWhiteSpaceWithNewline()) {
             requireNewlineBeforeLeaf(r, autoCorrect, emit, node.indent())
+        }
+    }
+
+    private fun insertNewLineBeforeSemi(
+        node: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit,
+    ) {
+        val previousCodeLeaf = node.prevCodeLeaf()?.lastChildLeafOrSelf() ?: return
+        val nextCodeLeaf = node.nextCodeLeaf()?.firstChildLeafOrSelf() ?: return
+        if (noNewLineInClosedRange(previousCodeLeaf, nextCodeLeaf)) {
+            requireNewlineAfterLeaf(node, autoCorrect, emit, indent = previousCodeLeaf.indent())
         }
     }
 
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
@@ -140,7 +140,7 @@ public class WrappingRule :
             ARROW -> rearrangeArrow(node, autoCorrect, emit)
             WHITE_SPACE -> line += node.text.count { it == '\n' }
             CLOSING_QUOTE -> rearrangeClosingQuote(node, autoCorrect, emit)
-            SEMICOLON -> insertNewLineBeforeSemi(node, autoCorrect, emit)
+            SEMICOLON -> insertNewLineAfterSemi(node, autoCorrect, emit)
         }
     }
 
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ At this point in time, it is not yet decided what the next steps will be. Ktlint
 * Extract rule `no-single-line-block-comment` from `comment-wrapping` rule. The `no-single-line-block-comment` rule is added as experimental rule to the `ktlint_official` code style, but it can be enabled explicitly for the other code styles as well. ([#1980](https://github.com/pinterest/ktlint/issues/1980))
 * Clean-up unwanted logging dependencies ([#1998](https://github.com/pinterest/ktlint/issues/1998))
 * Fix directory traversal for patterns referring to paths outside of current working directory or any of it child directories ([#2002](https://github.com/pinterest/ktlint/issues/2002))
+* Prevent multiple expression in same line separated by semicolon ([#1078](https://github.com/pinterest/ktlint/issues/1078))
```
paul-dingemans, line 70: Shouldn't `expression in same line` be `expressions on same line`?

ktlint-cli/src/main/kotlin/com/pinterest/ktlint/cli/internal/ReporterAggregator.kt
```diff
@@ -140,7 +140,8 @@ internal class ReporterAggregator(
         val stream =
             when {
                 reporterConfiguration.output != null -> {
-                    File(reporterConfiguration.output).parentFile?.mkdirsOrFail(); PrintStream(reporterConfiguration.output, "UTF-8")
+                    File(reporterConfiguration.output).parentFile?.mkdirsOrFail()
+                    PrintStream(reporterConfiguration.output, "UTF-8")
```
paul-dingemans, line 144: So cool, when a change in rules finds a violation in ktlint itself ;-)

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
```diff
@@ -529,6 +533,23 @@ public class WrappingRule :
...
+    private fun insertNewLineBeforeSemi(
+        node: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit,
+    ) {
+        val previousCodeLeaf = node.prevCodeLeaf()?.lastChildLeafOrSelf() ?: return
+        val nextCodeLeaf = node.nextCodeLeaf()?.firstChildLeafOrSelf() ?: return
+        if (previousCodeLeaf.treeParent.elementType == ENUM_ENTRY && nextCodeLeaf.elementType == RBRACE) {
+            // Allow
+            // enum class INDEX2 { ONE, TWO, THREE; }
```
paul-dingemans, line 545: I see no reason to allow this exception. Or is it added because wrapping is now run before the no-semi rule which removes this redundant semicolon? If so, then document that in a comment.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
```diff
@@ -529,6 +533,23 @@ public class WrappingRule :
...
+            return
+        }
+        if (noNewLineInClosedRange(previousCodeLeaf, nextCodeLeaf)) {
+            requireNewlineAfterLeaf(node, autoCorrect, emit, indent = previousCodeLeaf.indent())
```
paul-dingemans, line 549: The method is called `insertNewLineBeforeSemi` but actually the newline is added after the semicolon. The latter is preferred, because it will produce the best result when the `no-semi` rule is disabled.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
```diff
@@ -1871,6 +1871,303 @@ internal class WrappingRuleTest {
             .hasLintViolation(2, 10, "A newline was expected before 'Bar'")
             .isFormattedAs(formattedCode)
     }
+
+    @Test
+    fun `Issue 1078 - Given a multiline semi separated var initialisation then wrap each expression to a new line`() {
+        val code =
+            """
+            fun f1() {
```
paul-dingemans, line 1879: Please try to stick with `foo`, `bar`, `foobar` (same for `foo1` etc). Numbered identifiers like `foo1` should be used only when we have multiple of them.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
```diff
@@ -1871,6 +1871,303 @@ internal class WrappingRuleTest {
...
+                LintViolation(2, 15, "Missing newline after \";\""),
+                LintViolation(3, 26, "Missing newline after \";\""),
+            )
+            .isFormattedAs(formattedCode)
+    }
+
+    @Test
+    fun `Issue 1078 - Given a multiline semi separated statements then wrap each expression to a new line`() {
+        val code =
+            """
```
paul-dingemans, line 1904: Please split in multiple tests. If needed use a nested test class to group related tests.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
```diff
@@ -1871,6 +1871,303 @@ internal class WrappingRuleTest {
...
+                LintViolation(21, 14, "Missing newline after \";\""),
+                LintViolation(26, 13, "Missing newline after \";\""),
+            )
+            .isFormattedAs(formattedCode)
+    }
+
+    @Test
+    fun `Issue 1078 - Given a multiline semi separated control flow then wrap each expression to a new line`() {
+        val code =
+            """
```
paul-dingemans, line 1984: Please split in multiple tests. If needed use a nested test class to group related tests.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
```diff
@@ -1871,6 +1871,303 @@ internal class WrappingRuleTest {
...
+                val key: String = "";
+                val lambda: () -> Unit = {  }
+            }
+            """.trimIndent()
+        wrappingRuleAssertThat(code)
+            .hasLintViolations(
+                LintViolation(2, 15, "Missing newline after \";\""),
+                LintViolation(3, 26, "Missing newline after \";\""),
+            )
+            .isFormattedAs(formattedCode)
```
paul-dingemans, line 1898: Keep lines like this together on one line.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
```diff
@@ -1871,6 +1871,303 @@ internal class WrappingRuleTest {
...
+        val formattedCode =
+            """
+            fun f1() {
+                val a = 3;
+                val b = 2
+                val key: String = "";
+                val lambda: () -> Unit = {  }
+            }
+            """.trimIndent()
+        wrappingRuleAssertThat(code)
```
paul-dingemans, line 1893: In tests like this also add `.addAdditionalRuleProvider { NoSemicolonsRule() }` so that the formatted result also has cleaned up the unwanted semicolon.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
```diff
@@ -1871,6 +1871,303 @@ internal class WrappingRuleTest {
...
+    fun `Issue 1078 - Given a multiline semi separated with multiple items`() {
+        val code =
+            """
+            fun test() {
+                val a = 0; val b = 0; fun bar() {
+                    // no-op
+                }; for(i in 0..10) {
+                    println(i); println(i); a++; println(a)
+                }
+            }
```
paul-dingemans, line 2091: What is tested in this method that was not tested in the other tests?

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
```diff
@@ -1871,6 +1871,303 @@ internal class WrappingRuleTest {
...
+            """
+            class FirstClass(private val param: String) {
+                companion object;
+                private var toto: Boolean = false
+            }
+
+            class SecondClass(private val param: String) {
+                companion object; // with comment
+                private var toto: Boolean = false
+            }
```
paul-dingemans, line 2166: What is tested in this method that was not tested in the other tests?

paul-dingemans (changes requested):

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: But as per the doc of `addAdditionalRuleProvider`, it runs before the rule which is getting unit tested `An additional rule provider for a rule which actually is executed before the rule under test,`

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: More than one statement. But anyway I will look again

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: `companion object; // with comment` here `companion object;` is allowed I just added additional `// with comment` comment after the semi

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
atulgpt: Yes as when both rules are run final code will be 
```
enum class INDEX2 { ONE, TWO, THREE
}
```

If `no-semi` rule is not run than the final code will be like below
```
enum class INDEX2 { ONE, TWO, THREE;
}
```

In both the cases, final code was not looking good

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRule.kt
atulgpt: By `latter` if you mean new line after semi then yes that is also my understanding. I will update the name to `insertNewLineAfterSemi`

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
paul-dingemans: It does not always lead to problems. You can always try to add it anyways and document it like below:
```
                .hasLintViolations(
                    // Some violations below are only reported during linting but not on formatting. After wrapping the
                    // nullable type, there is no need to wrap the parameters.
                    LintViolation(2, 12, "Expected new line before function type as it does not fit on a single line"),
                    LintViolation(2, 13, "Parameter should start on a newline"),
                    LintViolation(2, 24, "Parameter should start on a newline"),
                    LintViolation(2, 35, "Parameter should start on a newline"),
                    LintViolation(2, 44, "Missing newline before \")\""),
                    LintViolation(2, 53, "Expected new line after function type as it does not fit on a single line"),
                ).isFormattedAs(formattedCode)
```

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: Updated. BTW @paul-dingemans do we also have some rules to enforce this formatting? Or is it related to https://github.com/pinterest/ktlint/issues/1953 Or should we create a separate issue for that?

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: Split the TCs into smaller ones grouped via `Nested` classes

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: I have added `addAdditionalRuleProvider` and it really cleaned semicolons 😅 It seems even after `An additional rule provider for a rule which actually is executed before the rule under test,` due to constraints `NoSemicolonsRule` make sure that it is run after the `WrappingRuleTest`

atulgpt: Also, I have not added `addAdditionalRuleProvider` as I also wanted to test this rule w/o `NoSemicolonsRule`

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: I have added a new TC with comments in the code

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
paul-dingemans: The idea was to enforce this with #1953.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
```diff
@@ -1871,6 +1871,628 @@ internal class WrappingRuleTest {
...
+                        THREE,
+                        ;
+                    }
+                    enum class INDEX7 {
+                        ONE,
+                        TWO,
+                        THREE,
+                        ;
+                        fun foo() = ""
+                    }
```
paul-dingemans, line 2417: I believe that those example are not relevant. A semicolon followed by a newline is always ignored by the wrapping rule and not specically in enums only. The `INDEX2` (which btw should called `Foo`) the real exception as wrapping the closing brace would result in ugly code after removing the semi-colon.

paul-dingemans (changes requested):
Tnx for addessing the review comments. The tests are way more readable now. I adressed one more issue.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
paul-dingemans: Downside is that it is now sometime confusing why the `NoSemicolonsRule` is not present in some test while it is added in other tests. As the default behavior of ktlint is to run both rules, I prefer to add the rule because the formattedCode in that case looks closest to the normal behavior of ktlint.
But when there is a specific need in a test to also validate the end result in case the rule is disabled, then I would suggest to make it explicit in the name of the test or with a comment why the the  `NoSemicolonsRule` is not added.

atulgpt: Yes this makes more sense. I have added one TC `Given two variables run without NoSemicolonsRule`

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: I have updated the TC name to be more clear here. TC name is `Given a multiline semi separated with variables, flow controls and method calls`

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/WrappingRuleTest.kt
atulgpt: These simply asserting that there is no false +ve here due to semi. `followed by a newline is always ignored by the wrapping rule` this is true only when we know the internal implementation of the `WrappingRule` but in future this assumption can change. Right? Shouldn't we have similar cases in TCs to prevent any break as well?

atulgpt: I have updated the name to use `FOO` in enums as well. Also let me know if you don't agree with my above statement(in that case I will remove those TCs as well) and think changing the name or creating TC some-other way can make it more clear

paul-dingemans:
Tnx for the contributation @atulgpt !

PR #2076, opened 2023-06-12
Remove misleading sample code

avinash3699:
Foo_Bar throws an error according to Rule id: enum-entry-name-case, but it is given as the example for allowed code.


paul-dingemans:
Tnx for again a documentation fix. It is great that you careful read the documentation an take the effort to fix the errors.

avinash3699:
Thank you. Will try to contribute more towards the coding part soon 😉

PR #2088, opened 2023-06-24
Fixes/1953/chained method continuation

atulgpt:
## Description

Add ChainMethodContinuation(chain-method-continuation) rule
    
This rule ensures chain operators(`.` or `?.`) aligns with next call
When there is `}` in the new line this also ensures that the chain operators align with `}` along with the next call method

This partially fixes https://github.com/pinterest/ktlint/issues/1953

## Checklist

- [x] PR description added
- [x] tests are added

In case of adding a new rule:
- [x] Rule is added to [rules documentation](https://pinterest.github.io/ktlint/rules/standard/)


ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,113 @@
...
+        chainOperator ?: return
+        val rightExpression = chainOperator.nextCodeSibling()?.firstChildLeafOrSelf() ?: return
+        val leftExpressionEndBrace =
+            chainOperator.prevLeaf { !it.isWhiteSpace() && it.elementType != ElementType.EXCLEXCL }
+                ?.lastChildLeafOrSelf() ?: return
+        val leftExpressionBeforeEndBrace = leftExpressionEndBrace.prevCodeSibling()?.lastChildLeafOrSelf() ?: return
+        val isLeftExpressionEndBraceInSeparateLine =
+            leftExpressionEndBrace.elementType == ElementType.RBRACE &&
+                hasNewLineInClosedRange(leftExpressionBeforeEndBrace, leftExpressionEndBrace)
+        if (!noNewLineInOpenRange(chainOperator, rightExpression)) {
```
paul-dingemans, line 60: We can define a general function like `hasNewLineInOpenRange`. But in this case, the code below also works because we already know that the boundary node are not a newline:
```
        if (hasNewLineInClosedRange(chainOperator, rightExpression)) {
```

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,113 @@
...
+            if (autoCorrect) {
+                if (isLeftExpressionEndBraceInSeparateLine) {
+                    /*
+                        Detects code like below
+                        bar {
+                            ...
+                        }.
+                        foo() // this should align with previous line }
+                     */
+                    chainOperator.upsertWhitespaceAfterMe("")
```
paul-dingemans, line 83: This leaves an empty PsiWhitespace elements in the AST. Technically that is not a problem. But I think that removing an unwanted whitespace element is more clear.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,113 @@
...
+                } else {
+                    chainOperator.upsertWhitespaceBeforeMe(rightExpression.indent())
+                    rightExpression.upsertWhitespaceBeforeMe("")
```
paul-dingemans, line 86: This leaves an empty PsiWhitespace elements in the AST. Technically that is not a problem. But I think that removing an unwanted whitespace element is more clear.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,113 @@
...
+                }
+                .foo() // this should align with previous line }
+             */
+            emit(
+                chainOperator.textRange.endOffset,
+                "${chainOperator.text} must must merge at the end of previous call",
+                true,
+            )
+            if (autoCorrect) {
+                chainOperator.upsertWhitespaceBeforeMe("")
```
paul-dingemans, line 109: This leaves an empty PsiWhitespace elements in the AST. Technically that is not a problem. But I think that removing an unwanted whitespace element is more clear.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
```diff
@@ -0,0 +1,172 @@
...
+    fun `Given method chain which is correctly wrapped in single line`() {
+        val code =
+            """
+            val foo = listOf(1, 2, 3).filter { it > 2 }!!.takeIf { it.count() > 100 }.map { it * it }?.sum()!!
+            """.trimIndent()
+        chainMethodContinuationAssertThat(code).hasNoLintViolations()
+    }
+
+    @Test
+    fun `Given method chain which is incorrectly wrapped`() {
```
paul-dingemans, line 20: I prefer to have smaller tests when possible. I think that this test should be split into 3 (one test for operators `.`, `?.` and `!!.`.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                .takeIf { it.count() > 100 }
+                .map {
+                    it * it
+                }?.map {
+                    it * it
+                }!!.sum()
+            """.trimIndent()
+        chainMethodContinuationAssertThat(code)
+            .hasLintViolations(
+                LintViolation(1, 27, ". must merge at the start of next call"),
```
paul-dingemans, line 47: What do you think about changing the message to `Unexpected newline after '.'`?

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                LintViolation(2, 25, ". must merge at the start of next call"),
+                LintViolation(3, 33, ". must merge at the start of next call"),
+                LintViolation(6, 8, "}?. must merge at the start of next call"),
+                LintViolation(9, 9, "}!!. must merge at the start of next call"),
+            ).isFormattedAs(formattedCode)
+    }
+
+    @Test
+    fun `Given method with RBRACE is not wrapped with next chain operator`() {
```
paul-dingemans, line 56: Just out of curious. Are you looking forward from the RBRACE to see whether it is followed by a method chain or are you looking backward if the DOT or SAFE_ACCESS is preceded by a dangling RBRACE? Usually it is better to look back instead of forward as all nodes that have been processed before are already cleaned up if needed, nodes after the current node are not yet cleaned.

And please split into smaller tests.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
```diff
@@ -0,0 +1,172 @@
...
+            """
+            val foo = object : Runnable {
+                override fun run() {
+                    /* no-op */
+                }
+            }.
+                run()!!.
+                hashCode()?.
+                hashCode()?.
+                hashCode()
```
paul-dingemans, line 116: Why is this relevant? This is also tested in the other tests.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
```diff
@@ -0,0 +1,172 @@
...
+            val foo2 = listOf(1, 2, 3).filter { it > 2 }!!
+                .sum()
+            val foo3 = listOf(1, 2, 3).filter { it > 2 }!!
+                ?.sum()
+            """.trimIndent()
+        chainMethodContinuationAssertThat(code).hasNoLintViolations()
+    }
+
+    @Test
+    fun `Given object declaration in chain which is incorrectly wrapped`() {
```
paul-dingemans, line 105: The rule seems to have no special handling for object declaration. So what is the relevance of this test?

paul-dingemans (changes requested):

paul-dingemans:
Formatting code below:
```
val bar =
    listOf(1, 2, 3). // xxx
        map {
            it * it
        }
```
results in code which can not be compiled:
```
val bar =
    listOf(1, 2, 3)
        .// xxxmap {
            it * it
        }
```

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,113 @@
...
+    override fun beforeVisitChildNodes(
+        node: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit,
+    ) {
+        node.takeIf {
+            it.elementType in chainOperators
+        }?.let {
+            visitChains(node, autoCorrect, emit)
+        }
```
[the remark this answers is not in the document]
atulgpt: Nice find :)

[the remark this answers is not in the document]
atulgpt: Although I have kept the name `chainOperator` as this node will always be a chain operator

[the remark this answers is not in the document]
atulgpt: @paul-dingemans don't we have any rule for `For me this reads a lot easier as things that belong together are on a single line.` to make the lambda in a single line when it doesn't violate the max line requirements

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private fun visitChains(
+        node: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit,
+    ) {
```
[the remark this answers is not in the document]
atulgpt: `you can just call chainOperator.isPrecededByMultilineChain() which is more expressive than using a variable` Indeed calling the extension method will be more expensive as we have to calculate it 3 times. Are you still suggesting it to move it to extension method?

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,113 @@
...
+            emit(
+                chainOperator.textRange.endOffset,
+                "${
+                    if (isLeftExpressionEndBraceInSeparateLine) {
+                        leftExpressionEndBrace.leavesIncludingSelf()
+                            .takeWhile { it != chainOperator && !it.isWhiteSpace() }.map { it.text }
+                            .joinToString(separator = "")
+                    } else {
+                        ""
+                    }
```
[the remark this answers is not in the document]
atulgpt: It takes `chainOperator`, `leftExpressionEndBrace` and `isLeftExpressionEndBraceInSeparateLine` variables. Will something like below work?
```
"${
    getLeftAlighingTextAfterAligningBrace(
        isLeftExpressionEndBraceInSeparateLine,
        leftExpressionEndBrace,
        chainOperator
    )
}${chainOperator.text} must merge at the start of next call"
```

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
atulgpt: Coincidentally I am looking backwards. But I was not aware of this. My code to find the brace is
```
val leftExpressionEndBrace =
    chainOperator
        .prevLeaf { !it.isWhiteSpace() && it.elementType != EXCLEXCL }
        ?.lastChildLeafOrSelf() ?: return
val leftExpressionBeforeEndBrace = leftExpressionEndBrace.prevCodeSibling()?.lastChildLeafOrSelf() ?: return
val isLeftExpressionEndBraceInSeparateLine =
    leftExpressionEndBrace.elementType == RBRACE &&
        hasNewLineInClosedRange(leftExpressionBeforeEndBrace, leftExpressionEndBrace)
```

paul-dingemans: Given that you're lookingbackward, you can also reflect this in the name of the. Given a chain operator on a separate line below the rbrace of a multiline chain.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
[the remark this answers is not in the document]
paul-dingemans: Actually, it will not be called more than twice because it is used in mutually exclusive branches of the if-else. I value code readability over premature code optimazations. If would be different if the function would be called inside a loop which is executed tens or hundreds of times.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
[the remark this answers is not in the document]
paul-dingemans: It is slightly better, but I am still not happy with for two reasons:
- I want string template to be short (one or max two chained methods) and a single line. If that is not possible, just extract it from the string.
- The method name `getLeftAlighingTextAfterAligningBrace` describes what it DOES but not what it IS. So, functionality wise I have to reverse engineer this piece of code to understand what it does. What "is" the text that you are looking for?

I think that the look also will become more clear if your refactor line 61-88 to smth like:
```
if (isLeftExpressionEndBraceInSeparateLine) {
    emit(...)
    if (autoCorrect) {
        ....
    }
} else {
    emit(...)
    if (autoCorrect) {
        ....
    }
}
```
at it keeps the emitted message and the auto correct together.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
atulgpt: What about `Expected newline before '.'` as someone can be confused the multiline chain is reported by the ktlint. Removing the newline after `.` will make the below code 
```
val foo = listOf(1, 2, 3).
                filter { it > 2 }!!.
```
to 
```
val foo = listOf(1, 2, 3).filter { it > 2 }!!.
```

atulgpt: Although for cases like below
```
val foo = listOf(1, 2, 3).filter {
    true
}
.sum()
```
I am thinking of `Unexpected newline before '}.'` as here your suggestion seems more right to me

paul-dingemans: Yeah, the message is depending on the situation. So, I am fine with both suggestions.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
atulgpt: Updated with `nextMultilineWhiteSpace.treeParent.removeChild(nextMultilineWhiteSpace)`

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
atulgpt: Updated with `nextMultilineWhiteSpace.treeParent.removeChild(nextMultilineWhiteSpace)`

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
atulgpt: Updated with `nextMultilineWhiteSpace.treeParent.removeChild(nextMultilineWhiteSpace)`

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
atulgpt: Made this change. Let me know after reviewing the latest changes more splitting is required for code readability

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
atulgpt: I have updated the verbiage a little bit also making it more consistent

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
atulgpt: removed this TC

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
atulgpt: Actually, this is added so that I can add method chains as well as object chains.

atulgpt:
> Formatting code below:
> 
> ```
> val bar =
>     listOf(1, 2, 3). // xxx
>         map {
>             it * it
>         }
> ```
> 
> results in code which can not be compiled:
> 
> ```
> val bar =
>     listOf(1, 2, 3)
>         .// xxxmap {
>             it * it
>         }
> ```

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,195 @@
...
+        chainOperator: ASTNode,
+        autoCorrect: Boolean,
+        emit: (offset: Int, errorMessage: String, canBeAutoCorrected: Boolean) -> Unit,
+    ) {
+        val rightExpression = chainOperator.nextCodeSibling()?.firstChildLeafOrSelf() ?: return
+        val previousEndBrace =
+            chainOperator.getPrevChainEndBrace() ?: return
+        val previousExpressionBeforeBrace = previousEndBrace.prevCodeSibling()?.lastChildLeafOrSelf() ?: return
+        val isPreviousChainElementMultiline =
+            previousEndBrace.elementType == RBRACE &&
```
atulgpt, line 54: hi @paul-dingemans shuld we also check `RBRACKET` and `RPAR` here?

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
```diff
@@ -0,0 +1,195 @@
...
+                foo() // unexpected new line after .
+
+                or
+                bar(). // expected newline before .
+                foo()
+             */
+            fixMisalignedChainOperator(
+                emit,
+                chainOperator,
+                isPreviousChainElementMultiline,
```
atulgpt, line 71: I have passed the required previously calculated value in the method let me know if you want this to be recalculated from `chainOperator` itself from some refactored extension method. To me this also reads fine

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
```diff
@@ -0,0 +1,452 @@
...
+        fun `Given method chain which is incorrectly wrapped in next line with previous RBrace with single line comments`() {
+            val code =
+                """
+                val foo = listOf(1, 2, 3).filter {
+                    it > 2
+                }// 1
+                    .sum()// 2
+                """.trimIndent()
+            chainMethodContinuationAssertThat(code)
+                .hasLintViolationsWithoutAutoCorrect(
```
atulgpt, line 432: @paul-dingemans also is this the correct way to assert that this violations are present but linter didn't do any autocorrect

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
paul-dingemans: Might be, but a code sample that you would like to detect would be helpful.

But you can also look at it another way. I think that you want to detect that there is no newline between end of the last code leaf before the chain operator and the start of the first code leaf after the chain operator. So then it doesn't matter whether it is RBRACE, RBRACKET or RPAR.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
paul-dingemans: yes, it is acceptable.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
paul-dingemans: Yes, but when there is only one lint violation then use `hasLintViolationWithoutAutoCorrect` for better readability.

ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuation.kt
atulgpt: `I think that you want to detect that there is no newline between the end of the last code leaf before the chain operator and the start of the first code leaf after the chain operator.` yes for cases when RBRACE is in multiline. I got this idea when in a few instances in my PR you pointed out that the below code is not indented
```
listOf(1, 2, 3)
    .map {
        it*2
    }
    .sum()
```
and you suggested below alternative
```
listOf(1, 2, 3)
    .map {
        it*2
    }.sum()
```

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationTest.kt
atulgpt: I like using same method as that make me able to change the TC or copy the TC without changing the assertion based on the number of violations

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationRuleTest.kt
```diff
@@ -0,0 +1,811 @@
...
+                }
+            }.run()
+            """.trimIndent()
+        chainMethodContinuationRuleAssertThat(code)
+            .hasLintViolation(5, 2, "Unexpected newline after '.'")
+            .isFormattedAs(formattedCode)
+    }
+
+    @Test
+    fun `Given nested chains which are correctly wrapped`() {
```
atulgpt, line 339: should we also add TC with index operator like below

```
list(1)[
    0
]
.inc()

list(1)[
    0
].
inc()

list(1)[
    0
]!!
.inc()
```

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationRuleTest.kt
```diff
@@ -0,0 +1,811 @@
...
+                .hasLintViolations(
+                    // Lint violation below will not be triggered during format as argument-list-wrapping rule prevents this error from occuring
+                    LintViolation(2, 30, "Expected newline before '.'"),
+                    LintViolation(3, 30, "Expected newline before '?.'"),
+                ).isFormattedAs(formattedCode)
+        }
+    }
+
+    @Test
+    fun `Given a single line chained expression, not exceeding max line length, with too many chain operators`() {
```
atulgpt, line 588: we can also add TC for binary operators on two long chains exceeding the max line lenght

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationRuleTest.kt
```diff
@@ -0,0 +1,811 @@
...
+
+    @Test
+    fun `Given nested chains which are correctly wrapped`() {
+        val code =
+            """
+            val foo = listOf(1, 2, 3)
+                .filter {
+                    listOf(1, 2, 3).map { it * it }.size > 1
+                }!!
+                .map {
```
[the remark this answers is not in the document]
paul-dingemans: No, it disaligns `filter` and `map` too much. At least, I found it very distracting.

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationRuleTest.kt
```diff
@@ -0,0 +1,811 @@
...
+
+    @Test
+    fun `Given a single line chained expression, not exceeding max line length, with too many chain operators`() {
+        val code =
+            """
+            val foo = listOf(1, 2, 3).foo1().foo2().foo3().foo4().foo5().foo6()
+            """.trimIndent()
+        val formattedCode =
+            """
+            val foo = listOf(1, 2, 3)
```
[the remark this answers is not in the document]
paul-dingemans: This is already catered for. Above is formatted as:
```
val foo =
    listOf(1, 2, 3)
        .foo1(
            listOf(1, 2, 3)
                .foo1()
                .foo2()
                .foo3()
                .foo4()
                .foo5()
                .foo6(),
        ).foo2()
        .foo3()
        .foo4()
        .foo5()
        .foo6()
```

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationRuleTest.kt
paul-dingemans: Hmm, this is not handled well (yet). It is formatted as:
```
    list(1)[
        0,
    ]
        .inc()

    list(1)[
        0,
    ]
        ?.inc() // Assumption is that here the `?.` chain operator was intended

    list(1)[
        0,
    ]!!
        .inc()
```

So a TC and fix is needed. I would say that expected outcome should be:
```
    list(1)[
        0,
    ].inc()

    list(1)[
        0,
    ]?.inc() // Assumption is that here the `?.` chain operator was intended

    list(1)[
        0,
    ]!!
        .inc()
```

ktlint-ruleset-standard/src/test/kotlin/com/pinterest/ktlint/ruleset/standard/rules/ChainMethodContinuationRuleTest.kt
paul-dingemans: Those are ignore as explained in the rule:
```
    private fun ChainedExpression.exceedsMaxLineLength() =
        with(rootASTNode) {
            if (treeParent.elementType == BINARY_EXPRESSION) {
                // Chained expressions which are enclosed inside a binary expression are skipped for now. It depends on the situation
                // whether wrapping on the binary expression takes precedence on the chained expression or vice versa.
                // This can be illustrated with following examples:
                //    - wrapping on chained expression first makes more sense in this case
                //        val foo = listOf("one", "two", "three").filter { it.length > 3 }
                //      resulting in:
                //        val foo = listOf("one", "two", "three")
                //            .filter { it.length > 3 }
                //    - wrapping on binary expression first makes more sense in this case
                //        if (someOtherExpression || listOf("one", "two", "three").any { it.length > 3 }) { ... }
                //      resulting in:
                //        if (someOtherExpression ||
                //            listOf("one", "two", "three").any { it.length > 3 }
                //        ) { ... }
                false
            } else {
```

Issue #2061, opened 2023-05-27
Upgrade to Kotlin 1.9 && remove extension point


paul-dingemans:
With #2044 the extension point `org.jetbrains.kotlin.com.intellij.treeCopyHandler` can optionally be disabled in ktlint `0.50.x`. When upgrading to Kotlin 1.9 this extension point has to be removed unless the extension point is being supported.

PR #1889, opened 2023-03-25
Replace testing on java 19 with testing on java 20

paul-dingemans:
## Description

Replace testing on java 19 with testing on java 20

Closes #1888

## Checklist

- [X] PR description added
- [ ] tests are added
- [ ] KtLint has been applied on source code itself and violations are fixed
- [ ] [documentation](https://pinterest.github.io/ktlint/) is updated
- [X] `CHANGELOG.md` is updated

In case of adding a new rule:
- [ ] Rule is added to [rules documentation](https://pinterest.github.io/ktlint/rules/standard/)

PR #1787, opened 2023-01-23
Run tests parallelly on CI

Goooler:
## Description

This is a regression of using toolchain, reverts some changes in #1294 & #1453 to save more time on CI, especially for the Windows related builds. More result diffs you can see https://github.com/Goooler/ktlint/pull/39.

## Checklist

- [x] PR description added
- [ ] tests are added
- [ ] KtLint has been applied on source code itself and violations are fixed
- [ ] [documentation](https://pinterest.github.io/ktlint/) is updated
- [ ] `CHANGELOG.md` is updated

In case of adding a new rule:
- [ ] Rule is added to [rules documentation](https://pinterest.github.io/ktlint/rules/standard/)

PR #2120, opened 2023-07-09
Setup toolchains, compile project with Java 20 only, run test on various Java versions

### LINKED ISSUE #1888: Add testing on Java 20 and remove testing for Java 19
Java 19 has reached end of life per March 21st, 2023. Java 20 has been release on that date. As of that, testing on Java 19 has to be replaced with testing on Java 20.

mateuszkwiecinski:
Closes #1888

## Description
I wanted to open the project, but it failed to sync for me locally (because my defaults point to Java 20), so I decided to make the build more deterministic. I'll comment out all changes in the PR since some of them require your input, but the tl;dr is:
- The project will be built only with a single Java version (that's what the release workflow does)
- Only tests will be invoked on various Java versions (that's what project consumers do)
- The project will compile on Java 20, and if someone doesn't have Java 11 or 20 installed locally, Gradle will fetch missing ones
- I also removed a few of annoying compiler and deprecation warnings 

## Checklist

- [x] At least one commit message contains a reference `Closes #<xxx>` or `Fixes #<xxx>` (replace`<xxx>` with issue number)
- [ ] Tests are added
- [X] PR description added

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -37,3 +37,6 @@ max_line_length = unset
 
 [gradle/verification-metadata.xml]
 indent_size = 3
+
+[*.yml]
+ij_yaml_spaces_within_brackets = false
--- a/.github/actions/setup-gradle-build/action.yml
+++ b/.github/actions/setup-gradle-build/action.yml
@@ -1,0 +1,32 @@
+name: Setup Gradle
+description: Sets up the environment to run Gradle
+
+inputs:
+  gradle-jvm-args:
+    description: "JVM args to pass to Gradle"
+    required: true
+    default: "-Xmx1g"
+  additional-java-versions:
+    description: "Java versions installed on the side of the default Java version required by the build"
+    required: false
+
+runs:
+  using: composite
+
+  steps:
+    - uses: actions/setup-java@v3
+      with:
+        distribution: 'zulu'
+        java-version: | # last version (set as default) should match all `jvmToolchain(xxx)` calls in the project
+          ${{ inputs.additional-java-versions }}
+          20
+
+    - name: Write Gradle build properties to local `gradle.properties`
+      run: |
+        mkdir -p ~/.gradle
+        printf "org.gradle.jvmargs=${{ inputs.gradle-jvm-args }}\n" >> ~/.gradle/gradle.properties
+      shell: bash
+
+    - uses: gradle/gradle-build-action@v2
+      with:
+        gradle-home-cache-cleanup: true
--- a/.github/actions/setup-gradle-build/action.yml
+++ b/.github/actions/setup-gradle-build/action.yml
@@ -5,6 +5,9 @@ inputs:
   gradle-jvm-args:
     description: "JVM args to pass to Gradle"
     required: true
+    # Github-Hosted nodes have only 7GB of RAM available. Looking at build scans Gradle process requires slightly more than 0.5GB.
+    # Keeping this setting low, allows other, forked JVM processes (like tests) to use remaining memory.
+    # Increase this value, only if GC time visible in build scans will take more than a few seconds.
     default: "-Xmx1g"
   additional-java-versions:
     description: "Java versions installed on the side of the default Java version required by the build"
@@ -21,7 +24,11 @@ runs:
           ${{ inputs.additional-java-versions }}
           20
 
-    - name: Write Gradle build properties to local `gradle.properties`
+      # Please note these settings will override the ones set via `gradle.properties` committed to the repository - https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties
+      # List of optimizations:
+      # - `jvm-args` fine-tuned to the CI runner & tasks being invoked
+      # - disabled File System Watching to improve Windows build times. CI runs don't modify source files, hence they don't need to pay extra cost to efficiently track changed files.
+    - name: Optimize Gradle build properties for CI
       run: |
         mkdir -p ~/.gradle
         printf "org.gradle.jvmargs=${{ inputs.gradle-jvm-args }}\n" >> ~/.gradle/gradle.properties
--- a/.github/actions/setup-gradle-build/action.yml
+++ b/.github/actions/setup-gradle-build/action.yml
@@ -25,6 +25,7 @@ runs:
       run: |
         mkdir -p ~/.gradle
         printf "org.gradle.jvmargs=${{ inputs.gradle-jvm-args }}\n" >> ~/.gradle/gradle.properties
+        printf "org.gradle.vfs.watch=false\n" >> ~/.gradle/gradle.properties
       shell: bash
 
     - uses: gradle/gradle-build-action@v2
--- a/.github/workflows/publish-release-build.yml
+++ b/.github/workflows/publish-release-build.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 19
+          java-version: 20
       - uses: gradle/gradle-build-action@v2
         with:
           gradle-home-cache-cleanup: true
--- a/.github/workflows/publish-release-build.yml
+++ b/.github/workflows/publish-release-build.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 20
+          java-version: 20 # This should match all `jvmToolchain(xxx)` calls in the project
       - uses: gradle/gradle-build-action@v2
         with:
           gradle-home-cache-cleanup: true
--- a/.github/workflows/publish-release-build.yml
+++ b/.github/workflows/publish-release-build.yml
@@ -1,8 +1,8 @@
 name: Publish release build
 
-on :
-  push :
-    tags :
+on:
+  push:
+    tags:
       - '*.*.*'
 
 jobs:
@@ -26,14 +26,14 @@ jobs:
         env:
           SONATYPE_NEXUS_USERNAME: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
           SONATYPE_NEXUS_PASSWORD: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
-          ORG_GRADLE_PROJECT_signingKey : ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEY }}
+          ORG_GRADLE_PROJECT_signingKey: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEY }}
           ORG_GRADLE_PROJECT_signingKeyId: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEYID }}
-          ORG_GRADLE_PROJECT_signingKeyPassword : ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
+          ORG_GRADLE_PROJECT_signingKeyPassword: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
 
-      - name : Extract release notes
-        id : release_notes
+      - name: Extract release notes
+        id: release_notes
         if: ${{ success() }}
-        uses : ffurrer2/extract-release-notes@v1
+        uses: ffurrer2/extract-release-notes@v1
 
       - name: Get version
         id: get_version
@@ -49,18 +49,18 @@ jobs:
           cp ktlint ktlint-${{ env.version }}/bin
           zip -rm ktlint-${{ env.version }}.zip ktlint-${{ env.version }}
 
-      - name : Create release
+      - name: Create release
         id: github_release
         if: ${{ success() }}
-        uses : softprops/action-gh-release@v1
-        with :
+        uses: softprops/action-gh-release@v1
+        with:
           draft: false
           prerelease: false
-          body : ${{ steps.release_notes.outputs.release_notes }}
+          body: ${{ steps.release_notes.outputs.release_notes }}
           files: |
             ktlint-cli/build/run/*
-        env :
-          GITHUB_TOKEN : ${{ secrets.GITHUB_TOKEN }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Bump Homebrew Formula
         if: ${{ success() }}
--- a/.github/workflows/publish-release-build.yml
+++ b/.github/workflows/publish-release-build.yml
@@ -14,13 +14,10 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: 'master'
-      - uses: actions/setup-java@v3
-        with:
-          distribution: 'zulu'
-          java-version: 20 # This should match all `jvmToolchain(xxx)` calls in the project
-      - uses: gradle/gradle-build-action@v2
-        with:
-          gradle-home-cache-cleanup: true
+
+
+      - uses: ./.github/actions/setup-gradle-build
+
       - name: Build executable and publish to Maven
         run: ./gradlew clean shadowJarExecutable publishMavenPublicationToMavenCentralRepository --no-daemon --no-parallel --no-configuration-cache
         env:
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -17,6 +17,10 @@ on:
       - '**/*.toml'
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
+
 env:
   ORG_GRADLE_PROJECT_signingKey: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEY }}
   ORG_GRADLE_PROJECT_signingKeyId: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEYID }}
@@ -32,20 +36,20 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
+
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: |
-            8
-            11
-            17
-            20
+          java-version: 20
+
       - uses: gradle/gradle-build-action@v2
         with:
           gradle-home-cache-cleanup: true
+
       - name: Build with release Kotlin version
         run: ./gradlew build ktlintCheck --no-configuration-cache
-      - name: Build with dev Kotlin version
+
+      - name: Build with assemble Kotlin version
         run: ./gradlew -PkotlinDev build ktlintCheck --no-configuration-cache
 
       - name: Check `data class`es are not part of public API
@@ -68,3 +72,41 @@ jobs:
               exit 1
           fi
         shell: bash
+
+  build-dev:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    name: "[build] OS=${{ matrix.os }}, Kotlin=dev"
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: ./.github/actions/setup-gradle-build
+
+      - name: Build with assemble Kotlin version
+        run: ./gradlew -PkotlinDev ktlintCheck build
+
+  tests:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        jdk: [ 8, 11, 17 ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: |
+            8
+            11
+            17
+            20
+
+      - uses: gradle/gradle-build-action@v2
+        with:
+          gradle-home-cache-cleanup: true
+
+      - run: ./gradlew testJdk${{ matrix.jdk }}
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -18,9 +18,9 @@ on:
   workflow_dispatch:
 
 env:
-  ORG_GRADLE_PROJECT_signingKey : ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEY }}
+  ORG_GRADLE_PROJECT_signingKey: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEY }}
   ORG_GRADLE_PROJECT_signingKeyId: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGKEYID }}
-  ORG_GRADLE_PROJECT_signingKeyPassword : ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
+  ORG_GRADLE_PROJECT_signingKeyPassword: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
   CLI_TEST_MAX_DURATION_IN_SECONDS: 10
 
 # Note that "jobs.build.strategy" and "jobs.build.runs-on" should be kept in sync with "pull-request-wth-code"
@@ -28,10 +28,10 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ ubuntu-latest, windows-latest ]
+        os: [ubuntu-latest, windows-latest]
         # When changing the list of JDK versions, the build configuration has to be changed by a repository admin. See
         # https://github.com/pinterest/ktlint/pull/1787#issuecomment-1409074092
-        jdk: [ 8, 11, 17, 20 ]
+        jdk: [8, 11, 17, 20]
         exclude: # windows with JDK8 are *really* flaky
           - os: windows-latest
             jdk: 8
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -110,4 +110,4 @@ jobs:
         with:
           gradle-home-cache-cleanup: true
 
-      - run: ./gradlew testOnJdk -PtestJdkVersion=${{ matrix.jdk }}
+      - run: ./gradlew testOnJdk${{ matrix.jdk }} -PtestJdkVersion=${{ matrix.jdk }}
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -38,14 +38,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: actions/setup-java@v3
-        with:
-          distribution: 'zulu'
-          java-version: 20 # This should match all `jvmToolchain(xxx)` calls in the project
-
-      - uses: gradle/gradle-build-action@v2
-        with:
-          gradle-home-cache-cleanup: true
+      - uses: ./.github/actions/setup-gradle-build
 
       - name: Build with release Kotlin version
         run: ./gradlew build ktlintCheck --no-configuration-cache
@@ -99,15 +92,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: actions/setup-java@v3
+      - uses: ./.github/actions/setup-gradle-build
         with:
-          distribution: 'zulu'
-          java-version: | # last version (set as default) should match all `jvmToolchain(xxx)` calls in the project
-            ${{ matrix.jdk }}
-            20
-
-      - uses: gradle/gradle-build-action@v2
-        with:
-          gradle-home-cache-cleanup: true
+          additional-java-versions: ${{ matrix.jdk }}
 
       - run: ./gradlew testOnJdk${{ matrix.jdk }} -PtestJdkVersion=${{ matrix.jdk }}
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -31,7 +31,7 @@ jobs:
         os: [ ubuntu-latest, windows-latest ]
         # When changing the list of JDK versions, the build configuration has to be changed by a repository admin. See
         # https://github.com/pinterest/ktlint/pull/1787#issuecomment-1409074092
-        jdk: [ 8, 11, 17, 19 ]
+        jdk: [ 8, 11, 17, 20 ]
         exclude: # windows with JDK8 are *really* flaky
           - os: windows-latest
             jdk: 8
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -34,6 +34,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
+    name: "[build] OS: ${{ matrix.os }}"
     steps:
       - uses: actions/checkout@v3
 
@@ -93,6 +94,7 @@ jobs:
         os: [ubuntu-latest, windows-latest]
         jdk: [ 8, 11, 17 ]
     runs-on: ${{ matrix.os }}
+    name: "[tests] OS: ${{ matrix.os }}, Java: ${{ matrix.jdk }}"
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -27,24 +27,21 @@ env:
   ORG_GRADLE_PROJECT_signingKeyPassword: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
   CLI_TEST_MAX_DURATION_IN_SECONDS: 10
 
-# Note that all "jobs" (build, tests) including "jobs.build.runs-on" should be kept in sync with "pull-request-without-code"
+# Note that all "jobs" (build, tests) including "jobs.*.runs-on" should be kept in sync with "pull-request-without-code"
 jobs:
   build:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
-    name: "[build] OS: ${{ matrix.os }}"
+    name: "[build] OS=${{ matrix.os }} Kotlin=stable"
     steps:
       - uses: actions/checkout@v3
 
       - uses: ./.github/actions/setup-gradle-build
 
       - name: Build with release Kotlin version
-        run: ./gradlew build ktlintCheck --no-configuration-cache
-
-      - name: Build with assemble Kotlin version
-        run: ./gradlew -PkotlinDev build ktlintCheck --no-configuration-cache
+        run: ./gradlew ktlintCheck build
 
       - name: Check `data class`es are not part of public API
         run: |
@@ -86,9 +83,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        jdk: [ 8, 11, 17 ] # list of Java versions to run tests against (excluding `java-compilation` version for which tests already have run during `build` job)
+        jdk: [8, 11, 17] # list of Java versions to run tests against (excluding `java-compilation` version for which tests already have run during `build` job)
     runs-on: ${{ matrix.os }}
-    name: "[tests] OS: ${{ matrix.os }}, Java: ${{ matrix.jdk }}"
+    name: "[tests] OS=${{ matrix.os }}, Java=${{ matrix.jdk }}"
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -41,7 +41,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 20
+          java-version: 20 # This should match all `jvmToolchain(xxx)` calls in the project
 
       - uses: gradle/gradle-build-action@v2
         with:
@@ -90,9 +90,10 @@ jobs:
 
   tests:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        jdk: [ 8, 11, 17 ]
+        jdk: [ 8, 11, 17 ] # list of Java versions to run tests against
     runs-on: ${{ matrix.os }}
     name: "[tests] OS: ${{ matrix.os }}, Java: ${{ matrix.jdk }}"
     steps:
@@ -102,13 +103,11 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: |
-            8
-            11
-            17
+            ${{ matrix.jdk }}
             20
 
       - uses: gradle/gradle-build-action@v2
         with:
           gradle-home-cache-cleanup: true
 
-      - run: ./gradlew testJdk${{ matrix.jdk }}
+      - run: ./gradlew testOnJdk -PtestJdkVersion=${{ matrix.jdk }}
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -27,7 +27,7 @@ env:
   ORG_GRADLE_PROJECT_signingKeyPassword: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
   CLI_TEST_MAX_DURATION_IN_SECONDS: 10
 
-# Note that "jobs.build.strategy" and "jobs.build.runs-on" should be kept in sync with "pull-request-wth-code"
+# Note that all "jobs" (build, tests) including "jobs.build.runs-on" should be kept in sync with "pull-request-without-code"
 jobs:
   build:
     strategy:
@@ -93,7 +93,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        jdk: [ 8, 11, 17 ] # list of Java versions to run tests against
+        jdk: [ 8, 11, 17 ] # list of Java versions to run tests against (excluding `java-compilation` version for which tests already have run during `build` job)
     runs-on: ${{ matrix.os }}
     name: "[tests] OS: ${{ matrix.os }}, Java: ${{ matrix.jdk }}"
     steps:
@@ -102,7 +102,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: |
+          java-version: | # last version (set as default) should match all `jvmToolchain(xxx)` calls in the project
             ${{ matrix.jdk }}
             20
 
--- a/.github/workflows/pull-request-with-code.yml
+++ b/.github/workflows/pull-request-with-code.yml
@@ -29,19 +29,17 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        # When changing the list of JDK versions, the build configuration has to be changed by a repository admin. See
-        # https://github.com/pinterest/ktlint/pull/1787#issuecomment-1409074092
-        jdk: [8, 11, 17, 20]
-        exclude: # windows with JDK8 are *really* flaky
-          - os: windows-latest
-            jdk: 8
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: ${{ matrix.jdk }}
+          java-version: |
+            8
+            11
+            17
+            20
       - uses: gradle/gradle-build-action@v2
         with:
           gradle-home-cache-cleanup: true
--- a/.github/workflows/pull-request-without-code.yml
+++ b/.github/workflows/pull-request-without-code.yml
@@ -19,7 +19,7 @@ on:
       - '!**/*.toml'
 
 # Add a dummy job that return true so that a PR not containing any code can be merged to master
-# Note that "jobs.build.strategy" and "jobs.build.runs-on" should be kept in sync with "pull-request-wth-code"
+# Note that all "jobs" (build, tests) including "jobs.build.runs-on" should be kept in sync with "pull-request-with-code"
 jobs:
   build:
     strategy:
--- a/.github/workflows/pull-request-without-code.yml
+++ b/.github/workflows/pull-request-without-code.yml
@@ -19,14 +19,23 @@ on:
       - '!**/*.toml'
 
 # Add a dummy job that return true so that a PR not containing any code can be merged to master
-# Note that all "jobs" (build, tests) including "jobs.build.runs-on" should be kept in sync with "pull-request-with-code"
+# Note that all "jobs" (build, tests) including "jobs.*.runs-on" should be kept in sync with "pull-request-with-code"
 jobs:
   build:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
-    name: "[build] OS: ${{ matrix.os }}"
+    name: "[build] OS=${{ matrix.os }} Kotlin=stable"
+    steps:
+      - run: 'echo "No build required"'
+
+  build-dev:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    name: "[build] OS=${{ matrix.os }}, Kotlin=dev"
     steps:
       - run: 'echo "No build required"'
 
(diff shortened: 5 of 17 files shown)


.editorconfig
```diff
@@ -37,3 +37,6 @@ max_line_length = unset
 
 [gradle/verification-metadata.xml]
 indent_size = 3
+
+[*.yml]
+ij_yaml_spaces_within_brackets = false
```
mateuszkwiecinski, line 42: `yml` files I modified, which got auto-formated by my IDE had unexpected diffs. I'm proposing to add a rule that matches what GitHub does it their docs (i.e. https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) and reformatted all files in a separate commit

.github/workflows/publish-release-build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: 19
+          java-version: 20
```
mateuszkwiecinski, line 20: To clear things out. This is now irrelevant to how the build will behave. Java version is controlled via Gradle toolchains and setting it to Java 20 only improves CI performance since it now doesn't have to be downloaded and installed on every run.
With toolchains the build will behave exactly the same way on CI & on local builds

.github/workflows/pull-request-with-code.yml
```diff
@@ -18,30 +18,28 @@ on:
...
+  ORG_GRADLE_PROJECT_signingKeyPassword: ${{ secrets.ORG_GRADLE_PROJECT_SIGNINGPASSWORD }}
   CLI_TEST_MAX_DURATION_IN_SECONDS: 10
 
 # Note that "jobs.build.strategy" and "jobs.build.runs-on" should be kept in sync with "pull-request-wth-code"
 jobs:
   build:
     strategy:
       matrix:
-        os: [ ubuntu-latest, windows-latest ]
-        # When changing the list of JDK versions, the build configuration has to be changed by a repository admin. See
```
mateuszkwiecinski, line 32: I removed this comment since it won't be relevant anymore (excluding this single time when someone has to changed via settings). If you _prefer_ having seprate checks for each java version, I can create seprate matrix that runs single task: `./gradlew testJdk${{ matrix.jdk }}`.

.github/workflows/pull-request-with-code.yml
```diff
@@ -18,30 +18,28 @@ on:
...
       - uses: actions/checkout@v3
       - uses: actions/setup-java@v3
         with:
           distribution: 'zulu'
-          java-version: ${{ matrix.jdk }}
+          java-version: |
+            8
+            11
+            17
+            20
```
mateuszkwiecinski, line 42: Again, local CI setup is not relevant. This pre-installs java versions only to avoid dowloading them via Gradle on each run

CHANGELOG.md
```diff
@@ -440,6 +440,7 @@ if (node.isRoot()) {
 * For the new code style `ktlint_official`, do not allow wildcard imports `java.util` and `kotlinx.android.synthetic` by default. Important: `.editorconfig` property `ij_kotlin_packages_to_use_import_on_demand` needs to be set to value `unset` in order to enforce IntelliJ IDEA default formatter to not generate wildcard imports `no-wildcard-imports` ([#1797](https://github.com/pinterest/ktlint/issues/1797))
 * Convert a single line block comment to an EOL comment if not preceded or followed by another code element on the same line `comment-wrapping` ([#1941](https://github.com/pinterest/ktlint/issues/1941))
 * Ignore a block comment inside a single line block `comment-wrapping` ([#1942](https://github.com/pinterest/ktlint/issues/1942))
+* Build the project with Java 20, run test on Java 8, 11, 17 and 20
```
mateuszkwiecinski, line 443: Ah, I should probably also include link to the issue 🤦

README.md
```diff
@@ -6,7 +6,7 @@
 
 <p align="center">
 <a href="https://kotlinlang.slack.com/messages/CKS3XG0LS"><img src="https://img.shields.io/badge/slack-@kotlinlang/ktlint-yellow.svg?logo=slack" alt="Join the chat at https://kotlinlang.slack.com"/></a>
-<a href="https://github.com/pinterest/ktlint/actions/workflows/gradle-snapshot-build.yml"><img src="https://github.com/pinterest/ktlint/actions/workflows/gradle-snapshot-build.yml/badge.svg" alt="Build status"></a>
+<a href="https://github.com/pinterest/ktlint/actions/workflows/publish-snapshot-build.yml"><img src="https://github.com/pinterest/ktlint/actions/workflows/publish-snapshot-build.yml/badge.svg" alt="Build status"></a>
```
mateuszkwiecinski, line 9: |Before|After|
|---|---|
|image|image|

build-logic/build.gradle.kts
```diff
@@ -6,14 +8,29 @@ repositories {
...
+
 dependencies {
     val kotlinPlugin =
         if (providers.gradleProperty("kotlinDev").orNull.toBoolean()) {
             // Pass '-PkotlinDev' to command line to enable kotlin-in-development version
             logger.warn("Enabling kotlin dev version!")
             libs.kotlin.plugin.dev
         } else {
-            libs.kotlin.plugin
+            libs.kotlin.plugin.asProvider()
```
mateuszkwiecinski, line 33: There was a compiler warning saying this is implicitly casted to Any

build-logic/build.gradle.kts
```diff
@@ -6,14 +8,29 @@ repositories {
...
+// `build-logic` can produce Java 20 compatible code, but only after the plugin is built with kotlin 1.9,
+// kotlin 1.9 has a fix for https://youtrack.jetbrains.com/issue/KT-57495/Add-JVM-target-bytecode-version-20
+// temporairly using 17, since this is the highest, preinstalled java version on the CI
+val targetJavaVersion = JavaVersion.VERSION_17
+tasks.withType<JavaCompile>().configureEach {
+    options.release.set(targetJavaVersion.majorVersion.toInt())
+}
+tasks.withType<KotlinCompile>().configureEach {
+    kotlinOptions.jvmTarget = targetJavaVersion.toString()
+}
```
mateuszkwiecinski, line 24: These few lines can be removed after you [bump Kotlin to 1.9](https://github.com/pinterest/ktlint/pull/2113). 
`build-logic` is an internal project and given we know the "actual" project is also built with <latest> java version it's fine to target <latest> java as well. 
As mentioned in the comment, I set it to Java 17, only because it is preinstalled on the CI

mateuszkwiecinski: These lines were missing from your first attempt: https://github.com/pinterest/ktlint/pull/1889, this is where the incompatibility came from

build-logic/settings.gradle.kts
```diff
@@ -5,3 +5,7 @@ dependencyResolutionManagement {
         }
     }
 }
+
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.5.0"
```
mateuszkwiecinski, line 10: This is the mechanism that downloads missing java distributions - https://docs.gradle.org/8.2/userguide/toolchain_plugins.html
If these were missing, the build would fail if the runner didn't have all required java versions available locally

mateuszkwiecinski: The plugin has to be applied to both projects ("project" represented by single `settings.gradle` file), since each of them has their own toolchain config

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -8,19 +8,32 @@ plugins {
...
-        jvmTarget.set(JvmTarget.JVM_1_8)
-    }
+    kotlinOptions.jvmTarget = targetJavaVersion.toString()
 }
 
-// compileJava task and compileKotlin task jvm target compatibility should be set to the same Java version.
-// For some reason, we fallback from toolchain using, see https://github.com/pinterest/ktlint/pull/1787
-tasks.withType<JavaCompile>().configureEach {
-    sourceCompatibility = JavaVersion.VERSION_1_8.toString()
-    targetCompatibility = JavaVersion.VERSION_1_8.toString()
```
mateuszkwiecinski, line 23: setting `sourceCompatibility`& `targetCompatibility` can be considered legacy, `options.release` is the preferred way: 
>Although not generally advised, these options were historically used to configure the Java version during compilation.

https://docs.gradle.org/8.2/userguide/building_java_projects.html#sec:java_cross_compilation

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -8,19 +8,32 @@ plugins {
...
 
+val targetJavaVersion = JavaVersion.VERSION_1_8
+tasks.withType<JavaCompile>().configureEach {
+    options.release.set(targetJavaVersion.majorVersion.toInt())
+}
 tasks.withType<KotlinCompile>().configureEach {
-    compilerOptions {
-        jvmTarget.set(JvmTarget.JVM_1_8)
-    }
+    kotlinOptions.jvmTarget = targetJavaVersion.toString()
```
mateuszkwiecinski, line 19: I can keep  `compilerOptions`, AFAIK it doesn't matter. it was easier for me to copy

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -8,19 +8,32 @@ plugins {
...
+        description = "Runs the test suite on JDK $majorVersion"
+        group = LifecycleBasePlugin.VERIFICATION_GROUP
+
+        // Copy inputs from normal Test task.
+        val testTask = tasks.test.get()
+        classpath = testTask.classpath
+        testClassesDirs = testTask.testClassesDirs
+    }
+    tasks.named("check").configure { dependsOn(jdkTest) }
 }
```
mateuszkwiecinski, line 37: ❗ 
This is the core change of this PR. Full explanation can be found on https://jakewharton.com/build-on-latest-java-test-through-lowest-java/
>No user is building these projects on different versions. No user is building these projects at all. Consumers are using the pre-built .jar which we ship to Maven Central built on a single version.

Instead of testing the library build with Java 8, the new flow is to build with Java 20 (link the publish workflow does) and test the bytecode generated by Java (with Java 8 target set) in listed Java runtimes. 

You can now run tests on jdk8 locally (`./gradlew testJdk8`)

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -39,7 +52,7 @@ tasks.withType<Test>().configureEach {
             (Runtime.getRuntime().availableProcessors() / 2).takeIf { it > 0 } ?: 1
         }
 
-    if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_16)) {
```
mateuszkwiecinski, line 42: 1.`JavaVersion.current()` will always be 20, so I updated the code to read test java version from `javaLauncher` 

2. I decreased the condition to compare with Java 11, to avoid non-actionable warnings (which in Java 16 became errors and are the sole reason why additional `jvmArgs` are needed)

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
mateuszkwiecinski: One side-effect is there was a comment JDK 8 tests are wonky on Windows. I haven't seen a details, I don't know why - but if this still is the case, I can ignore `testJdk8` task if running on windows machine

build.gradle.kts
```diff
@@ -44,6 +44,6 @@ tasks.register<JavaExec>("ktlintFormat") {
 
 tasks.wrapper {
     distributionSha256Sum =
-        URL("$distributionUrl.sha256")
+        URI.create("$distributionUrl.sha256").toURL()
```
mateuszkwiecinski, line 47: URL constructora are deprecated in Java 20: 
>The java.net.URL constructors are deprecated. Developers are encouraged to use [java.net.URI](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/net/URI.html) to parse or construct a URL.

https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String

settings.gradle.kts
```diff
@@ -8,13 +8,15 @@ pluginManagement {
 }
 
 dependencyResolutionManagement {
+    repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
```
mateuszkwiecinski, line 11: This is for extra safety this `dependencyResolutionManagement` is the single source of truth for all repositories used in the project

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
Goooler: There is a regression of toolchain using in #1787, see the reason in there.

mateuszkwiecinski: Can you summarize what was the regression specifically, it doesn't seem to be mentioned anywhere 👀

mateuszkwiecinski: Gotcha, thanks for letting me know that CI performance is a concern already 👍 
Given the issue was not about toolchains but rather parallelism and CI performance, I kept toolchains setup (to keep the build deterministic), but I introduced separate jobs which should behave somewhat the same way the current setup.
If this PR gets merged I can look into windows builds performance, for example, from my observations disabling VFS on windows saves ~30% of a CI build time.

.github/workflows/pull-request-with-code.yml
```diff
@@ -17,35 +17,63 @@ on:
       - '**/*.toml'
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
```
mateuszkwiecinski, line 22: New thing, this will cancel ongoing PR builds if someone pushes new commits to it

.github/workflows/pull-request-with-code.yml
```diff
@@ -17,35 +17,63 @@ on:
...
+            8
+            11
+            17
+            20
+
+      - uses: gradle/gradle-build-action@v2
+        with:
+          gradle-home-cache-cleanup: true
+
+      - run: ./gradlew testJdk${{ matrix.jdk }}
```
mateuszkwiecinski, line 79: For reference, I'll cross-link the discussion of why I'm proposing to introduce a separate step to run tests: https://github.com/pinterest/ktlint/pull/2120#discussion_r1257533353

settings.gradle.kts
```diff
@@ -8,13 +8,15 @@ pluginManagement {
...
     repositories {
         mavenCentral()
     }
 }
 
 plugins {
     `gradle-enterprise`
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.6.0"
```
paul-dingemans, line 19: Shouldn't this be managed via the libs.versions.toml? And is it correct that the plugin is required in both the `root` and in `build-logic` version of `settings.gradle.kt`.

paul-dingemans (commented):
Thanks you so much for all explanatory comments on the changes. As I have little Gradle knowledge, I wouldn't have understand what was changed. Now at least, I have some more understanding.

My biggest concern with this change is that there are many locations where a java version is to be changed whenever a new Java version is release or whenever an LTS version is dropped. So all location should be clearly marked with something like `Java version for compilation (e.g. latest release Java version)` or `Supported LTS versions of Java + latest non-LTS version if applicable`.

.github/workflows/publish-release-build.yml
mateuszkwiecinski: >Please add a comment that the latest java version needs to be used for compilation.

to clarify, do you mean to the changelog entry? Or should add an inline comment here? 

>the latest java version needs to be used for compilation.

To be extra clear of how toolchains work: even if you have only Java 11 installed and set as default, Gradle should automatically (via [foojay toolchains](https://github.com/gradle/foojay-toolchains)) fetch the Java version required by the project. So there should be no extra effort required to configure local environment. If you have _any_ Java installed and Gradle is able to run on it, it will compile both projects (`/build-logic` and `/`) using Java version prividet in `jvmToolchain()` call (which is currently set to Java 20).

.github/workflows/pull-request-with-code.yml
mateuszkwiecinski: My apologies, I should have re-visit this comment after I pushed CI latest updates. With my initial implementation, there was no per-JDK-version matrix, but @/Goooler pointed out an extra matrix is required because running all tests on a single GitHub node takes too much time.

>Is it correct that you offer to provide addition gradle tasks that can be used on the local machine to execute tests for a specific java version only?

Not exactly, these tasks are already there (you can checkout this branch and run `./gradlew testJdk11` to run test suite using Java 11 runtime). What I meant was to create a separate workflow step to explicitly run using specified JDK version, which I later did here: https://github.com/pinterest/ktlint/pull/2120/files#diff-0ec67fcdf5c2efde1f1460cace15d6c9c932621ecd1c5b5f2ec8116653436c13R79

I'll try to figure out a way to reduce the number of places Java version is specified, or I will re-add a comment similar to the one I removed

.github/workflows/pull-request-with-code.yml
```diff
@@ -17,35 +17,63 @@ on:
...
+
       - uses: gradle/gradle-build-action@v2
         with:
           gradle-home-cache-cleanup: true
+
       - name: Build with release Kotlin version
-        run: ./gradlew build ktlintCheck --no-configuration-cache
-      - name: Build with dev Kotlin version
-        run: ./gradlew -PkotlinDev build ktlintCheck --no-configuration-cache
+        run: ./gradlew build --no-configuration-cache
```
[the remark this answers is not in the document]
mateuszkwiecinski: Ah, my mistake, good catch! I saw `ktlintCheck` being invoked automatically on every check run ([here](https://github.com/pinterest/ktlint/blob/306325b909de9df9c60306dfa350deb4b614263e/ktlint-ruleset-template/build.gradle.kts#L46)), but now I realized this is true only for a single project, and `ktlintCheck` has to be invoked explicitly here 👍

build-logic/build.gradle.kts
```diff
@@ -6,14 +8,29 @@ repositories {
     mavenCentral()
 }
 
+kotlin {
+    jvmToolchain(20)
+}
+
+// `build-logic` can produce Java 20 compatible code, but only after the plugin is built with kotlin 1.9,
```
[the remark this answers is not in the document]
mateuszkwiecinski: I'll answer here first, and will later create a separate issue: 

>It is not clear to me what should be done and why it should be done.

Project `/build-logic` that holds all Gradle Plugins definitions produces `jar` file that is later appended to `/` build classpath. 

We know the `/` is built with Java 20, so it would be fine `/build-logic` if produced java-20 compatible bytecode - we know  `/` project is able to consume such.
Unfortunately, due to https://youtrack.jetbrains.com/issue/KT-57669/Add-Java-20-to-JvmTarget, `/build-logic` project can't be compiled with Java 20 target yet (JVM target => produced bytecode compatibility). The fix is available in Kotlin Gradle Plugin 1.9, but because `/build-logic` and `/` share a common Version Catalog  (`.toml`) it can't yet be upgraded and both projects use lower Kotlin version. 

Theoretically, due to the fact these are 2 separated projects, `/build-logic` could use Kotlin 1.9, while the `/` still uses Kotlin 1.8.22, but such change is not worth the extra complexity.

Regarding **what** should be done? After `/build-logic` has Kotlin Gradle Plugin 1.9.0 applied, there are 2 (well 3) options:
1. Setting explicit JVM target version here could be completely skipped, and its value would fully rely on the implicit behavior that toolchain JDK = target JVM. (Point 1 [here](https://docs.gradle.org/8.2/userguide/building_java_projects.html#sec:java_cross_compilation)). tl;dr remove the extra code I added, leave only `jvmToolchain(20)` setting.
2. Keep setting explicit JVM target version, but upgrade produced bytecode from Java 17 to Java 20
3. Keep the current config, it's also _fine_, but the KGP won't be able to use new features available in higher Java version when producing JVM bytecode. 

>To what plugin are you referring?

I was referring to all plugins published from `/build-logic` project:
image

Would a
```suggestion
// it would be OK if `build-logic` produced Java 20 compatible bytecode, but this requires using Kotlin Gradle Plugin 1.9, which can't be used yet. visit {{ link to the issue I'll create later }} for more details
```
explain my intentions better? 



Please let me know if anything I'm saying is not clear! 
My main point here is I wanted to stress out this repository consists of 2 separate builds/root projects: 
a) a `/build-logic` - fully internal, produces Gradle Plugins, used to configure `/` project 
b) `/` containing ktlint library, uses `a)` to configure itself, but doesn't expose a) dependencies to ktlint consumers. 
The fact there are 2 separate builds makes the amount of configuration code feel duplicated :)

build-logic/settings.gradle.kts
```diff
@@ -5,3 +5,7 @@ dependencyResolutionManagement {
         }
     }
 }
+
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.6.0"
```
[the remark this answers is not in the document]
mateuszkwiecinski: Will try, although I'm not sure if `settings.gradle` (which is also a separate type of project under the hood 😅) has access to version catalog 👀

settings.gradle.kts
mateuszkwiecinski: >And is it correct that the plugin is required in both the root and in build-logic version of settings.gradle.kt.

Unfortunately yes. I can describe a scenario to demonstrate when each foojay plugin will be used.
Assuming a person has only Java 18 installed locally and wants to run `testJdk11` task, gradle will:
1. Read `/`'s settings.gradle, and will conclude it has to compile `includedBuild` first:
2. Gradle moves its focus to `/build-logic` and attempts to compile it first 
3. it will realize it is missing Java 20, (`/build-logic` sets the requirement via `jvmToolchain(20)` in `/build-logic/build.gradle.kts` <- this is where build-logic's foojay plugin steps in and fetches and configures required Java
2. As a result of successful compilation, it produces `build-logic.jar` with bytecode compatible with corresponding `jvmTarget`
image

4. Gradle will then move back t o`/`. It will eventual learn that `/` also requires Java 20 (via the other `jvmToolchain` call, but it has it installed already, done. 
5. Next important step is it's time to configure the requested task, it has `javaLauncher` set to use Java 11, Gradle _sees_ it has only Java 18 and Java 20 available, so it has to fetch the missing one.  <- this is where the `/`s foojay plugin steps in

If I skipped one of the plugin application it could result in a configuration error, and the only workaround it would be  to require to installing Java X manually. Having both plugins installed, guarantees in all cases the whole required setup configures itself automatically

build-logic/build.gradle.kts
Goooler: If so, we need to do the same thing for `libs.kotlin.plugin.dev`.

build-logic/settings.gradle.kts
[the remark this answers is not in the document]
Goooler: Seems can't.

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
Goooler: If you want to keep toolchain using, maybe we can do something like https://github.com/detekt/detekt/pull/5735.

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
mateuszkwiecinski: >Should keep compilerOptions, kotlinOptions is deprecated.

Can you link the documentation?

Goooler: https://kotlinlang.org/docs/whatsnew18.html#limitations

build-logic/build.gradle.kts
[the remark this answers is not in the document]
paul-dingemans: Clear. And yes, the updated comment is much more clear. Now, I realized that it even can be written as a TODO:
```
TODO: After upgrade to Kotlin Gradle Plugin 1.9, ....
```
Before each release, I check all TODO's and if a constraint is met, I will always try to actually resolve before releasing.

.github/workflows/pull-request-with-code.yml
```diff
@@ -17,35 +17,62 @@ on:
...
       - name: Build with release Kotlin version
         run: ./gradlew build ktlintCheck --no-configuration-cache
-      - name: Build with dev Kotlin version
+
+      - name: Build with assemble Kotlin version
         run: ./gradlew -PkotlinDev build ktlintCheck --no-configuration-cache
+
+  tests:
+    strategy:
+      fail-fast: false
```
mateuszkwiecinski, line 58: 🆕 New thing, I can drop it if you don't like it. 
Previously, if one of the branches failed the remaining ongoing ones were cancelled. With `fail-fast: false` the workflow will continue running, so one will now know if the failure affects single OS x JDK configuration of i.e. all Windows builds are affected

.github/workflows/pull-request-with-code.yml
```diff
@@ -17,35 +17,62 @@ on:
...
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        jdk: [ 8, 11, 17 ] # list of Java versions to run tests against
```
mateuszkwiecinski, line 61: This doesn't have to be kept in sync with buildscript anymore. This is the only place where the tested configuration are defined. If one wants to start testing on Java 27 - they can just add another argument
```kotlin
        jdk: [ 8, 11, 17, 27 ]
```

`actions/setup-java` will install proper Java version, `testOnJdk -PtestJdkVersion=27` will invoke tests with proper config

build-logic/build.gradle.kts
mateuszkwiecinski: >If so, we need to do the same thing for libs.kotlin.plugin.dev.

Not necessarily, `.plugin` is a type of `KotlinPluginLibraryAccessors`, `.dev` is a `Provider<MinimalExternalModuleDependency>` already.

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -5,22 +5,33 @@ plugins {
     kotlin("jvm")
 }
 
+val projectLibs = extensions.getByType<VersionCatalogsExtension>().named("libs")
```
mateuszkwiecinski, line 8: `ktlint-kotlin-common` is a Gradle plugin, so it can't use the synthetic `libs.versions.java.compilation` accessor. I had to extract version the _native_ way via Gradle extensions

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -5,22 +5,33 @@ plugins {
...
+if (requestedJdkVersion != null) {
+    tasks.register<Test>("testOnJdk") {
+        javaLauncher = javaToolchains.launcherFor { languageVersion = JavaLanguageVersion.of(requestedJdkVersion) }
+
+        // Copy inputs from normal Test task.
+        val testTask = tasks.test.get()
+        classpath = testTask.classpath
+        testClassesDirs = testTask.testClassesDirs
+    }
 }
```
mateuszkwiecinski, line 35: This is my new suggestion. My initial approach eagerly registered tasks for all supported java versions. The new approach rely on command line arguments, so if one wants to run test locally they can run:
```cmd
./gradlew testOnJdk -PtestJdkVersion=18
```
which will attempt to run tests against Java 18. 

The only difference is these tasks can't be run via click in IDE + I'm not aware of any possibility to run again multiple java versions within single gradle invocation

gradle/libs.versions.toml
```diff
@@ -1,4 +1,6 @@
 [versions]
+java-compilation = "20" # `actions/setup-java` should install the same version
+java-target = "8"
```
mateuszkwiecinski, line 3: This is my new proposal to define project's Java compatibility. I hope the names are clear 😅 

`java-target` is much more important as it will affect consumers, it defines bytecode compatibility the produced `.jar`s will comply to.

.github/workflows/pull-request-with-code.yml
mateuszkwiecinski: Also, a small note: the tests against _current_ version are invoked as a part of `build` job. If the main pipeline takes too much time, you may consider running them on a separate node, in parallel to the _main_ `build` job. 
This can be done by using `./gradlew build -x test` + adding current java major version to the `jdk` list here.

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -5,22 +5,33 @@ plugins {
...
 
+tasks.withType<JavaCompile>().configureEach {
+    options.release.set(javaTargetVersion.asInt())
+}
 tasks.withType<KotlinCompile>().configureEach {
-    compilerOptions {
-        jvmTarget.set(JvmTarget.JVM_1_8)
-    }
+    compilerOptions.jvmTarget.set(JvmTarget.fromTarget(JavaVersion.toVersion(javaTargetVersion).toString()))
```
mateuszkwiecinski, line 22: This is the _cost_ of having everything defined at one place - I had to juggle between types :/ 
As an input I may have `"8"` or `"11"` string (representing major java language version) which has to be transformed into kotlin's `JvmTarget` class which expects `"1.8"` or `"11"` string 🤷 The solution I can up with goes through Gradle `JavaVersion` which is able to transform between all variants/types, but maybe there is any better option? 🤷

mateuszkwiecinski (commented):
I pushed another proposal, with less _distributed_ config:
- Java compilation settings are now driven by versions defined in `.toml` file
- Tested Java versions are now defined only within workflow file
- Java version hardcoded in workflow files still have to be kept in sync, so I added necessary comments reminding to update the value

@paul-dingemans Please let me know if this suits you any better, or if the complexity in the configuration is not worth having since source of truth for Java setup

.github/workflows/pull-request-with-code.yml
paul-dingemans: I think it is fine to add this. It might take a couple of minutes more before the build fails as the Windows builds are considerably slower than the Ubuntu builds.

One thing that would be very nice is if the `ktlintCheck` is run on all modules before the tests are executed.

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
paul-dingemans: Usually, I only build against latest java version on my local machine via IDE click. I have trouble to remember all gradle command :-( 
It might be an idea to document the build commands for ktlint in the documentation. That would not only help me, but also other developers on the project.

gradle/libs.versions.toml
paul-dingemans: Maybe change to something like:
```
# The java-compilation version is the latest supported (not necessarily LTS) version of Java. It should be identical to Java-version used in `actions/setup-java`
java-compilation = "20"
# The java-target version is the lowest supported LTS version of Java. Jar's produced are bytecode compatible with this version.
java-target = "8"
```

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
paul-dingemans: Indeed, not the most pretty one. But with a comment like `Convert Java version (e.g. "1.8" or "11") to Kotlin JvmTarget ("8" resp. "11")` it is clear enough.

.github/workflows/pull-request-with-code.yml
paul-dingemans: Please document that current java-version does not need to be listed here.
```
jdk: [ 8, 11, 17 ] # list of Java versions (excluding java-compatibility version for which tests already have run during build)  to run tests against
```

.github/workflows/pull-request-with-code.yml
```diff
@@ -17,35 +17,62 @@ on:
...
+    name: "[tests] OS: ${{ matrix.os }}, Java: ${{ matrix.jdk }}"
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: |
+            ${{ matrix.jdk }}
+            20
```
paul-dingemans, line 72: Add ```# This should match all `jvmToolchain(xxx)` calls in the project```?

.github/workflows/pull-request-without-code.yml
```diff
@@ -24,13 +24,18 @@ jobs:
...
-        exclude: # windows with JDK8 are *really* flaky
-          - os: windows-latest
-            jdk: 8
+        os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
+    name: "[build] OS: ${{ matrix.os }}"
+    steps:
+      - run: 'echo "No build required"'
+
+  tests:
```
paul-dingemans, line 33: This step should also be mentions in the comment on line 23.

build-logic/build.gradle.kts
```diff
@@ -6,14 +9,29 @@ repositories {
...
 }
 
+kotlin {
+    jvmToolchain(libs.versions.java.compilation.get().toInt())
+}
+
+// TODO: Remove setting `options.release` and `compilerOptions.jvmTarget` after upgrade to Kotlin Gradle Plugin 1.9
+//  it would be OK if `build-logic` produced Java 20 compatible bytecode, but setting target to Java 20 is supported only in KGP 1.9
+//  @see https://github.com/pinterest/ktlint/pull/2120#discussion_r1260229055 for more details
+val targetJavaVersion = JavaVersion.VERSION_17
```
paul-dingemans, line 19: The name `targetJavaVersion` is now confusing compared to `java-target` in `gradle/libs.versions.toml`.

paul-dingemans (changes requested):

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
mateuszkwiecinski: >I only build against latest java version on my local machine via IDE click.

I personally use IDE only to edit the code, I run all commands via terminal 😅 
If you prefer running tasks via IDE, I restored that behavior, and you should now see all additional tasks available under `verification` subtree:
image

.github/workflows/pull-request-without-code.yml
mateuszkwiecinski: I noticed the comment was initially added to the whole `steps` block, so instead of adding a comment, I updated the existing one to mention all steps rather than just `jobs.build`. That's what I believe was the original intention (to keep all jobs in sync, regardless of their number 👀

build-logic/build.gradle.kts
mateuszkwiecinski: I renamed the variable to `buildLogicTargetJavaVersion`, I hope that's fine for now (+ I hope this code will be removed soon™, after KGP 1.9 upgrade)

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -5,22 +5,38 @@ plugins {
...
 }
 
-// compileJava task and compileKotlin task jvm target compatibility should be set to the same Java version.
-// For some reason, we fallback from toolchain using, see https://github.com/pinterest/ktlint/pull/1787
-tasks.withType<JavaCompile>().configureEach {
-    sourceCompatibility = JavaVersion.VERSION_1_8.toString()
-    targetCompatibility = JavaVersion.VERSION_1_8.toString()
+val requestedJdkVersion = project.findProperty("testJdkVersion")?.toString()?.toInt()
+// List all non-current Java versions the developers may want to run via IDE click
+setOf(8, 11, 17, requestedJdkVersion).filterNotNull().forEach { version ->
```
mateuszkwiecinski, line 28: I tried to explain that all listed values here will show up in the IDE. Most likely this should match the ones CI tests against and be LTS version, but it doesn't have to. The list here is responsible only for tasks that will show up in IDE.

A set of pre-registered tasks are available, but also, one can pass custom `testJdkVersion` (and that's what CI will leverage). `setOf().filterNotNull()` makes sure the build script doesn't register test task for the same Java version twice

.github/workflows/pull-request-with-code.yml
mateuszkwiecinski: I couldn't add the line directly next to `20`, so I added it to the closest place possible (and I hope it'll pass the CI).
Heads up: I have already another improvement to the CI queued up ([preview](https://github.com/mateuszkwiecinski/ktlint/compare/set_toolchains...mateuszkwiecinski:ktlint:toolchains_testing)), where I'll try to unify how workflows set up the environment for Gradle builds (so I don't have to hardcode <current> java version that many times).

.github/workflows/pull-request-with-code.yml
mateuszkwiecinski: >One thing that would be very nice is if the ktlintCheck is run on all modules before the tests are executed.

Do I understand correctly, you'd imagine having a
```
step 1: ktlintCheck
step 2 (in parallel)
	- build + build kotlinDev
	- testOnJdk8
	- testOnJdk11
    - testOnJdk17
```
so to basically defer both build and tests until `ktlintCheck` completes? Or did you mean something else in mind? 

❔ Additional question: Should I move `kotlinDev` to a separate job as well? 

I'll open another PR targeting this one, to have clear diff, pl to avoid updating branch protection settings more than once 👀

paul-dingemans (commented):
Just saw that I had some more pending review comments.

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
paul-dingemans: Cool. Could be written as:
```
setOfNotNull(8, 11, 17, requestedJdkVersion).forEach { version ->
```

.github/workflows/pull-request-with-code.yml
paul-dingemans: Yes, lets discuss that in the other PR. But you understood the gist of my idea. I am not sure what moving `kotlinDev` to a separate job would entail and do no see any direct benefit from it. I will leave that up to you to decide.

build-logic/src/main/kotlin/ktlint-publication.gradle.kts
```diff
@@ -7,13 +7,14 @@ plugins {
...
+val definedVersion: String = providers.gradleProperty("VERSION_NAME").get()
 
-project.version = providers.gradleProperty("VERSION_NAME").orNull
-    ?: throw GradleException("Project version property is missing")
+project.version =
+    if (hasProperty("kotlinDev")) {
+        "${definedVersion.removeSuffix("-SNAPSHOT")}-kotlin-dev-SNAPSHOT"
+    } else {
+        definedVersion
+    }
```
mateuszkwiecinski, line 17: ❗ 
Here's another thing I claim didn't work, but I might be wrong, I haven't found anything mentioning `isKotlinDev` property, so I assumed this should match `kotlinDev` property used elsewhere. 
Looking at the implementation, it concluded the intention was to publish artifacts using kotlinDev version to follow their own versioning, I updated the configuration and run `./gradlew publishToMavenLocal -PkotlinDev` to verify this will now follow previous naming convention:
image

Please let me know if this is not intended, and I should rather remove the `isKotlinDev` if L10-13, or if it serves some other purpose

gradle/libs.versions.toml
```diff
@@ -1,6 +1,10 @@
 [versions]
+# The java-compilation version is the latest supported (not necessarily LTS) version of Java. It should be identical to Java-version used in `actions/setup-java`
+java-compilation = "20"
+# The java-target version is the lowest supported LTS version of Java. Jar's produced are bytecode compatible with this version.
+java-target = "8"
 kotlin = "1.8.22"
-kotlinDev = "1.8.22"
+kotlinDev = "1.9.0"
```
mateuszkwiecinski, line 7: ❗ 
To showcase the new setup now properly uses kotlinDev version, I updated `kotlinDev` version to 1.9.0. I hoped it also would be helpful to work on https://github.com/pinterest/ktlint/issues/2061, so I decided to keep it. 
Please let me know if I misunderstood its purpose, and it doesn't mean to be used as Kotlin version for local Kotlin development 👀

mateuszkwiecinski (commented):
Welp, I couldn't create separate PR, targeting this one since this branch lives only in my forked repository :/ 
I pushed another set of commits, on top of old ones, but if you want to see the diff alone, it can be viewed here: https://github.com/mateuszkwiecinski/ktlint/pull/3 🤷

.github/workflows/pull-request-with-code.yml
mateuszkwiecinski: I'll mark this discussion as resolved, I pushed another set of changes and commented them out: https://github.com/pinterest/ktlint/pull/2120#discussion_r1264650925, let's continue the discussion there :)

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
mateuszkwiecinski: I wasn't sure which part you were referring to specifically (the link leads to PR not to specific comment or line of code), but I pushed another round of improvement. Feel free to review them if you want :) 

>If you want to keep toolchain using

I'd only clarify, it's not like I _want_, I don't know any other solution to make the build deterministic. Checking out `master` locally and running `./gradlew` throws:
```
* What went wrong:
Execution failed for task ':build-logic:compileKotlin'.
> Error while evaluating property 'compilerOptions.jvmTarget' of task ':build-logic:compileKotlin'.
   > Failed to calculate the value of property 'jvmTarget'.
      > Unknown Kotlin JVM target: 20
```
I'm not aware of any other solution that fixes the root cause of issue, so toolchains are not my _preference_ but the only solution I know, so they are somewhat _required_ 😉

.github/actions/setup-gradle-build/action.yml
```diff
@@ -0,0 +1,33 @@
+name: Setup Gradle
+description: Sets up the environment to run Gradle
+
+inputs:
+  gradle-jvm-args:
+    description: "JVM args to pass to Gradle"
+    required: true
+    default: "-Xmx1g"
```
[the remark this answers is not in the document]
paul-dingemans: Please add this explanation as comment in the code. It provides the reasoning for this setting.

.github/actions/setup-gradle-build/action.yml
```diff
@@ -0,0 +1,33 @@
...
+      with:
+        distribution: 'zulu'
+        java-version: | # last version (set as default) should match all `jvmToolchain(xxx)` calls in the project
+          ${{ inputs.additional-java-versions }}
+          20
+
+    - name: Write Gradle build properties to local `gradle.properties`
+      run: |
+        mkdir -p ~/.gradle
+        printf "org.gradle.jvmargs=${{ inputs.gradle-jvm-args }}\n" >> ~/.gradle/gradle.properties
```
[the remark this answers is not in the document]
paul-dingemans: You often have very good explanations in your review remarks WHY you do something. This is one more example for which I would love to have it documented in the code. In this example changing the name to `Optimize Gradle build properties for CI` provides the context while from the steps itself I can easily what is happening (writing properties to a file).

build-logic/src/main/kotlin/ktlint-publication.gradle.kts
paul-dingemans: I think that your reasoning is valid (see https://pinterest.github.io/ktlint/0.50.0/install/snapshot-build/#kotlin-development-version-snapshot). Maybe this got lost when rewriting the build from groovy to kotlin script. But, I don't  expect anyone ever has used the `kotlin-dev-SNAPSHOT`...

paul-dingemans:
We have to wait till next week before @shashachu  is back from holidays. I cannot change the configuration so that the remaining checks are run as well.

paul-dingemans:
@mateuszkwiecinski I am not sure whether we are both waiting for the other to proceed. As far as I know, there are still some conversations which I have not resolved because they are not fixed, but I could be wrong.

mateuszkwiecinski:
Yes, sorry, I saw your comments and I still plan to address them as soon as I find some time. Both here and in the other PR I have opened.
If you wanted to speed things up, you should be able to push to the branch, if not, I should be able to take another _round_ tomorrow or on Thursday :)

paul-dingemans:
> Yes, sorry, I saw your comments and I still plan to address them as soon as I find some time. Both here and in the other PR I have opened. If you wanted to speed things up, you should be able to push to the branch, if not, I should be able to take another _round_ tomorrow or on Thursday :)

Sure, no problem. Just wanted to make sure that we are aligned.

.github/actions/setup-gradle-build/action.yml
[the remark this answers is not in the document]
mateuszkwiecinski: >You often have very good explanations in your review remarks WHY you do something

That's a positive feedback! Thanks! 😄 

The step name should now be updated + I added a few more lines elaborating each optimization

.github/actions/setup-gradle-build/action.yml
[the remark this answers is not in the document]
mateuszkwiecinski: Comment added ✅

I haven't got used to write code comments yet 😅 Personally, in general, I don't write code comments since they get outdated easily. 
When debugging stuff I usually look at the corresponding Pull Request to see the discussion with other people, how the code evolved and what other alternative approaches were considered. This IDE feature is extremely helpful:
image
I'll try to bear in mind to comment out things I normally would put as an explanation comment in a PR :)

mateuszkwiecinski:
Oookay, I think I addressed all remaining items now. Please let me know in case I missed something 👀 

If you manage to merge the new CI pipeline, I'll monitor build scans for a week or two to see if there are any bottlenecks or low-hanging fruits to improve CI build times 👀

.github/actions/setup-gradle-build/action.yml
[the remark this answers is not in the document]
paul-dingemans: I agree with:
> I haven't got used to write code comments yet 😅 Personally, in general, I don't write code comments since they get outdated easily.

But comments which explain the reason why something is done are less vulnerable to that opposed to comments that try to explain what the code does.

paul-dingemans:
I have merged #2136 earlier today. As of that a merge conflict appead in this PR. I am not sure how to resolve. So please have a look.

shashachu:
I think the checks have been updated properly now.

paul-dingemans:
> I think the checks have been updated properly now.

Looks like it. The build succeeded. I will merge.

build-logic/src/main/kotlin/ktlint-kotlin-common.gradle.kts
```diff
@@ -6,22 +6,38 @@ plugins {
...
+    // Convert Java version (e.g. "1.8" or "11") to Kotlin JvmTarget ("8" resp. "11")
+    compilerOptions.jvmTarget.set(JvmTarget.fromTarget(JavaVersion.toVersion(javaTargetVersion).toString()))
 }
 
-// compileJava task and compileKotlin task jvm target compatibility should be set to the same Java version.
-// For some reason, we fallback from toolchain using, see https://github.com/pinterest/ktlint/pull/1787
-tasks.withType<JavaCompile>().configureEach {
-    sourceCompatibility = JavaVersion.VERSION_1_8.toString()
-    targetCompatibility = JavaVersion.VERSION_1_8.toString()
+val requestedJdkVersion = project.findProperty("testJdkVersion")?.toString()?.toInt()
```
Goooler, line 27: We should keep using lazy APIs, https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#gradleProperty-java.lang.String-

mateuszkwiecinski: What's the value of doing extra abstraction if the property is consumed immediately, at the configuration time? 

Lazy APIs are useful to take the burden off the single-threaded configuration phase and do more things during task execution time. The documentation: https://docs.gradle.org/current/userguide/lazy_configuration.html#lazy_properties doesn't mention our setup as an example use case to showcase. 
I'd still suggest using simpler version, as long as it's not deprecated, or adding extra complexity comes with _any_ performance benefit 🤷

Goooler: It's the Gradle recommending way, I'd prefer to follow the best practice.

mateuszkwiecinski: > It's the Gradle recommending way

Can you link where Gradle recommends using providers in use cases like this? I agree it's in general recommended to:
1. wire Gradle models
2. wiring Task inputs & outputs
3. Avoiding doing work during configuration phase
None of them are applicable here, and blindly using lazy APIs doesn't bring any value.

I can also link [this](https://docs.gradle.org/current/userguide/lazy_configuration.html#lazy_configuration_faqs): which says:
>Avoid simplifying calls like obj.getProperty().get() and obj.getProperty().set(T) in your code by introducing additional getters and setters.

which is exactly what you'd have to do if you used `gradleProperty()` api 🤷 

An alternative approach to using native Gradle APIs (like I did), is to use [kts delegates](https://docs.gradle.org/current/userguide/kotlin_dsl.html#kotdsl:properties), in this case:`val testJdkVersion: String? by project`, but that's a syntactic sugar and doesn't change behavior (in context of being lazy or not) 🤷 

I'll be happy to be proven wrong, otherwise forcing to use lazy for properties that are consumed only during configuration phase is useless 🤷

PR #2136, opened 2023-07-16
Replace kotlin public `data class`es with Poko compiler plugin generated ones

### LINKED ISSUE #2133: Reduce probability of breaking binary compatibility - stop using `data class`es in public API
## Expected Behavior

As per https://kotlinlang.org/docs/jvm-api-guidelines-backward-compatibility.html#don-t-use-data-classes-in-an-api data classes aren't recommended to be used in public library API, since they can easily lead to breaking binary compatibility.

Related issues: 
- https://github.com/pinterest/ktlint/issues/1975
- https://github.com/pinterest/ktlint/issues/2111

## Current Behavior

data classes are used in the public api, which may have impact on post-1.0 binary compatibility

## Additional information
I'm creating this issue as a result of [this](https://github.com/pinterest/ktlint/pull/2131#discussion_r1264526383) comment.

mateuszkwiecinski:
Fixes #2133

## Description

This is a _quick_ attempt to replace public `data class`es with `equals`/`hashCode`/`toString` methods generated by [Poko](https://github.com/drewhamilton/Poko) compiler plugin. 
The alternative are:
1. Generate `equals`/`hashCode`/`toString` manually via IDE - but this is the least preferred option, since it's super easy to forget updating their implementation, when modifying the class
2. Go through the code and identify all pieces of code relying on `equals/`hashCode` (calls like `==`, `List#distinct()` or `Collection#contains` etc) and then refactor the implementation to use custom comparators.
3. Keep using data classes, but never modify them

I'm opening this PR to gather some feedback, show the diff as a proof how it can be done, and maybe someone has any better idea 😉

## Checklist

- [ ] Tests are added
- [X] PR description added


ktlint-test/src/main/kotlin/com/pinterest/ktlint/test/KtLintAssertThat.kt
```diff
@@ -652,12 +653,15 @@ internal class MissingEolMarker : RuntimeException(
...
  */
-public data class LintViolation(
-    val line: Int,
-    val col: Int,
-    val detail: String,
-    val canBeAutoCorrected: Boolean = true,
-)
+@Poko
+public class LintViolation
+    @JvmOverloads
```
mateuszkwiecinski, line 658: ❗ 
This is unrelated, I just spotted, this class could use an extra interoperability with java, and I made the _default_ constructor explicitly defined, so it can be called from Java. 
This change is reflected in public api here:
image

paul-dingemans (commented):
This worked out differently than I expected beforehand when just (quickly) reading the Poko documentation. Overall, the code changes do not seem too intrusive to me.

Would it be possible to define some gradle task that fails the build it it finds something like "public data class" in the source code? Because, I really believe that I will forget about not using data class in public API.

Also nice to see that the binary-compatibility-validator requires changes.

mateuszkwiecinski:
>Would it be possible to define some gradle task that fails the build it it finds something like "public data class" in the source code? Because, I really believe that I will forget about not using data class in public API.

I did a bit of research, but I couldn't find any _standard_ solution. If I had to come up with some custom ideas then:
1. cheapest: have a bash script that iterates through all `.api` files and fails if it finds a line with `component1()` function signature (may produce false negatives if someone adds function named `component1()`, but that feels rather unlikely scenario)
2. probably most proper one: have a custom ktlint rule that fails if `public data class` is used, that should be doable, right? 
3. Not have extra checks, and pay extra attention when reviewing binary validator diffs

paul-dingemans:
> > Would it be possible to define some gradle task that fails the build it it finds something like "public data class" in the source code? Because, I really believe that I will forget about not using data class in public API.
> 
> I did a bit of research, but I couldn't find any _standard_ solution. If I had to come up with some custom ideas then:
> 
> 1. cheapest: have a bash script that iterates through all `.api` files and fails if it finds a line with `component1()` function signature (may produce false negatives if someone adds function named `component1()`, but that feels rather unlikely scenario)

Alternative to scanning `.api` files we could also grep the source itself for `public data class` or `protected data class`. But this also quite hacky.

> 2. probably most proper one: have a custom ktlint rule that fails if `public data class` is used, that should be doable, right?

It would be possible, but it would be a rule that should not be exposed because public data classes are valid to be used in case ktlint is used on a project that is not exposing an API.

> 3. Not have extra checks, and pay extra attention when reviewing binary validator diffs

True, except that I made this remark because I know that I going to forget about it. Also, it doesn't warn a contributor about making this mistake.

Shall we try out the hacky option 1?

mateuszkwiecinski:
>Alternative to scanning .api files we could also grep the source itself for public data class or protected data class. But this also quite hacky.

Scanning `.api` files feels much safer since it is somewhat guaranteed it checks only the public API surface (scanning sources is susceptible to data classes used within non-published project, within test sources or within code comments)

>It would be possible, but it would be a rule that should not be exposed

Absolutely 👍 Assuming this project used one of the gradle plugins, my idea was to do something like:
```
dependencies {
    ktlintRuleSet(project(":completely-internal-ruleset-just-for-ktlint-project-not-published-anywhere"))
```
but with CLI, I'd have to produce `.jar` file first, and only then append it somewhere to CLI invocation? idk, I haven't worked with CLI 👀 

Anyway, in the first iteration, I can propose a bash script that runs on CI. If the bash script turns out unreliable and/or if I find enough motivation - I'll try to come up with a project-specific Rule for ktlint, as a replacement for the bash script.

Here's how a failure will look like:
Screenshot 2023-07-26 at 21 09 34

paul-dingemans:
For now, let's proceed with the current check you have added.

PR #2192, opened 2023-08-22
Multiline function literal with single statement

paul-dingemans:
## Description

The idea of this change was to rewrite multiline function literals which only contains one single line statement:
```
val foo =
    listOf(1, 2, 3)
        .filter {
            it > 2
        }.count()
```
to a single line function literal given that it fits on a single line:
```
val foo =
    listOf(1, 2, 3)
        .filter { it > 2 }
        .count()
```

When applying this changes on the Ktlint code base itself, the results were mixed. There was a number of cases where rewriting the code, resulted in nicer code. Unfortunately, there were also a number of cases in which the formatted code was less readable or too different from common usage.

Before:
```
tasks.withType<JavaCompile>().configureEach {
    options.release.set(buildLogicTargetJavaVersion.majorVersion.toInt())
}
```
After:
```
tasks.withType<JavaCompile>().configureEach { options.release.set(buildLogicTargetJavaVersion.majorVersion.toInt()) }
```

Before:
```
dependencyResolutionManagement {
    versionCatalogs {
        create("libs") {
            from(files("../gradle/libs.versions.toml"))
        }
    }
}
```
After:
```
dependencyResolutionManagement {
    versionCatalogs { create("libs") { from(files("../gradle/libs.versions.toml")) }  }
}
```

## Checklist

- [X] Tests are added
- [X] PR description added

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -1,13 +1,9 @@
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
-plugins {
-    `kotlin-dsl`
-}
+plugins { `kotlin-dsl` }
 
-repositories {
-    mavenCentral()
-}
+repositories { mavenCentral() }
 
 kotlin {
     jvmToolchain(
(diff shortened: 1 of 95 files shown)


paul-dingemans:
Because of unwanted side effects, this should not be merged.

PR #2194, opened 2023-08-22
Define EditorConfigOverride for dynamically loaded ruleset

### LINKED ISSUE #2190: Provide EditorConfigOverride API which can be defined at runtime
Based on: https://kotlinlang.slack.com/archives/CKS3XG0LS/p1692379551133839

If an API consumer calls the `KtlintRuleEngine` it can provide an `EditorConfigOverride`. Currently, this `EditorConfigOverride` can only be defined via factory method `from(vararg properties: Pair<EditorConfigProperty<*>, *>): EditorConfigOverride`.  The advantage from ktlint perspective is that it can be guarranteed at compile-time that the `EditorConfigOverride` only contains valid `EditorConfigProperty`. The downside of it, is that the API consumer must include the definitions of the `EditorConfigProperty` at compile time as well. Typically the `EditorConfigProperty` are defined inside rules which can be provided by different artifacts (either ktlint or provided by another rule provider).

From perspective of `ktlint-gradle` (and possibly also for other API consumer), it is preferred to provide the artifacts containing the `EditorConfigProperty` definitions only  as runtime dependencies. This requires an API to build and `EditorConfigOverride` which is more Lazily evaluated. The signature of the factory method, could look like `from(vararg properties: Pair<String, String>): EditorConfigOverride` where the pair contains the name  and value of a `.editorconfig` property.
When a Lazy `EditorConfigOverride` is provided to the `KtlintRuleEngine`, the properties provided have to be matched with properties that are used by the rules that are provided. Properties which can not be matched with a property provided by the rules can be ignored silently (but with a log warning). Properties which can be matched but have an invalid value according to the EditorConfigProperty definition will throw an exception before files are processed.

paul-dingemans:
## Description

Extend API Consumer example to showcase how EditorConfigOverride can be defined with dynamically loaded rulesets

Closes #2190

## Checklist

- [ ] Tests are added
- [ ] PR description added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 Class "org.jetbrains.kotlin.com.intellij.treeCopyHandler" is no longer registered as extension point for the compiler as this is not supported in Kotlin 1.9. Please test your custom rules. In case of unexpected exceptions during formatting of code, see [#2044](https://github.com/pinterest/ktlint/pull/2044) for possible remediation.
 
+#### EditorConfigOverride - Ktlint API Consumer example
+
+The `EditorConfigOverride` parameter of the `KtlintRuleEngine` can be defined using the factory method `EditorConfigOverride.from(vararg properties: Pair<EditorConfigProperty<*>, *>)`. This requires the `EditorConfigProperty`'s to be available at compile time. Some common `EditorConfigProperty`'s are defined in `ktlint-rule-engine-core` which is loaded as transitive dependency of `ktlint-rule-engine` and as of that are available at compile.
+
+If an `EditorConfigProperty` is defined in a `Rule` that is only provided via a runtime dependency, it gets a bit more complicated. The `ktlint-api-consumer` example has now been updated to show how the `EditorConfigProperty` can be retrieved from the `Rule`.
+
 ### Added
 
 * Add experimental rule `class-signature`. This rule rewrites the class header to a consistent format. In code style `ktlint_official`, super types are always wrapped to a separate line. In other code styles, super types are only wrapped in classes having multiple super types. Especially for code style `ktlint_official` the class headers are rewritten in a more consistent format. See [examples in documentation](https://pinterest.github.io/ktlint/latest/rules/experimental/#class-signature). `class-signature` [#875](https://github.com/pinterest/ktlint/issues/1349), [#1349](https://github.com/pinterest/ktlint/issues/875)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ If an `EditorConfigProperty` is defined in a `Rule` that is only provided via a 
 * Add experimental rule `function-expression-body`. This rule rewrites function bodies only contain a `return` or `throw` expression to an expression body. [#2150](https://github.com/pinterest/ktlint/issues/2150)
 * Add new experimental rule `statement-wrapping` which ensures function, class, or other blocks statement body doesn't start or end at starting or ending braces of the block ([#1938](https://github.com/pinterest/ktlint/issues/1938)). This rule was added in `0.50` release, but was never executed outside the unit tests. The rule is now added to the `StandardRuleSetProvider` ([#2170](https://github.com/pinterest/ktlint/issues/2170))
 * Add experimental rule `chain-method-continuation` to the `ktlint_official` code style, but it can be enabled explicitly for the other code styles as well. This rule requires the operators (`.` or `?.`) for chaining method calls, to be aligned with each other. This rule is enabled by ([#1953](https://github.com/pinterest/ktlint/issues/1953))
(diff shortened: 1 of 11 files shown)


paul-dingemans:
The `EditorConfigOverride` is defined as:
```
    val editorConfigOverride =
        EditorConfigOverride
            .from(
                // Property types provided by ktlint-rule-engine-core
                INDENT_STYLE_PROPERTY to IndentConfig.IndentStyle.SPACE,
                INDENT_SIZE_PROPERTY to 4,
                EXPERIMENTAL_RULES_EXECUTION_PROPERTY to RuleExecution.enabled,
                // Properties defines in the ktlint-ruleset-standard can only be used statically when that dependency is provided at compile
                // time.
                // FUNCTION_BODY_EXPRESSION_WRAPPING_PROPERTY to always
            ).plus(
                // For properties that are defined in rules for which the dependency is provided at runtime only, the property name can be
                // provided as String and the value as Any type. In case the property value is invalid, the KtlintRuleEngine logs a warning.
                ruleProviders.findEditorConfigProperty("ktlint_function_signature_body_expression_wrapping") to "alwaysx",
                // In case an unknown property would be provided, an exception is thrown by the helper method
                // ruleProviders.findEditorConfigProperty("unknown_property") to "some-value",
            )
```

The helper method `ruleProviders.findEditorConfigProperty(...)` is defined in the ApiConsumer as:
```
private fun Set<RuleProvider>.findEditorConfigProperty(propertyName: String): EditorConfigProperty<*> {
    val properties =
        map { it.createNewRuleInstance() }
            .flatMap { it.usesEditorConfigProperties }
            .distinct()
    return properties
        .find { it.type.name == propertyName }
        ?: throw EditorConfigPropertyNotFoundException(
            properties
                .map { it.type.name }
                .sorted()
                .joinToString(
                    prefix = "Property with name '$propertyName' is not found in any of given rules. Available properties:\n\t",
                    separator = "\n\t",
                ) { "- $it" },
        )
}
```

When running the main method in the `KtlintApiConsumer` the following is logged.

Both the statically typed properties as well as the property from a dynamically loaded ruleset are recognized by ktlint:
```
[main] DEBUG com.pinterest.ktlint.rule.engine.internal.EditorConfigLoader - Effective editorconfig properties for file 'ktlint-api-consumer/src/main/kotlin/com/example/ktlint/api/consumer/KtlintApiConsumer.kt':
	tab_width: 4
	indent_style: SPACE
	indent_size: 4
	ktlint_experimental: enabled
	ktlint_function_signature_body_expression_wrapping: alwaysx
```

The invalid value for the property of a dynamically loaded ruleset is reported in same way as for static properties:
```
[main] WARN com.pinterest.ktlint.rule.engine.core.api.editorconfig.EditorConfig - Editorconfig property 'ktlint_function_signature_body_expression_wrapping' contains an invalid value 'alwaysx'. The default value 'default' is used instead.
```

paul-dingemans:
New API `EditorConfigPropertyRegistry` is added. It can be used as follows to initialized the `EditorConfigOverride` in case rule sets are available at runtime only:
```
    val editorConfigPropertyRegistry = EditorConfigPropertyRegistry(ruleProviders)

    // Providing rule dependencies at compile time has the advantage that properties defined in those rules can be used to build the
    // EditorConfigOverride using static types. However, providing the rule dependencies at runtime can offer flexibility to the API
    // Consumer. It comes with the cost that the EditorConfigOverride can not be build with static types.
    val editorConfigOverride =
        EditorConfigOverride
            .from(
                // Properties provided by ktlint-rule-engine-core are best to be loaded statically as they are available at compile time as
                // they are provided by the ktlint-rule-engine-core module.
                INDENT_STYLE_PROPERTY to IndentConfig.IndentStyle.SPACE,
                INDENT_SIZE_PROPERTY to 4,
                EXPERIMENTAL_RULES_EXECUTION_PROPERTY to RuleExecution.enabled,
                //
                // Properties defined in the ktlint-ruleset-standard can only be loaded statically when that dependency is provided at
                // compile time. In this example project this ruleset is loaded at runtime, so following decommenting next line results in
                // a compilation error:
                // FUNCTION_BODY_EXPRESSION_WRAPPING_PROPERTY to always
                //
                // For properties that are defined in rules for which the dependency is provided at runtime only, the property name can be
                // provided as String and the value as Any type. As the property is not available at compile time, the value has to be
                // specified as a String (exactly as would be done in the `.editorconfig` file).
                // In case the property value is invalid, the KtlintRuleEngine logs a warning.
                editorConfigPropertyRegistry.find("ktlint_function_signature_body_expression_wrapping") to "always",
                //
                // The properties for enabling/disabling a rule or entire rule set can be set as well. Note that the values of this
                // property can be set via the `RuleExecution` enum which is available at compile time as it is provided by the
                // ktlint-rule-engine-core module.
                editorConfigPropertyRegistry.find("ktlint_standard_function-signature") to RuleExecution.disabled,
                editorConfigPropertyRegistry.find("ktlint_standard") to RuleExecution.disabled,
                //
                // In case an unknown property is provided, an exception is thrown:
                // ruleProviders.findEditorConfigProperty("unknown_property") to "some-value",
            )
```

PR #2282, opened 2023-09-28
Add maven relocation

paul-dingemans:
## Description

Add maven relocation block to `ktlint-cli` and `ktlint-cli-reporter-baseline`

When publishing to the local Maven repository, a `0.51.0-FINAL` version of the `pom.xml` is created for the modules above. The modules contain the expected content. But the `pom.xml` is missing the signature. Another side effect is that the `1.0.1-SNAPSHOT` version of the modules above is not created anymore as the `signing` block is now added to the module.

In 1.0.0 more modules have been relocated. They need to be added as well when it is clear how this should be approached.

## Checklist

- [ ] Commit message are well written. In addition to a short title, the commit message also explain why a change is made.
- [ ] Tests are added
- [ ] PR description added


paul-dingemans:
@Goooler Can you please help me out with this change? I have sent a slack message a while ago with more details.

ktlint-cli-reporter-baseline/build.gradle.kts
```diff
@@ -9,3 +9,26 @@ dependencies {
 
     testImplementation(projects.ktlintTest)
 }
+
+publishing {
+    publications {
+        // TODO: Remove in release after ktlint 1.0.1
+        create<MavenPublication>("relocation-ktlint-cli-reporter-baseline") {
```
Goooler, line 16: Not be familiar with maven relocation, but if you wan't to sign this, could add something like:

```kotliln
signing {
    sign(publishing.publications["relocation-ktlint-cli-reporter-baseline"])
}
```

Ref: https://github.com/pinterest/ktlint/blob/ea8df7e85abb85d14deff59dd056357bb5700feb/build-logic/src/main/kotlin/ktlint-publication.gradle.kts#L90

Goooler:
And if this is only for maven coordinate relocation, can we rebase this PR to `0.51.0-FINAL` branch and publish it, instead of on the master.

ktlint-cli-reporter-baseline/build.gradle.kts
paul-dingemans: My question is not specifically about the maven relocation, but how to avoid code duplication in Gradle build. I have now added a `publication` and `signing` block in each module for which the maven coordinate has been changed. It is good enough as it will be removed soon.

paul-dingemans:
> And if this is only for maven coordinate relocation, can we rebase this PR to `0.51.0-FINAL` branch and publish it, instead of on the master.

I don't know how to create a release from a branch. I might not have enough privileges for that. I have tested on my local machine that publish to maven local creates both artifacts for `0.51.0-FINAL` and `1.0.1`. Only downside is that I have to remove the changes of this PR after publication of the next release.

PR #2377, opened 2023-11-25
Fix searching from inside a hidden directory

kitterion:
## Description

When searching files like this:
```
mkdir .hidden_folder
cd .hidden_folder
touch File.kt
ktlint '*.kt'
```
ktlint reported that it found no files.
The fact that the current directory is hidden should have no effect.

## Checklist

- [x] Tests are added

### DIFF
--- a/ktlint-cli/src/main/kotlin/com/pinterest/ktlint/cli/internal/FileUtils.kt
+++ b/ktlint-cli/src/main/kotlin/com/pinterest/ktlint/cli/internal/FileUtils.kt
@@ -136,9 +136,14 @@ internal fun FileSystem.fileSequence(
                         dirPath: Path,
                         dirAttr: BasicFileAttributes,
                     ): FileVisitResult =
-                        if (Files.isHidden(dirPath) && dirPath != commonRootDir) {
-                            LOGGER.trace { "- Dir: $dirPath: Ignore" }
-                            FileVisitResult.SKIP_SUBTREE
+                        if (Files.isHidden(dirPath)) {
+                            if (dirPath == commonRootDir) {
+                                LOGGER.trace { "- Dir: $dirPath: Traverse started from hidden directory" }
+                                FileVisitResult.CONTINUE
+                            } else {
+                                LOGGER.trace { "- Dir: $dirPath: Ignore traversal of hidden directory" }
+                                FileVisitResult.SKIP_SUBTREE
+                            }
                         } else {
                             LOGGER.trace { "- Dir: $dirPath: Traverse" }
                             FileVisitResult.CONTINUE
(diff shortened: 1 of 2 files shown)


paul-dingemans:
Tnx for your contribution. Unfortunately, I have to reject it in its current form.

What are your reasons to use hidden folders and what is your need to scan them?

Hidden directories are excluded on purpose. I think, the original reason was to exclude those directories as they are used by version control systems like git. Scanning files in `.git` folders is for example unwanted for two reasons: 1) this folder can be massively large (e.g. deeply nested); 2) files in this directory should not be manipulated manually.

Of course the current exclusion of *all* hidden folders, might lead to exclusion of folders for which you do want to enable scanning. The default behavior of excluding hidden folders may not be changed. 

Have you tried to specify the path to your hidden folder as glob? Something like below, does also scan the hidden folders for me:
```
ktlint **/*.kt **/*.kts src/main/kotlin/.hidden_folder/**
```

kitterion:
> Hidden directories are excluded on purpose. I think, the original reason was to exclude those directories as they are used by version control systems like git. Scanning files in `.git` folders is for example unwanted for two reasons: 1) this folder can be massively large (e.g. deeply nested); 2) files in this directory should not be manipulated manually.

Understandable, nothing changes in those cases. The hidden folders _under_ the root folder continue to be skipped. All the existing tests pass as well, including those that test that ktlint does not enter hidden folders. There is a side-effect of the current logic, however: if the root folder _itself_ is hidden from the point of view of its parent then no searching will be done at all.
In other words, entering the hidden folder (`cd .hidden_folder` in the example above) and searching within that folder should work but doesn't. I feel like if I enter a hidden folder then I have expressed intent that I want to search within that folder.

> What are your reasons to use hidden folders and what is your need to scan them?

I have a tool that launches subcommands from within a hidden folder, and unfortunately I have little control over it. One of the subcommands is ktlint, and ktlint refuses to search any files, no matter how I lay them out on the filesystem.

> Something like below, does also scan the hidden folders for me:
> ```
>  ktlint **/*.kt **/*.kts src/main/kotlin/.hidden_folder/**
> ```

In this example as written, the shell expands `**`, not ktlint. In this case ktlint ignores the fact that files might be in a hidden directory, it just uses the full filepaths, because that's the only thing it sees. It doesn't traverse the filesystem because it doesn't need to.
If I change that example slightly by adding quotes
```bash
ktlint '**/*.kt' '**/*.kts' 'src/main/kotlin/.hidden_folder/**'
```
then ktlint stops searching files under 'src/main/kotlin/.hidden_folder'.
What I'm fixing is (note the quotes here as well)
```bash
cd src/main/kotlin/.hidden_folder
ktlint '**'
```

paul-dingemans:
Tnx for detailed explanation! Now I understand what the change is about. I have added a commit to add a log message and made the unit less confusing.

kitterion:
No worries, thanks!

PR #2421, opened 2023-12-10
Make ktlint.bat more environment agnostic

TWiStErRob:
## Description

I was trying to use `ktlint.bat` and ran into multiple problems. These few minor changes solve these.
I'll comment on each why it was changed as conversation.

## Checklist

- [ ] Tests are added

- [x] [Snapshot documentation](https://github.com/pinterest/ktlint/tree/master/documentation/snapshot) in case documentation is to be released together with a code change
- [x] [Release documentation](https://github.com/pinterest/ktlint/tree/master/documentation/release-latest) in case documentation is related to a released version of ktlint and has to be published as soon as the change is merged to master

### DIFF
--- a/ktlint-cli/src/main/scripts/ktlint.bat
+++ b/ktlint-cli/src/main/scripts/ktlint.bat
@@ -1,8 +1,10 @@
 @echo off
+setlocal
+
 REM By default this batch file and the "ktlint" are located in same directory. Note that "ktlint" is a JAR
 REM despite it is not being suffixed with extension ".jar". Please adjust path below when this batch file
 REM and the jar file are located in different directories.
-if not defined JAR_PATH set JAR_PATH=.\ktlint
+if not defined JAR_PATH set JAR_PATH=%~dp0ktlint
 
 REM The --add-opens argument is needed for java 16+ (see https://github.com/pinterest/ktlint/issues/1986)
 java --add-opens=java.base/java.lang=ALL-UNNAMED -jar "%JAR_PATH%" %*


ktlint-cli/src/main/scripts/ktlint.bat
```diff
@@ -1,8 +1,10 @@
 @echo off
+setlocal
```
TWiStErRob, line 2: `set` command is equivalent to `export` in Unix, meaning it'll change the surrounding environment. Executing `setlocal` will ensure that `JAR_PATH` is only a local variable and it only affects the single command after it's set.

ktlint-cli/src/main/scripts/ktlint.bat
```diff
@@ -1,8 +1,10 @@
...
+
 REM By default this batch file and the "ktlint" are located in same directory. Note that "ktlint" is a JAR
 REM despite it is not being suffixed with extension ".jar". Please adjust path below when this batch file
 REM and the jar file are located in different directories.
-if not defined JAR_PATH set JAR_PATH=.\ktlint
+if not defined JAR_PATH set JAR_PATH=%~dp0ktlint
```
TWiStErRob, line 7: The `ktlint.bat` and `ktlint(.jar)` files are meant to be in the same folder, right?

`.` refers to CWD, but one might call this bat file from another directory where there's no `ktlint` file. `%~dp0` refers to the "Folder containing this .bat file" to ensure that it can be called from anywhere.

`%0` is the bat file, `dp` is drive and path and `~` removes auto-quotes. see https://stackoverflow.com/a/5034119/253468

TWiStErRob, line 7: Theoretically one would write `JAR_PATH="%~dp0ktlint"` to make sure that spaces are handled, but the quotes will become part of the `JAR_PATH` variable, and therefor can break future usages. The expansion is space-safe, and `JAR_PATH` can contain spaces. I tested it.

paul-dingemans: > The `ktlint.bat` and `ktlint(.jar)` files are meant to be in the same folder, right?

Not necessarily. As the comment explains, the user can choose to get/store the ktlint.jar in some other location. For example when you use a package manager like `brew`, it will install the ktlint jar at a specific location. The user might want to have the bat file to be stored in another place.

> `.` refers to CWD, but one might call this bat file from another directory where there's no `ktlint` file. `%~dp0` refers to the "Folder containing this .bat file" to ensure that it can be called from anywhere.
> 
> `%0` is the bat file, `dp` is drive and path and `~` removes auto-quotes. see https://stackoverflow.com/a/5034119/253468

Tnx for explaining. My windows knowledge is rusty nowadays.

TWiStErRob: Yep, I downloaded the files from releases, put them in my "tools" (on PATH) and called it from my project, as I do with many other CLIs.

For static relocation where bat and jar needs to be separate the JAR_PATH should be used as I understood. Btw, this PR might worth a line in the release notes. Because if someone has ktlint(.jar) in each of their projects and the bat in a central location, this change might break them.

I just realised I could've checked how ktlint.sh does, but it's embedded inside the jar, so they're always in the "same folder" 😅.

Thanks for the quick turnaround, you're amazing keeping this repo live! I was wondering how you keep the issue count so low, now I see!

> My windows knowledge is rusty nowadays.

Feel free to tag me if you need something like this change tested on Windows any time!

paul-dingemans: > Btw, this PR might worth a line in the release notes.

Title of PR will be mentioned in release notes with reference to it.

> Thanks for the quick turnaround, you're amazing keeping this repo live! I was wondering how you keep the issue count so low, now I see!

Tnx, this is appreciated.

> Feel free to tag me if you need something like this change tested on Windows any time!

Tnx again. I will certainly do so.

PR #2436, opened 2023-12-14
Improve generateEditorConfig example in docs

BD103:
## Description

I deleted a duplicate command in the documentation. (`ktlint generateEditorConfig`) I don't believe it was intentional.

I figured the fix was simple enough that adding an extended description to the commit message is not necessary. I can change this, though.

I added the fix to both the snapshot and release documentations. It's not necessarily pressing enough to need being updated in the release version, but I figured it couldn't do too much harm.

## Checklist

- [ ] Tests are added

- [x] [Snapshot documentation](https://github.com/pinterest/ktlint/tree/master/documentation/snapshot) in case documentation is to be released together with a code change
- [x] [Release documentation](https://github.com/pinterest/ktlint/tree/master/documentation/release-latest) in case documentation is related to a released version of ktlint and has to be published as soon as the change is merged to master

### DIFF
--- a/documentation/release-latest/docs/install/cli.md
+++ b/documentation/release-latest/docs/install/cli.md
@@ -141,9 +141,12 @@ Some rules can be tweaked via the [`editorconfig file`](../../rules/configuratio
 A scaffold of the `.editorconfig file` can be generated with command below. Note: that the generated file only contains configuration settings which are actively used by the [rules which are loaded](#rule-sets):
 
 ```shell title="Generate .editorconfig"
+# By default use ktlint_official
 ktlint generateEditorConfig
 # or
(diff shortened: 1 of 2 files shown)


paul-dingemans:
The documentation is indeed not correct. It would be better to change it something like below:
```
ktlint generateEditorConfig
# or
ktlint generateEditorConfig intellij_idea
# or
ktlint --ruleset=/path/to/custom-ruleset.jar generateEditorConfig  android_studio```

The idea was that to demonstrate that if no code style is provided that `ktlint_official` is used as default. If you want another code style then you need to provide the id of that code style.

BD103:
Ok, I've updated the example. :)

PR #2466, opened 2023-12-26
Run `argument-list-wrapping`, `class-signature` and `function-signature` when comment rules are disabled

### LINKED ISSUE #2445: ktlint 1.1.0 doesn't respect disabled discouraged-comment-location rule in .editorconfig
We don't want to enforce comments location in our project because we like the following code style:

```
        val object = Object(
            someParam1 = 10000, // comment here about why
            someParam2 = 100, // comment here about why
            sameParam3 = emptyList(), // comment here about why
        )
```
It's compact and nice to work with, but discouraged by ktlint.

To disable the `A comment in a 'value_argument_list' is only allowed when placed on a separate line (cannot be auto-corrected)` warnings for such usages, our `.editorconfig` file looks the following way:
```
root = true

[*]
ktlint_code_style = intellij_idea
ktlint_standard_discouraged-comment-location = disabled
```

It works for ktlint 1.0.1, but ignored when we try to update ktlint to 1.1.0.

### Additional context

We are using org.jlleitschuh.gradle.ktlint:12.0.3 gradle plugin configured the following way:
```
    ktlint {
        version.set(libs.versions.ktlint)
        coloredOutput.set(true)
        android.set(false)
        ignoreFailures.set(false)
        filter {
            exclude("**/generated/**")
            exclude { element -> element.file.path.contains("generated/") }
        }
    }
```

paul-dingemans:
## Description

Loosen dependencies on `value-argument-comment`, `value-parameter-comment`, `type-argument-comment`, and `type-parameter-comment` rules to be disabled

Run `argument-list-wrapping`, `class-signature` and `function-signature` when comment rules above are disabled.

Closes #2445

## Checklist

- [X] Tests are added

- [X] [Release documentation](https://github.com/pinterest/ktlint/tree/master/documentation/release-latest) in case documentation is related to a released version of ktlint and has to be published as soon as the change is merged to master

### DIFF
--- a/documentation/snapshot/docs/rules/standard.md
+++ b/documentation/snapshot/docs/rules/standard.md
@@ -2041,14 +2041,15 @@ Disallows comments to be placed at certain locations inside a type argument (lis
         >
     ```
 
-Note: Although code sample below might look ok, it is semantically and programmatically unclear to which element `some comment 1` refers. As of that comments are only allowed when starting on a separate line. 
-```kotlin
-fun Foo<
-    out Bar1, // some comment 1
-    // some comment 2
-    out Bar2, // some comment
+!!! note
+    In some projects it is an accepted practice to use EOL comments to document the parameter *before* the comma as is shown below:
+    ```kotlin
+    fun Foo<
+        out Bar1, // some comment
+        out Bar2, // some other comment
     >.foo() {}
-```
+    ```
+    Although this code sample might look ok, it is semantically and programmatically unclear to which type `some comment` refers. From the developer perspective it might be clear that it belongs to type `Bar1`. From the parsers perspective, it does belong to type `Bar2`.
 
 Rule id: `type-argument-comment` (`standard` rule set)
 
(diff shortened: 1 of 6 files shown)


vanniktech:
>   - Rule with id 'RuleId(value=standard:argument-list-wrapping)' requires rule with id 'RuleId(value=standard:value-argument-comment)' to be loaded

is fixed by this when running ktlint 1.1.0, right? I was trying to update and yet again I was being hit by Rules .. requires Rule .. to be loaded.

I can only agree with @eygraber's [comment](https://github.com/pinterest/ktlint/issues/2445#issuecomment-1863323069)

paul-dingemans:
> is fixed by this when running ktlint 1.1.0, right?

No. As you can see, this issue is labeled with milestone `1.1.1` which is not yet released. You can use the [snapshot](https://pinterest.github.io/ktlint/latest/install/snapshot-build/) if you like.

vanniktech:
Ah sorry I meant to say that I had gotten that exception when running 1.1.0 and is fixed with this PR (since this PR sits on top of the 1.1.0 version).

Issue #2532, opened 2024-02-04
Cannot have working filename-based rules and avoid .editorconfig loading at the same time.


mgroth0:
## Expected Behavior

To be able to process code snippets with a virtual file path with the following two behaviors:
1. To not check the filesystem for .editorconfig files
2. To have rules that depend on the file path such as `standard:filename` work correctly

## Observed Behavior

- If we use `Code.fromSnippetWithPath`, the `standard: filename` rule works, but editorconfig files are checked.
- If we use `Code.fromSnippet`, `.editorconfig` files are ignored but rules such as `standard:filename` will not work.

## Context

- I lint kotlin files using the ktlint API. I want to use rules such as `standard:filename` that depend on the filepath, but I do not want the filesystem to be checked for .editorconfig files. I do use a .editorconfig file, but I handle the IO with it myself and provide it to the engine as a specially constructed `EditorConfigDefaults`.

## Suggested Solution

`RuleExecutionContext.createRuleExecutionContext` is hardcoded such that `Code.filePath` is used both for providing the filePath to the psiFile for use in rules (`PsiFileFactory.createFileFromText`) and for loading .editorconfig files. (`EditorConfigLoader.load`). Somehow this should be refactored so that API users may provide file paths to the psi file and/or provide file paths for .editorconfig loading separately. They should not be required to use both features in an all-or-nothing way since they are not interdependent.

paul-dingemans:
> I lint kotlin files using the ktlint API. I want to use rules such as standard:filename that depend on the filepath, but I do not want the filesystem to be checked for .editorconfig files. I do use a .editorconfig file, but I handle the IO with it myself and provide it to the engine as a specially constructed EditorConfigDefaults.

Out of curisosity, why do you to handle the IO regarding `.editorconfig` yourself? One of the key features of Ktlint is that it does respect the `.editorconfig` when it does find it on the path. The `.editorconfig` standard offer more functionality than it seems to do at first glance. Implementing this yourselves might lead to deviations from that standard. Of course that is not problem when you are the only user of your API. But if you would expose this API to others, it can lead to confusion, and in the end to complaints about Ktlint not working correctly.

Instead of using `EditorConfigDefaults` it might be better to the `EditorConfigOverride`. The latter takes precedence on settings found in real `.editorconfig` file(s), while the `EditorConfigDefaults` only is used whenever a setting is not found in the real `.editorconfig` file(s).

paul-dingemans:
[...]

PR #2537, opened 2024-02-05
Allow making requests to engine with raw .editorconfig string, disabling .editorconfig IO.

mgroth0:
## Description

Fixes https://github.com/pinterest/ktlint/issues/2532

Allows the API user to make requests to the engine with a raw .editorconfig string in like so:

```kotlin
Code.fromSnippet(
                content = "kotlinCode",
                editorConfig = "[*]\nktlint_simple-test_a = disabled",
)
// or
Code.fromSnippetWithPath(
                content = "kotlinCode",
                virtualPath = "/virtual/path.kt", // no .editorconfig IO, but used in `standard:filename`
                editorConfig = "[*]\nktlint_simple-test_a = disabled",
)
```

This should allow the API user to completely avoid the engine from doing any IO at all in relation to .editorconfig files, providing more flexibility possible(not measured) performance gains, and stability.

Added one simple test to verify that the .editorconfig string is correctly applied. However, more tests and refining the current test would be good.

Did not modify or add any documentation yet.

This is a rough draft. It might be controversial, so I kept it minimal just as an example. If it looks good as is, it just needs better tests and documentation. But I expect reviewers might have opinions on how this should be implemented in a way that enhances flexibility while minimizing confusion and complexity for API users.

Note that the `virtualPath` for `fromSnippetWithPath` is still useful and sometimes neccesary, as it gets used in rules that use the file path such as `standard:filename`.

## Checklist

- [ ] Commit message are well written. In addition to a short title, the commit message also explain why a change is made.
- [ ] Tests are added
- [ ] PR title is short and clear (it is used as description in the release changelog)

### DIFF
--- a/ktlint-rule-engine/api/ktlint-rule-engine.api
+++ b/ktlint-rule-engine/api/ktlint-rule-engine.api
@@ -1,9 +1,10 @@
 public final class com/pinterest/ktlint/rule/engine/api/Code {
 	public static final field Companion Lcom/pinterest/ktlint/rule/engine/api/Code$Companion;
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/nio/file/Path;ZZLkotlin/jvm/internal/DefaultConstructorMarker;)V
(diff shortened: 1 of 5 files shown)


paul-dingemans:
As documented in #2532 this feature will not be integrated in ktlint. The idea behind this request seems valid, but the targeted user group is way too small for this moment. Next to that, the proposed change causes multiple breaking changes for the existing integrators.

Issue #2675, opened 2024-05-28
operator get with lambda: incorrect error


Jolanrensen:
## Expected Behavior

Given:
```kt
class DB {
    operator fun <T> get(query: DB.() -> T): T = TODO()
}

val db = DB()
val query = db[{ 123 }]
```
so, an operator get function with a lambda argument inside, KtLint should be happy with this notation. It is accepted by IntelliJ.

## Observed Behavior
We get a linting error: 
`Lint error > [standard:curly-spacing] Missing spacing before "{"`.

If I run the formatter instead it performs a fix and reverts it again:
`Format fixed > [standard:curly-spacing] Missing spacing before "{"`

`Format fixed > [standard:square-brackets-spacing] Unexpected spacing after '['`

## Steps to Reproduce
See the example above.

## Your Environment
- KtLint 1.2.1
- Using https://github.com/jeremymailen/kotlinter-gradle/tree/master 4.3.0
- `ktlint_code_style = intellij_idea`

PR #2677, opened 2024-05-29
Fixed [{ }] notation for issue #2675

Jolanrensen:
Fixes https://github.com/pinterest/ktlint/issues/2675

## Description

Fixes the erroneous error for notations like `[{ }]`, as described in the issue.
The `}]` case was already handled by https://github.com/pinterest/ktlint/pull/596, but the `[{` notation needed some additional fix. There was a special exception for `({`, and since this one is similar, I added the case for LBRACKET too.

Two tests were added, one to check that notations like `[{ a -> a }]`, `[{ }]`, and `[{}]` are allowed, and a second one to check that `[ { } ]` is formatted correctly.


paul-dingemans:
Tnx for reporting and fixing. I made some small changes from consistency viewpoint.

PR #2758, opened 2024-08-01
fix: don't remove arrow from lambdas that are when/if leaf nodes

tKe:
## Description

this fixes a problem with the `function-literal` rule removing non-redundant arrows from lambdas that are direct leafs of a when-entry or then/else. Without the arrow, the expression is left as a block instead.

## Checklist

- [x] At least one commit message contains a reference `Closes #<xxx>` or `Fixes #<xxx>` (replace`<xxx>` with issue number)
- [x] Tests are added


paul-dingemans:
Tnx for your contribution. I refactored the code a bit, so that it complies more with current coding styles.

PR #2796, opened 2024-09-14
failing test: chain-method-continuation fail for fully qualified call

bcmedeiros:
Just found an issue with `chain-method-continuation`, raising a PR with the repro only for now.

Happy to work on the fix if someone points me in the right direction.

Repro for https://github.com/pinterest/ktlint/issues/2797


paul-dingemans:
Please see [docs](https://pinterest.github.io/ktlint/latest/rules/standard/#chain-method-continuation) for configuration of the rule. Expression will be wrapped when the configured number of operators is reached. In case you do not want that, please set it to some high number (`Integer.MAX` for example).

So there is no need for fixing this.

bcmedeiros:
https://github.com/pinterest/ktlint/commit/9a10ec09f1cf6d58d5c70b03e57a1b69cd75191f supersedes this.

PR #2911, opened 2024-12-24
Pass additional information for logger when formatting snippets

mgroth0:
## Description

I generate multiple kotlin files in a large gradle build with custom gradle tasks. The final step of the kotlin code generation uses ktlint to format the code *before* it is written to a file. As such, the code is passed to ktlint using `Code.fromSnippet`. This normally works, but occasionally I get log warning messages such as:

```
Format was not able to resolve all violations which (theoretically) can be autocorrected in file <stdin>
```

The issue is that I cannot determine which one of the generated files is causing this issue. This PR adds an optional parameter to allow passing more information to this log message.

## Checklist

- [ ] Tests are added

### DIFF
--- a/ktlint-rule-engine/api/ktlint-rule-engine.api
+++ b/ktlint-rule-engine/api/ktlint-rule-engine.api
@@ -2,6 +2,7 @@ public final class com/pinterest/ktlint/rule/engine/api/Code {
 	public static final field Companion Lcom/pinterest/ktlint/rule/engine/api/Code$Companion;
 	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/nio/file/Path;ZZLkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun fileNameOrStdin ()Ljava/lang/String;
+	public final fun filePathOrFileNameOrStdin ()Ljava/lang/String;
 	public final fun filePathOrStdin ()Ljava/lang/String;
 	public final fun getContent ()Ljava/lang/String;
 	public final fun getFileName ()Ljava/lang/String;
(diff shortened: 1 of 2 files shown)


paul-dingemans:
This solution is not acceptable to me for following reasons:
* The `fromSnippet` method signature is changed. Although the default parameter prevents backward compatibility issue for Kotlin projects, it will still break Java integrations like Spotless.
* Why can you not use `fromSnippetWithPath` instead?

Note that any of your solution will still be limited by the built-in behavior regarding rerunning ktlint multiple times on the same file. The maximum is limited to three attempts (hard coded). This is  a limit that works most of the time. But, for code bases on which ktlint never has run, this might not be enough. The maximum number of runs is limited, as it otherwise can happen that formatting ends in an endless loop when rules are reverting each other changes. I would not be against a Handler that is implemented by the KtLintRuleEngine that gives the API Consumer control about the retry runs. But, the API Consumer should be warned about endless loop behavior when overriding the default behavior.

paul-dingemans:
> This would trigger filesystem IO as it searches for .editorconfig files if I understand correctly.

I see your use case. But there is no way to avoid filesystem IO. The filepath is extracted from the `Code` object on line https://github.com/pinterest/ktlint/blob/dc4e2fb48d0a7dcd7f792db7da58061c4946c328/ktlint-rule-engine/src/main/kotlin/com/pinterest/ktlint/rule/engine/internal/RuleExecutionContext.kt#L218. If no path information is provided in the `Code` object, the path `File.kt` or `File.kts` is used. On line https://github.com/pinterest/ktlint/blob/dc4e2fb48d0a7dcd7f792db7da58061c4946c328/ktlint-rule-engine/src/main/kotlin/com/pinterest/ktlint/rule/engine/internal/RuleExecutionContext.kt#L226 a temporary file is created as this is needed for the embeded Kotlin compiler. On line https://github.com/pinterest/ktlint/blob/dc4e2fb48d0a7dcd7f792db7da58061c4946c328/ktlint-rule-engine/src/main/kotlin/com/pinterest/ktlint/rule/engine/internal/RuleExecutionContext.kt#L243 the `.editorconfig` file is being located for the (dummy) file represented by the `Code` object. Note that the EditorConfig is heavily cached. So for all your code snippets that you format via the same instance of `KtLintRuleEngine` the search for path `File.kt` / `File.kts` happens exactly once.

For your use case, I would say it is best to ignore the warning below entirely:
```
Format was not able to resolve all violations which (theoretically) can be autocorrected in file <stdin>
```
Especially, because you will not be able to guarrantee that you can resolve the problem by rerunning format yet another time.

PR #3025, opened 2025-06-28
Improve build

mateuszkwiecinski:
## Description

As discussed on Slack - I'm proposing a few changes to the build files. Each are in separate commit, I can drop any of them if they are not considered an improvement 😅 

## Checklist

- [ ] Tests are added

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -23,5 +23,6 @@ org.gradle.caching=true
 org.gradle.configuration-cache=true
 org.gradle.configuration-cache.parallel=true
 
-# Enable helpers for migration to Dokka V2
-org.jetbrains.dokka.experimental.gradle.pluginMode=V2EnabledWithHelpers
+# Use V2 of the Dokka library and silence its warnings
+org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
+org.jetbrains.dokka.experimental.gradle.pluginMode.noWarn=true


ktlint-cli/build.gradle.kts
```diff
@@ -51,37 +51,36 @@ val shadowJarExecutable by tasks.registering(DefaultTask::class) {
 
     dependsOn(tasks.shadowJar)
 
-    val isReleaseBuild = !version.toString().endsWith("SNAPSHOT")
```
mateuszkwiecinski, line 54: ❗ 
I couldn't find a reason why snapshot would not need to have the executable files written 👀 Please let me know if I'm missing something. 
With the new config:
- executable files will always be created
- signature files will be created only if signing is enabled (driven by [this](https://github.com/pinterest/ktlint/blob/013b8a66506d1d278205da72e4ee925fdcb67999/build-logic/src/main/kotlin/ktlint-publication.gradle.kts#L105) logic)

ktlint-cli/build.gradle.kts
```diff
@@ -107,18 +106,9 @@ val shadowJarExecutable by tasks.registering(DefaultTask::class) {
             appendBytes(ktlintCliAllJarFile.readBytes())
 
             setExecutable(true, false)
-
-            if (isReleaseBuild) {
-                logger.lifecycle("Creating the signature file: $selfExecutableKtlintSignatureOutputPath")
-                signing.sign(this)
```
mateuszkwiecinski, line 113: This was reason why configuration cache had been failing. Since `signing.sign()` is just a config, it doesn't have to be within `doLast` block.

PR #1952, opened 2023-04-18
Exclude asc from list of signed files

shashachu:
## Description

Recreating PR #1951 with the version change merged in.

PR #3047, opened 2025-07-06
Make `build-logic` plugins expose binary plugins

mateuszkwiecinski:
## Description

Latest development has been facing challenges due to Kotlin version enforced by [kotlin-dsl](https://github.com/pinterest/ktlint/blob/2a3c922ba6b9b34092f7782e09e779530878b8b3/build-logic/build.gradle.kts#L4) plugin. 
This PR migrates build-logic to leverage binary plugins so the build-logic project is no longer affected by kotlin-dsl plugin restrictions. 

## Checklist

Before submitting the PR, please check following (checks which are not relevant may be ignored):
- [x] Commit message are well written. In addition to a short title, the commit message also explain why a change is made.
- [ ] At least one commit message contains a reference `Closes #<xxx>` or `Fixes #<xxx>` (replace`<xxx>` with issue number)
- [ ] Tests are added
- [x] KtLint format has been applied on source code itself and violations are fixed
- [x] PR title is short and clear (it is used as description in the release changelog)
- [x] PR description added (background information)

[Documentation](https://pinterest.github.io/ktlint/) is updated. See [difference between snapshot and release documentation](https://github.com/pinterest/ktlint/tree/master/documentation)
- [ ] [Snapshot documentation](https://github.com/pinterest/ktlint/tree/master/documentation/snapshot) in case documentation is to be released together with a code change
- [ ] [Release documentation](https://github.com/pinterest/ktlint/tree/master/documentation/release-latest) in case documentation is related to a released version of ktlint and has to be published as soon as the change is merged to master

PR #3058, opened 2025-07-14
Update signing config, restore MD5 checksums, fix configuration-cache compatibility

mateuszkwiecinski:
## Description

This PR introduces the following set of behavior changes:
- signing config has been extracted from `shadowJarExecutable` task into `shadowJarExecutableChecksum` to make it compatible with configuration cache
    - running `shadowJarExecutable` automatically runs `shadowJarExecutableChecksum` so the user-facing behavior is the same. 
- Running `shadowJarExecutable` will also generate md5 checksums (disabled by https://github.com/pinterest/ktlint/pull/1952/files#diff-10bb684577395d2b98c65f30064e347f353307c7c4956b0a147b26e3085f3bbfL88).
   - I couldn't find any usages of `shadowJarExecutableChecksum` task, which directly addressed: https://github.com/pinterest/ktlint/issues/695 . Please let me know if checksums aren't meant to be published I'll then propose to remove all the checksum related code. 

|Before|After|
|---|---|
|image|image|

- Tasks expose very explicit inputs&outputs rather than passing parent directory reference around. That way running task multuple times (with `--rerun` flag results in a deterministic behavior)

## Checklist

- [ ] Tests are added


paul-dingemans:
> * Please let me know if checksums aren't meant to be published I'll then propose to remove all the checksum related code.

Looking at the commits in the PR #1952 it is clear that the change was initiated by @Goooler. At April 20, 2023 @shashachu reverted the signing, followed by the release of version [0.49.0](https://github.com/pinterest/ktlint/releases/tag/0.49.0) on April 21, 2023. So it looks like that during release of `0.49.0` a blocking problem occurred, which was resolved by partially reverting the signing. To me this sounds that the signing was accidentally removed to unblock the release.

Having said that, we can argue about the value of having the md5 files being published. No one has complained about a regression since the md5 files are missing.

Does any of you have an argument in favor, or against re-adding the md5 functionality?

mateuszkwiecinski:
Artifacts available on Maven central have their own authenticity certification: [signature](https://repo1.maven.org/maven2/com/pinterest/ktlint/ktlint-cli/1.7.0/ktlint-cli-1.7.0-all.jar.asc), [hash](https://repo1.maven.org/maven2/com/pinterest/ktlint/ktlint-cli/1.7.0/ktlint-cli-1.7.0-all.jar.asc.md5), so from my perspective those hashes are useful only for people downloading artifacts directly from github release page (does that include homebrew? I have close to zero knowledge how homebrew works 😬 )

At the same time nothing prevents an attacker to swap both executable and hash file (releases can be edited, new artifact files can be uploaded), so I don't know if the checksum comes with any _safety_ guarantee 😅 

tl;dr I don't think these extra checksums bring any value, especialy they are gone for >2 years 😅

paul-dingemans:
> so from my perspective those hashes are useful only for people downloading artifacts directly from github release page (does that include homebrew? I have close to zero knowledge how homebrew works 😬 )

Github release pahe publishes the SHA256 for this:
Screenshot 2025-07-15 at 21 09 08

Also in [ktlint documentation](https://pinterest.github.io/ktlint/latest/install/cli/#download-using-curl) it is explained how to verify the downloaded file.

> tl;dr I don't think these extra checksums bring any value, especialy they are gone for >2 years 😅

Agree. @Goooler, @shashachu, what do you think?

mateuszkwiecinski:
>Github release pahe publishes the SHA256 for this:

:TIL:, looks like relatively new feature? None of the previous releases had such metadata attached 👀

build-logic/src/main/kotlin/PublicationPlugin.kt
```diff
@@ -134,10 +134,17 @@ abstract class PublicationPlugin : Plugin<Project> {
...
-                val signingPassword = System.getenv("ORG_GRADLE_PROJECT_signingKeyPassword")
-                useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
+                val signingKeyId = System.getenv("ORG_GRADLE_PROJECT_signingKeyId")?.takeIf { it.isNotEmpty() }
+                val signingKey = System.getenv("ORG_GRADLE_PROJECT_signingKey")?.takeIf { it.isNotEmpty() }
+                val signingPassword = System.getenv("ORG_GRADLE_PROJECT_signingKeyPassword")?.takeIf { it.isNotEmpty() }
+                if (signingKeyId != null && signingKey != null && signingPassword != null) {
+                    // Avoid setting empty strings as signing keys. This avoids breaking the build when PR is opened from a fork.
+                    // Also, due to https://github.com/gradle/gradle/issues/18477 signing tasks try to prematurely access the signatory.
+                    // This also improves error messages if something's misconfigured
+                    useInMemoryPgpKeys(signingKeyId, signingKey, signingPassword)
```
mateuszkwiecinski, line 144: The new task I added (`signShadowJarExecutable`) gets created eagerly, tries to access the signatory and fails if the signing info contains empty string (not nulls) 

As a workaround I'm proposing to not configure the signing. This solves the misbehavior of the Sign task but also should not fail PRs from froks created _during_ release (the `isRequired` property is not the only check that determines if the signing tasks should run - [link](https://github.com/gradle/gradle/blob/c5118d30e4d4996aa1612ff584e68469310f9a3c/platforms/software/signing/src/main/java/org/gradle/plugins/signing/Sign.java#L100))

mateuszkwiecinski:
The checksum calculating code has been removed in https://github.com/pinterest/ktlint/commit/9ec1ec6f2951856c13876b94071ae59c87a4829a ✅

Goooler:
I don't like this idea, I don't like #3047 too.

mateuszkwiecinski:
>I don't like this idea, I don't like #3047 too.

@Goooler You have to be more specific 😄 I fail to find how such feedback would be actionable 😅

build-logic/src/main/kotlin/ShadowJarExecutableTask.kt
```diff
@@ -0,0 +1,78 @@
...
+        val outputDirectory = objects.directoryProperty()
+
+        @OutputFile
+        val selfExecutable = outputDirectory.map { it.asFile.resolve("ktlint") }
+
+        @OutputFile
+        val windowsBatchScript = outputDirectory.map { it.asFile.resolve("ktlint.bat") }
+
+        init {
+            description = "Creates self-executable file, that runs generated shadow jar"
```
paul-dingemans, line 35: Description is not fully accurate. What about `Creates the ktlint CLI executable, and the Windows OS batch file to run the ktlint jar.`?

build-logic/src/main/kotlin/ShadowJarExecutableTask.kt
```diff
@@ -0,0 +1,78 @@
...
+            group = "Distribution"
+        }
+
+        @TaskAction
+        fun action() {
+            logger.lifecycle("ktlint-cli: Base jar to build self-executable file: ${allJarFile.get()}")
```
paul-dingemans, line 41: "Base jar" is confusing to me. Shouldn't this message be identical to the description in the `init` block? Or, alternatively: "Start ShadowJarExecutableTask"?

build-logic/src/main/kotlin/ShadowJarExecutableTask.kt
```diff
@@ -0,0 +1,78 @@
...
+
+            selfExecutable.get().apply {
+                logger.lifecycle("Creating the self-executable file: $this")
```
paul-dingemans, line 44: Change to "ktlint-cli: build self-executable file $this"

paul-dingemans (commented):
My understanding of this part of the build is increased a lot. I have a lot of naming things that might improve the code even more. Please have a look at the suggestions.

mateuszkwiecinski:
Thanks for clearing that up in https://github.com/pinterest/ktlint/pull/3078 🙏

Issue #3099, opened 2025-08-18
Add Rule Version Filtering Based on @SinceKtlint Annotations


hugoncosta:
# Enhancement: Add Rule Version Filtering Based on @SinceKtlint Annotations

## Expected Behavior

Add a new configuration option to KtLint plugin that allows users to filter out rules that were introduced after a specified KtLint version. This would use the existing `@SinceKtlint` annotations present on all rules to determine which rules to include.

**Proposed Configuration:**
```kotlin
ktlint {
    version.set("1.5.0")
    maxRuleVersion.set("0.46.0") // Only include rules introduced up to version 0.46.0
}
```

Why do I want this?
Upgrading to higher versions of Kotlin (e.g. 2.1.0) requires a newer version of Ktlint (we're currently on 1.2.1). If I upgrade it, I will cause failures due to the new rules not being respected, which I'd like to avoid. 

With the proposed feature, I would be able to upgrade to the latest version (e.g. 1.7.1) which _does_ have Kotlin 2.2 compatibility without having to introduce all the new rules between 1.2.1 and 1.7.1. This would then allow me to upgrade the rule set on my own time.

## Current Behavior

No filtering allowed unless I manually exclude new rules myself.

## Additional Information

* Current version of ktlint: all

paul-dingemans:
This feature request should be submitted to the KtLint Gradle project. That project is a wrapper around the KtLint project. I doubt it will work what you request. Your problem is related to the KtLintRuleEngine which is using a newer version of the embedded Kotlin compiler, and not on the rules themselves.

From ktlint perspective there is nothing that I can do to help you mitigating the problem.

hugoncosta:
Yes, there are currently Kotlin compatibility issues that can probably be addressed in the Ktlint Gradle plugin project. I would still very much like to have the ability to be running on the latest version of Ktlint but _only_ bring in the rules up to a particular version. 

Internally, we centrally configure Ktlint for all projects, and we cannot be bump the version without causing breakages across the company. This feature would allow us to set a baseline version (we upgrade it for every major Gradle version) and let the customers bump it manually if they want to be on the latest and greatest.

paul-dingemans:
I understand what you are aiming at. But the proposed configuration block would need to be supported by the KtLint-Gradle plugin.

[...]

PR #3101, opened 2025-08-18
Validate @SinceKtlint annotations with test and uniformize existing rules

hugoncosta:
## Description

Adds comprehensive validation for @SinceKtlint annotations and fixes multiple annotation issues to support the implementation discussed in #3099.

### Changes Made

1. Added a unit test to validate all rules follow the @SinceKtlint annotation usage as expected. Checks for missing annotations, proper experimental/stable status alignment, and version format compliance. For this, I had to add kotlin-reflect as a test implementation for proper reflection support.

2. Fixed Version Format Violations (8 rules)
Corrected patch-level versions to major.minor format as required:
• NoBlankLineBeforeRbraceRule: "0.10.2" → "0.10"
• WhenEntryBracing: "1.4.0" → "1.4"
• ConditionWrappingRule: "1.1.0" → "1.1"
• BlankLineBetweenWhenConditions: "1.2.0" → "1.2"
• ExpressionOperandWrappingRule: "1.7.0" → "1.7"
• KdocRule: "1.2.0" → "1.2"
• MixedConditionOperatorsRule: "1.1.0" → "1.1"
• NoSingleLineBlockCommentRule: "0.49.1" → "0.49"

3. Added Missing STABLE Annotations (3 rules)
Rules that were promoted from experimental to stable but missing STABLE annotations:
• ClassNamingRule: Added @SinceKtlint("1.0", STABLE)
• ConditionWrappingRule: Added @SinceKtlint("1.3", STABLE)
• FunctionTypeReferenceSpacingRule: Added @SinceKtlint("1.0", STABLE)

4. Updated @SinceKtlint annotation documentation to clarify version format requirements

## Checklist

• [x] Commit message are well written. In addition to a short title, the commit message also explain why a change is made.
• [x] At least one commit message contains a reference Closes #<xxx> or Fixes #<xxx> (replace<xxx> with issue number)
• [x] Tests are added
• [x] KtLint format has been applied on source code itself and violations are fixed
• [x] PR title is short and clear (it is used as description in the release changelog)
• [x] PR description added (background information)

[Documentation](https://pinterest.github.io/ktlint/) is updated. See [difference between snapshot and release documentation](https://github.com/pinterest/ktlint/tree/master/
documentation)
• [ ] [Snapshot documentation](https://github.com/pinterest/ktlint/tree/master/documentation/snapshot) in case documentation is to be released together with a code change
• [ ] [Release documentation](https://github.com/pinterest/ktlint/tree/master/documentation/release-latest) in case documentation is related to a released version of ktlint and has to be published as soon as the change is merged to master

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -16,6 +16,7 @@ kotlinx-binary-compatibiltiy-validator = "org.jetbrains.kotlinx.binary-compatibi
 kotlin-compiler = { module = "org.jetbrains.kotlin:kotlin-compiler-embeddable", version.ref = "kotlin" }
 kotlin-plugin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
 kotlin-plugin-dev = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlinDev" }
+kotlin-reflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlin" }
 clikt = "com.github.ajalt.clikt:clikt:5.0.3"
 dokka = "org.jetbrains.dokka:dokka-gradle-plugin:2.0.0"
 ec4j = "org.ec4j.core:ec4j-core:1.1.1"
--- a/ktlint-rule-engine-core/src/main/kotlin/com/pinterest/ktlint/rule/engine/core/api/SinceKtlint.kt
+++ b/ktlint-rule-engine-core/src/main/kotlin/com/pinterest/ktlint/rule/engine/core/api/SinceKtlint.kt
@@ -2,6 +2,9 @@ package com.pinterest.ktlint.rule.engine.core.api
 
 /**
  * The version in which the rule was introduced.
+ *
+ * @param version The version in "major.minor" format (e.g., "1.2"), never including patch level
+ * @param status The status of the rule in the specified version
  */
 @Repeatable
 @Target(
@@ -14,7 +17,7 @@ package com.pinterest.ktlint.rule.engine.core.api
     AnnotationTarget.PROPERTY_SETTER,
     AnnotationTarget.TYPEALIAS,
 )
-@Retention(AnnotationRetention.BINARY)
+@Retention(AnnotationRetention.RUNTIME)
 @MustBeDocumented
 public annotation class SinceKtlint(
     val version: String,
--- a/ktlint-ruleset-standard/build.gradle.kts
+++ b/ktlint-ruleset-standard/build.gradle.kts
@@ -9,6 +9,7 @@ dependencies {
     api(projects.ktlintRuleEngineCore)
 
     testImplementation(projects.ktlintTest)
+    testImplementation(libs.kotlin.reflect)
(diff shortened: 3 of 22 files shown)


paul-dingemans:
I am doubting whether the `@SinceKtlint` annotation should be exposed for runtime filtering of rules. When checking this PR, I found more inconsistencies:

For example, I found following rule:
```
@SinceKtlint("1.1.0", EXPERIMENTAL)
public class ConditionWrappingRule : StandardRule(id = "condition-wrapping") {
```

* The version numbers sometimes include the patch level. The intent was to exclude the patch level, as rule should only be introduced in major, and minor versions.
* The annotation might not be in sync with the interfaces implemented by the class. A rule is only experimental when the `Rule` class implements interface `Experimental`. So the `ConditionWrappingRule` above is either missing the `@SinceKtlint` annotation with the `STABLE` version, or it should have implemented the `Experimental` interface.

Using the annotation for filtering seems brittle. From ktlint perspective it is not discoverable, except for the retention parameter,  that the annotation might be used by API consumers. So there should at least be some unit tests that show how the annotation is (intended to be) used by the ktlint-gradle plugin. But even with this unit test in place, removal of the `@SinceKtlint` annotation on a specific rule class will not break any class, but it might be a breaking change for your use case.

hugoncosta:
I'm definitively not the happiest to take a dependency on an annotation, this means I have to use Reflection in ktlint-gradle to make it work. Plus, versions are string, easy to potential misplace a dot, makes the math harder. But so far, I'm not seeing a better way. 

On the ktlint-gradle side, I've added a few fault tolerance behaviours, like, if there are no annotations, we assume it's stable.

Would you consider a compromise of creating a unit test within Ktlint that, using reflection, ensures that all current and future rules have the expected annotations? We could also add a version validation to ensure that "1.1.0" would fail, prompting the contributor to use "1.1" instead.

hugoncosta:
Gave my idea above a try and updated the PR. There were indeed a few rules missing the expected annotation, so the test definitively helped me to fix all of that. Excuse any github shenanigans I may be doing, I have lost all my PR knowledge in favour of my company's equivalent 🙃

paul-dingemans:
> Gave my idea above a try and updated the PR. There were indeed a few rules missing the expected annotation, so the test definitively helped me to fix all of that. Excuse any github shenanigans I may be doing, I have lost all my PR knowledge in favour of my company's equivalent 🙃

I think this is fine. At least there are some checks in place now. It is still not fool proof, but it is good enough. Even when the change would not be accepted by the ktlint-gradle plugin, this is still valuable from ktlint perspective.

I have refactored your test by splitting it into multiple tests, and added an additional test to detect duplicate annotations. 

Unfortunately, I am not able to push to your branch directly. Can you please pull changes from https://github.com/pinterest/ktlint/tree/pr/3101-refactored in your branch?

hugoncosta:
Cherry picked and added to my branch, should be good?

paul-dingemans:
I will merge the changes. If https://github.com/JLLeitschuh/ktlint-gradle/issues/942 somehow is not accepted, I will revert the retention annotation on @SinceKtlint to compile again.

PR #3166, opened 2025-11-07
Allow `_` as backing property name (needed for KEEP-412)

serras:
## Description

This PR refines the standard "backing property naming" rule to allow using a single `_`. The reason for the change is that `_` is getting a new meaning under [KEEP-412](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0412-underscores-for-local-variables.md), which and used in the [unused return value checker](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0412-unused-return-value-checker.md) to identify values that a explicitly ignored.

In other words, without this change enabling the unused value checker gives you a lot of KtLint warnings about `_`, but this name cannot be changed due to the special behavior.

## Checklist

- [X] Tests are added


paul-dingemans:
Tnx for your contribution. I have slightly adjusted the implementation.

PR #3173, opened 2025-11-12
[Fix]: Give release workflow write permissions

shashachu:
## Description

PR #3115 changed all workflows to read-only, which is causing the release workflow to fail with a 403. This PR gives the workflow `content:write` permissions to allow it to create new releases.

### DIFF
--- a/.github/workflows/publish-release-build.yml
+++ b/.github/workflows/publish-release-build.yml
@@ -1,6 +1,6 @@
 name: Publish release build
 permissions:
-  contents: read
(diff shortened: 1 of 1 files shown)


paul-dingemans:
> PR https://github.com/pinterest/ktlint/pull/3115 changed all workflows to read-only, which is causing the release workflow to fail with a 403. This PR gives the workflow content:write permissions to allow it to create new releases.

Ah indeed, I forgot that I restricted all workflows.

PR #3177, opened 2025-11-14
Handle trailing comments on when entries in `BlankLineBetweenWhenConditions`

jonapoul:
## Description

I have a block of code in my project like below:

```kotlin
val x = when (something) {
    "a" -> 123 // Some comment
    "b" -> 456
    else -> 789
}
```

But this gave a linting problem with the new 1.8.0 update from the `standard:blank-line-between-when-conditions` rule. When I tried to auto-format to see how it should be, I got:

```kotlin
val x = when (something) {
    "a" -> 123
    
    // Some comment
    "b" -> 456

    else -> 789
}
```

The comment was placed on the wrong entry! Personally I think trailing comments should be allowable on simple when statements like this, interested to hear opinions on this though. The misformatting of the above example makes me think this is unintentional though.

To fix, I've added some more checks to `isPrecededByComment` to make sure the check for a multiline when entry handles this case.

No existing issue for this, so far as I can see.

## Checklist

- [x] Tests are added

### DIFF
--- a/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/BlankLineBetweenWhenConditions.kt
+++ b/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/rules/BlankLineBetweenWhenConditions.kt
@@ -99,7 +99,15 @@ public class BlankLineBetweenWhenConditions :
         children20
             .any { it.elementType == WHEN_ENTRY && (it.textContains('\n') || it.isPrecededByComment()) }
 
-    private fun ASTNode.isPrecededByComment() = prevSibling { !it.isWhiteSpace20 }?.isPartOfComment20 ?: false
+    private fun ASTNode.isPrecededByComment(): Boolean {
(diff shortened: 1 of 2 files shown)


paul-dingemans:
Tnx for reporting and fixing.

> The comment was placed on the wrong entry! Personally I think trailing comments should be allowable on simple when statements like this, interested to hear opinions on this though. The misformatting of the above example makes me think this is unintentional though.

EOL comments are always painful and causing problems. From the compiler perspective the EOL comment is at the same level as your literals "a" and "b". To match the expectation of the developer, rules quite often need additional logic to keep the EOL comment at the expected place. So this was definitely not the intended behavior. Unfortunately this was never reported while the rule was experimental (since Ktlint 1.2).

nuhkoca:
Hey @paul-dingemans is it possible to have 1.8.1 released near soon? As we are struggling from this issue a bit. Thanks!

paul-dingemans:
> Hey @paul-dingemans is it possible to have 1.8.1 released near soon? As we are struggling from this issue a bit. Thanks!

No, it will be released with next regular release (due in a couple of months). As mentioned before, this is a long outstanding problem:
> Unfortunately this was never reported while the rule was experimental (since Ktlint 1.2).

PR #3284, opened 2026-04-13
feat: add GraalVM native-image binaries to release (linux, macOS, Windows)

zeitlinger:
## Problem

ktlint currently ships a self-executing JAR (`ktlint`) and a Windows batch wrapper (`ktlint.bat`), both requiring a JVM to be present. This blocks tooling that wants to invoke ktlint without managing a JVM installation — such as editor plugins, CI containers, and language-agnostic formatting orchestrators.

## Solution

Add a `build-native-image` matrix job to the release workflow that produces self-contained native executables via GraalVM `native-image`. The job runs after the existing `publish` job (which creates the GitHub release) and uploads the binaries directly to the release.

**Platforms:**
- `ktlint_linux-x86-64` — built on `ubuntu-22.04` (older glibc for broader compatibility; see [google/google-java-format#1072](https://github.com/google/google-java-format/issues/1072))
- `ktlint_darwin-arm64` — built on `macos-latest`
- `ktlint_windows-x86-64.exe` — built on `windows-latest`

The naming convention mirrors what [google/google-java-format](https://github.com/google/google-java-format) already ships.

## Implementation details

- Uses `graalvm/setup-graalvm@v1` with GraalVM Community Edition, Java 21
- Builds the existing `ktlint-cli` shadow JAR via `./gradlew :ktlint-cli:shadowJar`, then invokes `native-image --no-fallback -H:+ReportExceptionStackTraces -march=compatibility` on it
- Adds `ktlint-cli/src/main/resources/META-INF/native-image/reachability-metadata.json` so GraalVM can find the ServiceLoader-registered reporter and ruleset provider classes at image-build time

## Notes

The reflection config covers all built-in reporter and ruleset providers. If ktlint uses additional reflection paths (e.g. through logback XML config or other dynamic dispatch), those may need to be added after observing a first native-image build run. The `native-image-job-reports: true` flag will surface any missing registrations in the CI job summary.


paul-dingemans:
Cool idea, never thought about adding this. Thanks for contributing.

Can you please add relevant documentation (if needed) about configuration from the user perspective? I will have a look at the PR later this, or next week.

documentation/snapshot/docs/install/cli.md
```diff
@@ -22,6 +26,22 @@ curl -sSLO https://github.com/ktlint/ktlint/releases/download/1.8.0/ktlint && ch
...
+```sh title="Download Linux native for x86-64"
+curl -sSLO https://github.com/ktlint/ktlint/releases/download/1.8.0/ktlint_linux-x86-64 && chmod a+x ktlint_linux-x86-64 && sudo mv ktlint_linux-x86-64 /usr/local/bin/ktlint
+```
+
+```sh title="Download Linux native for macOS Apple Silicon"
+curl -sSLO https://github.com/ktlint/ktlint/releases/download/1.8.0/ktlint_darwin-arm64 && chmod a+x ktlint_darwin-arm64 && sudo mv ktlint_linux-x86-64 /usr/local/bin/ktlint
+```
+
+```sh title="Download Linux native for macOS Apple Silicon"
+curl -sSLO https://github.com/ktlint/ktlint/releases/download/1.8.0/ktlint_windows-x86-64.exe && chmod a+x ktlint_windows-x86-64.exe && sudo mv ktlint_linux-x86-64 /usr/local/bin/ktlint
```
paul-dingemans: @zeitlinger It has been a long time ago that I used windows. Can you please check this windows command (e.g. the `chmod` and `sudo mv`)?

paul-dingemans:
Hi @zeitlinger,
Sorry for being unresponsive for a while. I have updated the PR:
* Repository has been moved to new Ktlint organization so coordinates have been changed.
* Documentation is slightly altered. See review comment to check validity of windows command.

Finally, I have one question about file `ktlint-cli/src/main/resources/META-INF/native-image/reachability-metadata.json`. How was the content of this file determined?

paul-dingemans:
The native-image builds are failing as `ktlint-api-consumer` requires Java 25 or newer (https://github.com/ktlint/ktlint/actions/runs/26504185379/job/78052252661?pr=3284#step:3:200). I am not sure how to resolve that. For what it is worth, this artifact is not needed for the native images as it just provides an example how an API consumen can integrate with Ktlint.

documentation/snapshot/docs/install/cli.md
```diff
@@ -15,10 +15,14 @@ Each release contains multiple ways to run `ktlint`:
...
 
-### Download using curl
+### Download using curl (Linux or macOS)
 
 A particular version of `ktlint` can be downloaded with the commands below. Each command uses `curl` for downloading. The downloaded file will be renamed to `/usr/local/bin/ktlint` and changed to an executable file.
 
+On macOS Apple Silicon, replace `ktlint_linux-x86-64` with `ktlint_darwin-arm64` in the command above.
+
+On Windows, download `ktlint_windows-x86-64.exe` from the same release page and add the directory containing it to your `%PATH%`.
+
```
paul-dingemans: It looks that you misunderstood my remark about changing the windows command. I rewrote you initial documentation to following, where I have separate commands per OS. The third one (title of the command should still be changed to clarify that it is intended for Windows), is the one that I was unsure about where it would need the `chmod` and `sudo mv` commands.

Image

paul-dingemans: I saw some more errors (made by me) an fixed them. The page now looks as follows:

Screenshot 2026-05-28 at 14 18 26

Screenshot 2026-05-28 at 14 18 42

zeitlinger:
Hopefully all fixed now.

paul-dingemans:
@zeitlinger setup-graalvm@v1 set jaav version to 21:
```
    - name: Set up GraalVM
      uses: graalvm/setup-graalvm@v1
      with:
        java-version: "21"
        distribution: "graalvm-community"
```
which is also printed in logging (see https://github.com/ktlint/ktlint/actions/runs/26574391176/job/78301924501?pr=3284#step:3:125):
```
Run graalvm/setup-graalvm@v1
  with:
    java-version: 21
    distribution: graalvm-community
```
This project always compiles ktlint at latest LTS java version (25), so it results in compilation errors (see https://github.com/ktlint/ktlint/actions/runs/26574391176/job/78301924501?pr=3284#step:3:201) when building the native images only.

Did you have any specific reason to set java version of graalvm to `21` instead of `25`?

paul-dingemans:
> @zeitlinger setup-graalvm@v1 set jaav version to 21:
> 
> ```
>     - name: Set up GraalVM
>       uses: graalvm/setup-graalvm@v1
>       with:
>         java-version: "21"
>         distribution: "graalvm-community"
> ```
> 
> which is also printed in logging (see https://github.com/ktlint/ktlint/actions/runs/26574391176/job/78301924501?pr=3284#step:3:125):
> 
> ```
> Run graalvm/setup-graalvm@v1
>   with:
>     java-version: 21
>     distribution: graalvm-community
> ```
> 
> This project always compiles ktlint at latest LTS java version (25), so it results in compilation errors (see https://github.com/ktlint/ktlint/actions/runs/26574391176/job/78301924501?pr=3284#step:3:201) when building the native images only.
> 
> Did you have any specific reason to set java version of graalvm to `21` instead of `25`?

I have just pushed a change with java 25 version in graalvm setup.

paul-dingemans:
> > @zeitlinger setup-graalvm@v1 set jaav version to 21:
> > ```
> >     - name: Set up GraalVM
> >       uses: graalvm/setup-graalvm@v1
> >       with:
> >         java-version: "21"
> >         distribution: "graalvm-community"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which is also printed in logging (see https://github.com/ktlint/ktlint/actions/runs/26574391176/job/78301924501?pr=3284#step:3:125):
> > ```
> > Run graalvm/setup-graalvm@v1
> >   with:
> >     java-version: 21
> >     distribution: graalvm-community
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This project always compiles ktlint at latest LTS java version (25), so it results in compilation errors (see https://github.com/ktlint/ktlint/actions/runs/26574391176/job/78301924501?pr=3284#step:3:201) when building the native images only.
> > Did you have any specific reason to set java version of graalvm to `21` instead of `25`?
> 
> I have just pushed a change with java 25 version in graalvm setup.

After upping the java version, the build of the native image finally started, but resulted in an error. See https://github.com/ktlint/ktlint/actions/runs/26582218674/job/78463238652#step:3:373. Please have a look @zeitlinger

paul-dingemans:
@zeitlinger Unfortunately the problem is not yet fixed. Can you please have another look?

zeitlinger:
> @zeitlinger Unfortunately the problem is not yet fixed. Can you please have another look?

gave it another go

.github/workflows/pull-request-with-code.yml
```diff
@@ -9,14 +9,24 @@ on:
     paths:
       - '**/*.kt'
       - '**/*.kts'
+      - '**/*.gradle'
```
paul-dingemans: Why is path "**/*.gradle" required? As far as I can see (on my local machine) it only picks up files from /build-logic/build/tmp/generatePrecompiledScriptPluginAccessors which shoud never be commited, so will never be in the PR.

.github/workflows/pull-request-with-code.yml
```diff
@@ -9,14 +9,24 @@ on:
...
+      - '**/*.gradle.kts'
```
paul-dingemans: Why is path `**/*.gradle.kts` added? I would expect this path to be matched by `**.kts`.

.github/workflows/pull-request-with-code.yml
```diff
@@ -9,14 +9,24 @@ on:
...
       - '**/*.properties'
       - '**/*.toml'
+      - '.github/actions/**'
+      - '.github/workflows/**'
```
paul-dingemans: Those patterns seems to wide to me. I would expect/prefer that that these patterns are restricted to files that are essential for the native build. For example a change in workflow file `publish-snapshot-docs.yml` should not trigger the `pull-request-with-code` workflow.

.github/workflows/publish-release-build.yml
```diff
@@ -88,3 +88,39 @@ jobs:
...
+    name: "Build GraalVM native-image (${{ matrix.os }})"
+    runs-on: ${{ matrix.os }}
+    permissions:
+      contents: write
+    needs: publish
+    if: github.repository == 'ktlint/ktlint'
+    strategy:
+      matrix:
+        # ubuntu-22.04 is used instead of ubuntu-latest to produce a binary compatible with a wider range
+        # of Linux distributions (glibc version constraint). See https://github.com/google/google-java-format/issues/1072.
```
paul-dingemans: Under which conditions would you advise to upgrade to newer Ubuntu versions?

paul-dingemans (changes requested):
@zeitlinger Congrats, the build has succeeded. That was quite a journey. Thanks for your persistence.

I have added some new review remarks. Also, there is one open review remark which was not addressed before:

* How is content of file ktlint-cli/src/main/resources/META-INF/native-image/reachability-metadata.json determined?

.github/workflows/pull-request-with-code.yml
zeitlinger: You are right; `**/*.gradle` was broader than needed here, so I removed it.

.github/workflows/pull-request-with-code.yml
zeitlinger: Also right — `**/*.kts` already covers `*.gradle.kts`, so I removed the redundant pattern.

.github/workflows/pull-request-with-code.yml
zeitlinger: Agreed. I narrowed the workflow triggers to the native-image action, the related PR/release workflows, and `ktlint-cli/**`, so unrelated workflows like `publish-snapshot-docs.yml` no longer trigger this workflow.

.github/workflows/publish-release-build.yml
zeitlinger: I added a note next to `ubuntu-22.04`: the plan is to revisit it once ktlint intentionally raises its minimum glibc compatibility baseline.

zeitlinger:
I pushed 077f26b4 to address the workflow-scope review comments.

On `ktlint-cli/src/main/resources/META-INF/native-image/reachability-metadata.json`: I derived it from the ServiceLoader registrations that the CLI bundles — the built-in `ReporterProviderV2` implementations plus `StandardRuleSetProvider`. Those are the provider classes GraalVM needs explicitly in the native image because they are discovered dynamically via ServiceLoader.

zeitlinger:
Yes - this was getting bigger than I thought

.github/workflows/publish-release-build.yml
paul-dingemans: Thank your for adding the note. But honestly, it does not really help me. I do not understand when, or when not, to raise the minimum glibc constraint as I don't know what the consequences are of doing so. So for example lets assume that a security vulnerability is detected in glibc which is not backported to Ubuntu 22.04. How would I detect that? How should I judge whether upgrading to a newer Ubuntu version is a good traid off?

.github/workflows/pull-request-with-code.yml
paul-dingemans: I do not understand why `.github/workflows/pull-request-without-code.yml` is added to the list of path that would trigger the `pull-request-with-code` workflow. Why would it be needed to run the full build including the  native image when I would just change the `pull-request-without-code` workflow itself?

Is it correct that `ktlint-cli/**` was added to ensure that the `reachability-metadata.json` file is being picked up? Or does it intend to pick up other files as well?

paul-dingemans:
> I pushed [077f26b](https://github.com/ktlint/ktlint/commit/077f26b414e0996e574d1fbe34e8191182750b07) to address the workflow-scope review comments.
> 
> On `ktlint-cli/src/main/resources/META-INF/native-image/reachability-metadata.json`: I derived it from the ServiceLoader registrations that the CLI bundles — the built-in `ReporterProviderV2` implementations plus `StandardRuleSetProvider`. Those are the provider classes GraalVM needs explicitly in the native image because they are discovered dynamically via ServiceLoader.

Ok clear. It is not a great that this needs to be done manually. I have another concern though. Ktlint CLI can use external rulesets and reporters as well which also use the `ReporterProviderV2` and `StandardRuleSetProvider` to load them dynamically from JARs provided via the commandline. Do you expect this to work with the native image, or not?

.github/workflows/publish-release-build.yml
zeitlinger: Fair — the note was too vague. Concrete criteria:

**When to bump:**
- GitHub Actions deprecates the `ubuntu-22.04` runner (announced on the [runner-images repo](https://github.com/actions/runner-images) 3–6 months ahead). This is the forcing function.
- A glibc CVE lands that Ubuntu 22.04 doesn't backport. Ubuntu 22.04 is in standard support until April 2027 (Pro/ESM later) and Canonical backports security fixes during that window, so realistically this collapses into the same signal as runner deprecation.

**Detection:** GitHub's deprecation announcement is the practical channel. No separate glibc-CVE monitor is worth wiring up — CVEs get patched into 22.04 automatically.

**Trade-off when bumping:** the native binary links against the build host's glibc. Bumping from 22.04 → 24.04 raises the minimum glibc from 2.35 to 2.39, which breaks users on RHEL 8, CentOS 7, Debian 11, and older Amazon Linux 2. So bump only when (a) forced by runner deprecation, or (b) a real user need requires a newer glibc feature.

I've replaced the comment with a shorter version pointing at these criteria in c86dd94b.

.github/workflows/pull-request-with-code.yml
zeitlinger: You're right on both counts.

- `pull-request-without-code.yml`: removed from the with-code trigger list. Editing the without-code workflow has no bearing on the with-code build. (The `name:` parity comment at the top of the file is just a GitHub PR-status convention — it doesn't require cross-triggering.)
- `ktlint-cli/**`: yes, the only reason it was there is `reachability-metadata.json` (Kotlin sources under `ktlint-cli/` are already covered by `**/*.kt`). Narrowed to `ktlint-cli/src/main/resources/META-INF/native-image/**` so unrelated ktlint-cli changes don't retrigger the native-image job.

Fixed in c86dd94b.

zeitlinger:
Good question, and honestly a limitation worth calling out.

Native-image is AOT — there is no JVM class loader at runtime, and `ServiceLoader` is resolved against providers that were reachable at image-build time. External ruleset/reporter JARs supplied via `-R`/`--reporter=…,artifact=…` on the CLI won't be discoverable in the native binary: their classes are not part of the compiled image, and native-image doesn't support loading arbitrary bytecode at runtime.

So the native binaries only work with the built-in rulesets/reporters. Anyone who needs custom or third-party JARs should keep using the JAR distribution. I've added a warning note to the install docs in c86dd94b so users pick the right artifact.

.github/workflows/pull-request-without-code.yml
```diff
@@ -12,13 +12,23 @@ on:
       - '!**/*.kts'
       - '!**/*.properties'
       - '!**/*.toml'
+      - '!.github/actions/build-native-image/**'
+      - '!.github/workflows/pull-request-with-code.yml'
+      - '!.github/workflows/pull-request-without-code.yml'
```
paul-dingemans: Shouldn't this line be removed, now it is removed from `pull-request-with-code`?

.github/workflows/pull-request-without-code.yml
```diff
@@ -12,13 +12,23 @@ on:
...
   pull_request:
     paths:
       - '**/*'
       - '!**/*.kt'
       - '!**/*.kts'
       - '!**/*.properties'
       - '!**/*.toml'
+      - '!.github/actions/build-native-image/**'
+      - '!.github/workflows/pull-request-with-code.yml'
+      - '!.github/workflows/pull-request-without-code.yml'
```
paul-dingemans: Shouldn't this line be removed, now it is removed from `pull-request-with-code`?

paul-dingemans (changes requested):

.github/workflows/pull-request-without-code.yml
zeitlinger: Good catch — yes, those stale exclusions should be removed as well. I just pushed 39533019 to drop them from `pull-request-without-code.yml`, so the with-code/without-code workflow routing is consistent now.

.github/workflows/pull-request-without-code.yml
zeitlinger: Good catch — yes, those stale exclusions should be removed as well. I just pushed 39533019 to drop them from `pull-request-without-code.yml`, so the with-code/without-code workflow routing is consistent now.

zeitlinger:
Great - can you create a release so that the binaries can be used?

paul-dingemans:
> Great - can you create a release so that the binaries can be used?

They binaries will be published with the upcoming 2.0.0 release. Currently I am working on the publication of an alpha release, and running in some troubles with the native images (https://github.com/ktlint/ktlint/actions/workflows/publish-release-build.yml).

paul-dingemans:
@zeitlinger I am having a concern and a problem with adding the native images to the release.

1) According to GraalVM build output description some of the [additional build artifacts](https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/BuildOutput.md#build-artifacts) have to be provided with the native image.

I see that the [Linux native image has additional build artifacts](https://github.com/ktlint/ktlint/actions/runs/32479975920/job/96764819772#step:3:469):
```
  Build artifacts:
   /home/runner/work/ktlint/ktlint/ktlint_linux-x86-64 (executable)
   /home/runner/work/ktlint/ktlint/libawt.so (jdk_library)
   /home/runner/work/ktlint/ktlint/libawt_headless.so (jdk_library)
   /home/runner/work/ktlint/ktlint/libawt_xawt.so (jdk_library)
   /home/runner/work/ktlint/ktlint/libfontmanager.so (jdk_library)
   /home/runner/work/ktlint/ktlint/libjava.so (jdk_library_shim)
   /home/runner/work/ktlint/ktlint/libjvm.so (jdk_library_shim)
```

[Additional build artifacts for Windows](https://github.com/ktlint/ktlint/actions/runs/32479975920/job/96764819764#step:3:620):
```
   D:\a\ktlint\ktlint\awt.dll (jdk_library)
   D:\a\ktlint\ktlint\java.dll (jdk_library_shim)
   D:\a\ktlint\ktlint\javaaccessbridge.dll (jdk_library)
   D:\a\ktlint\ktlint\jawt.dll (jdk_library)
   D:\a\ktlint\ktlint\jvm.dll (jdk_library_shim)
   D:\a\ktlint\ktlint\ktlint_windows-x86-64.exe (executable)
```

MacOs has no additional build artifacts.

I am not in favor to add all those additional build artifacts to the GitHub Release. Is it really needed to provide those additional artifacts?

2) Although the windows native image is created succesfully in this [job step](https://github.com/ktlint/ktlint/actions/runs/32479975920/job/96764819764), the upload of that image fails as it is not found. For Linux and MacOs it works as expected.

paul-dingemans:
> 2. Although the windows native image is created succesfully in this [job step](https://github.com/ktlint/ktlint/actions/runs/32479975920/job/96764819764), the upload of that image fails as it is not found. For Linux and MacOs it works as expected.

At least, this problem is resolved in https://github.com/ktlint/ktlint/actions/runs/32497408309/job/96821858530.

paul-dingemans:
Native images (without the additional build artifacts) are now published on release page: https://github.com/ktlint/ktlint/releases#release-2.0.0-ALPHA-4

PR #3309, opened 2026-05-28
Match kotlin language version (2.3) with Gradle 9.5

paul-dingemans:
## Description

Match kotlin language version (2.3) with Gradle 9.5

## Checklist

- [ ] Tests are added
- [ ] PR description added (background information)

[Documentation](https://ktlint.github.io/ktlint/) is updated. See [difference between snapshot and release documentation](https://github.com/pinterest/ktlint/tree/master/documentation)

### DIFF
--- a/build-logic/src/main/kotlin/KotlinCommonPlugin.kt
+++ b/build-logic/src/main/kotlin/KotlinCommonPlugin.kt
@@ -36,8 +36,8 @@ abstract class KotlinCommonPlugin : Plugin<Project> {
                 jvmToolchain(jdkVersion = javaCompilationVersion.asInt())
 
                 compilerOptions {
-                    apiVersion.set(KotlinVersion.KOTLIN_2_2)
-                    languageVersion.set(KotlinVersion.KOTLIN_2_2)
+                    apiVersion.set(KotlinVersion.KOTLIN_2_3)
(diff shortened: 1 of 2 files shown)


build-logic/src/main/kotlin/KotlinCommonPlugin.kt
```diff
@@ -36,8 +36,8 @@ abstract class KotlinCommonPlugin : Plugin<Project> {
                 jvmToolchain(jdkVersion = javaCompilationVersion.asInt())
 
                 compilerOptions {
-                    apiVersion.set(KotlinVersion.KOTLIN_2_2)
-                    languageVersion.set(KotlinVersion.KOTLIN_2_2)
+                    apiVersion.set(KotlinVersion.KOTLIN_2_3)
+                    languageVersion.set(KotlinVersion.KOTLIN_2_3)
```
Goooler, line 40: It should be still 2.2 here

https://docs.gradle.org/current/userguide/compatibility.html#kotlin

paul-dingemans: Oops, I misread the compatibility table than. Tnx for catching this. But I do wonder what possible could go wrong as all tests succeeds.

Issue #3362, opened 2026-08-15
standard:indent cannot format multiline where clauses with tab indentation


autaut03:
Hi! ktlint 1.8.0 appears unable to produce valid formatting for multiline `where` clauses when tabs are enabled.

`.editorconfig`:

```editorconfig
root = true

[*.{kt,kts}]
indent_style = tab
indent_size = 4
```

`Repro.kt`:

```kotlin
private interface First

private interface Second

private fun <T> example(value: T)
    where T : First,
          T : Second {
    println(value)
}
```

Run:

```shell
ktlint --format Repro.kt
ktlint Repro.kt
```

The formatter correctly changes the function body indentation to a tab, but leaves the `where` clause indented with spaces. The following lint run then reports:

```text
Repro.kt:6:1: Unexpected space character(s) (standard:indent)
Repro.kt:7:1: Unexpected space character(s) (standard:indent)
```

`--format` also warns that it could not resolve all autocorrectable violations, and rerunning it does not help.

Expected: `ktlint --format` should produce output accepted by a subsequent ktlint check when `indent_style = tab`.

This may be related to #2175, but that report used space indentation and concerned alignment width; this reproduction is specific to tab indentation.

Environment: standalone ktlint CLI 1.8.0 on macOS.

PR #3374, opened 2026-08-21
fix(indent): accept aligned where constraints with tabs

autaut03:
Hey! This is continuation of https://github.com/ktlint/ktlint/pull/3363, in relation to issue #3362. It only solved part of the problem. I rely on AI too much these days so apologies for that.

## Description

Previous PR fixed the actual indentation produced for `where` - that still applies. But actual indentation check has a method that whitelists specific combinations of tabs + whitespaces (smart tabs). One of those cases is the 6 spaces before a type constraint in `where`:

```
                TYPE_CONSTRAINT -> {
                    // 6 spaces (length of "where" keyword plus a separator space) to indent type constraints as below:
                    //    where A1 : RecyclerView.Adapter<V1>,
                    //          A1 : ComposableAdapter.ViewTypeProvider,
                    TYPE_CONSTRAINT_CONTINUATION_INDENT
                }
```

The previous fix relied on this working correctly, but it doesn't - because `nextLeaf == IDENTIFIER "A1"` in the example above, which fails the check. This PR fixes it to look for the `nextCodeSibling` instead (which is the `TYPE_CONSTRAINT` we're looking for).

Again, apologies for coming back to this and not testing it properly.

## Checklist

- [X] Tests are added

[Documentation](https://ktlint.github.io/ktlint/) is updated. See [difference between snapshot and release documentation](https://github.com/ktlint/ktlint/tree/master/documentation)
- [ ] [Snapshot documentation](https://github.com/ktlint/ktlint/tree/master/documentation/snapshot) in case documentation is to be released together with a code change
- [ ] [Release documentation](https://github.com/ktlint/ktlint/tree/master/documentation/release-latest) in case documentation is related to a released version of ktlint and has to be published as soon as the change is merged to master


paul-dingemans:
@autaut03 The suggested fix is still not entirely correct. 

You added test case below:
```
        @Test
        fun `Given a comma at the start of a continuation line`() {
            val code =
                """
                private fun <T> example(value: T)
                ${TAB}where T : First
                ${TAB}      , T : Second {
                ${TAB}println(value)
                }
                """.trimIndent()

            indentationRuleAssertThat(code)
                .withEditorConfigOverride(INDENT_STYLE_TAB)
                .hasLintViolation(3, 1, "Unexpected space character(s)")
        }
```

This confused me for multiple reasons. In this example I would not have expected the violation to be emitted as the indentation before the comma is correct. Next thing that I found was that if the actual indentation contains too few trailing space to fix the WHERE indentation the message should not be emitted as well. Last reason of confusion is that a continuation line in a type constraint list should not start with a comma. That comma should have been moved to the end of the previous line by the `comma-spacing` rule for which I will create a separate fix.

paul-dingemans:
> Last reason of confusion is that a continuation line in a type constraint list should not start with a comma. That comma should have been moved to the end of the previous line by the `comma-spacing` rule for which I will create a separate fix.

Fixed in `wrapping` rule: https://github.com/ktlint/ktlint/pull/3378

